Issue #47, opened 2015-05-22
Bob has only one test per default?


ghajba:
I find it a bit strange to have only one test for bob which is running when you run the tests. And the exercise can be completed with the default case of the condition.

Is this on purpose?

rubysolo:
The ruby test suites are set up to start with only one failing test and everything else marked as "skip".  At the time these were ported to Elixir, ExUnit had no facility for doing this, so we tried to achieve the same thing by commenting out the later tests.  Now that ExUnit supports filtering by tags, we should uncomment these, tag with `@pending` and update the instructions to describe filtering out pending test cases.

ghajba:
In this case I need some clarification why the tests are skipped in ruby by default?
Most users would look at the tests to see if they are ignored / pending but there would be some who test against only one case and if it is OK they submit the solution. Then there would be some angry nitpicks mentioning that the code did not pass all the tests...

rubysolo:
That has happened.  :smile:  I believe the original intent was to not present a wall of errors to users who are new to programming and/or TDD, as that might be overwhelming.  By focusing on a single test failure and then re-enabling just the next test when one is passing, the user has a path to follow.

ghajba:
OK :)
It is another approach to have the problem solvers look at the test cases too. But this should be done consistent across all the tracks, shouldn't it? I mean in the Java or Python tracks for example you get all the test failures right at the start so you can see that is wrong with the code -- or at least what cases you have to solve so you don't miss anything and shield yourself against angry nitpicks.

rubysolo:
Yes, I would agree that it should be consistent.  If the model is to have pending tests, enabled one at a time, then the suites should be updated to match (if each language / testing framework has a way to do that).  Also, it should be clear in the instructions that the exercise is not complete until there are no more pending tests and no more failures.

pminten:
I know of at least one track, the Go one, where there was an explicit choice to enable all tests because hiding errors is not the Go way.

The idea behind one test at a time seems to be a TDD-like red/green/refactor. However that doesn't work because often what the user will do is write all of the code, see that the first test passes and turn on all other tests.

The only real way in which I could see "staged tests" working is if you collect them into groups of equivalent level and have the exercism tool check for whether the tests pass and if so enable the next group. With group I mean something like for bob to have a group that tests the basic logic and then a second group that tests edge cases and stuff like unicode.

Barring this I would recommend just enabling all tests all the time.

kytrinyx:
[...]

PR #48, opened 2015-05-26
bob: tag tests pending

ghajba:
Added an ExUnit tag for pending test cases. This enables excluding tests from the default test suite without commenting out the tests.

With a future release (or building currently from source) of Elixir the number of skipped tests will be visible too (https://github.com/elixir-lang/elixir/pull/2898):

```14 tests, 1 failure, 13 skipped```

This would be a good approach for #47

### DIFF
--- a/bob/bob_test.exs
+++ b/bob/bob_test.exs
@@ -5,6 +5,7 @@ else
 end
 
 ExUnit.start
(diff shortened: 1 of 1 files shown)


rubysolo:
:+1: thanks!

PR #49, opened 2015-06-02
Add additional student name to Grade School test

bartj3:
Without the third student name the test might pass without an implementation (if the student names are prepended instead of appended in the add implementation).


pminten:
Good catch. Thanks.

PR #51, opened 2015-06-14
Bring tests and example code up to date with v1.0.4

tjcelaya:
This is for exercism/xelixir#12

I wrote a small utility to run all the tests with the example code and uncommented the anagram tests since the `@tag pending` method of skipping tests is now possible. Unfortunately ExUnit doesn't seem to be reporting skipped tests yet for some reason, just the fact that a tag is being excluded so I'll throw those in there once skipped tests start being reported (which will deal with exercism/xelixir#47). ~~I'm opening a PR on elixir-lang/elixir to get the CLI output implemented, the data is already in place it's just not being output~~ It seems like the v1.0 branch diverged from master a long time ago and the "#{x} skipped" output is stuck in a commit from November 2014, still checking into elixir-lang/elixir#2898.

Let me know if there's anything I missed or can improve.

(I presume I should squash the commits? Never really done a PR with more than a single commit before.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,6 @@
 .DS_Store
 tmp
 bin/configlet
-bin/configlet.exe
+bin/configlet*
+deps
+_build
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
-language: bash
+language: elixir
 
 script:
   - bin/fetch-configlet
   - bin/configlet .
+  - mix deps.get
+  - EXERCISM_TEST_EXAMPLES=true mix test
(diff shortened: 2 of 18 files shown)


kytrinyx:
Yeah, it would be great to squash the commits. This looks great--I'll let the elixir folks review and then merge when they're happy. Thanks!

NobbZ:
Can this one be merged? I just applied patch to `zipper` by hand, because it can't be used in elixir 1.0.x because of `defstruct`.

pminten:
Thanks. I'm not very up-to-speed on the latest Elixir developments, but this looks good (and travis passes).

PR #52, opened 2015-07-06
Alter tests to ignore order of returned allergies

meadsteve:
The spec doesn't make it clear what order these should be returned in so it shouldn't matter.

### DIFF
--- a/allergies/allergies_test.exs
+++ b/allergies/allergies_test.exs
@@ -11,35 +11,35 @@ defmodule AllergiesTest do
   use ExUnit.Case, async: true
 
   test "no_allergies_at_all" do
-    assert Allergies.list(0) == []
+    Allergies.list(0) |> assert_is_a_set_containing []
   end
 
   test "allergic_to_just_eggs" do
-    assert Allergies.list(1) == ["eggs"]
+    Allergies.list(1) |> assert_is_a_set_containing ["eggs"]
   end
 
   test "allergic_to_just_peanuts" do
-    assert Allergies.list(2) == ["peanuts"]
+    Allergies.list(2) |> assert_is_a_set_containing ["peanuts"]
   end
 
   test "allergic_to_just_strawberries" do
-    assert Allergies.list(8) == ["strawberries"]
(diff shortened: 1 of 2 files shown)


kytrinyx:
Yeah, I agree that these should be order agnostic.

rubysolo:
Maybe it should be changed to return a HashSet instead of a List.

meadsteve:
@rubysolo that could work. What would you have as the keys? Something like:
```elixir
%{
  1 => "eggs",
  4 => "strawberries"
}
```

meadsteve:
Although I think all the other languages are lists rather than hashsets. So might be better to keep it consistent across all of them

kytrinyx:
I don't particularly think we need to be consistent across languages (in case that helps make a decision).

meadsteve:
@kytrinyx I think it's nice when there is the consistency though. It feels like it helps highlight the differences/similarities between the languages.

kytrinyx:
I never thought about it that way. Is there a `set` implementation that would make it order independent?

rubysolo:
Yeah, that's `HashSet` in Elixir.

```elixir
Enum.into([1,2,3], %HashSet{})
iex(1)> Enum.into([1,2,3], %HashSet{})
#HashSet<[2, 3, 1]>
iex(2)> Enum.into([1,2,3], %HashSet{}) == Enum.into([3,2,1], %HashSet{})
true
```

kytrinyx:
I'm going to make sure this is order independent in go, which means using a map, not a slice. Same deal as here, really.

meadsteve:
@rubysolo sorry I completely misread your first message as hashmap. The set is an excellent idea! :+1: I'll update my PR.

meadsteve:
@rubysolo @kytrinyx What do you think about the test and example now? I've update the example so it uses a HashSet and tests accept any valid Set.

rubysolo:
Great, thanks!

meadsteve:
No problem. Thank you both for an awesome tool!

PR #53, opened 2015-07-10
grains_test: use === to disallow floats

henrik:
`1 == 1.0` is true but `1 === 1.0` is false.

You easily get floats here if you use `:math.pow`.


pminten:
Good catch. Thanks.

PR #58, opened 2015-09-04
Add a test: correct # of digits but wrong format

brianriley:
I copied over a test that I think is important from the Ruby version. It
checks that the number of digits might be correct while the formatting
of the phone number string might be wrong.

I think that the key to this exercise is to check for the 3-3-4 format,
where the separators can be anything, but there needs to be a group of 3
digits followed by another group of 3 digits followed by another group
of 4 digits, optionally prefixed with a single "1".

When I added this test to my initial iteration, the suite failed. Most
of the other submissions I saw would, I believe, also fail. So I think
this test is important for clarifying the problem space.

### DIFF
--- a/phone-number/example.exs
+++ b/phone-number/example.exs
@@ -13,17 +13,14 @@ defmodule Phone do
   @spec number(String.t) :: String.t
   def number(raw) do
     raw
-    |> numeric
-    |> String.graphemes
-    |> do_number
+    |> to_parts
     |> to_string
   end
 
-  defp numeric(input), do: String.replace(input, ~r/[^0-9]*/, "")
-
-  defp do_number(input) when length(input) == 10, do: input
-  defp do_number(["1"|input]) when length(input) == 10, do: input
-  defp do_number(_), do: @bad_result
+  @spec to_parts(String.t) :: [String.t]
+  defp to_parts(raw) do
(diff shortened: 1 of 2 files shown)


kytrinyx:
Oh, nice! Did you check that the example solution passes with the new test?

brianriley:
I just ran the test suite with the example solution, and it fails. The very first transformation in the example, and in my initial solution and a lot of the other solutions I saw submitted, is to strip all non-digit characters from the string. This means that "1a2a3a4a5a6a7a8a9a0a" is valid, when the test says that it should be invalid.

I'm not quite sure of a quick fix to make the example pass--except, I guess, to turn it into my final solution. Thoughts on how to proceed here?

kytrinyx:
Feel free to replace it with your solution. I'm sure others will jump in later if they have improvements.

brianriley:
I added my change to the example solution. I guess another approach could be do define valid separators (plus signs, parentheses, hyphens, spaces, and periods?) and strip only those then check to see if there are any non-digits left in the result, but this approach does that all in one pass while changing the focus of the problem to be to look for the 3-3-4 pattern of US phone numbers.

kytrinyx:
With the reference solutions we aim for code that is as straight forward as possible, without necessarily making it exemplary. The solution is used only to validate the test suite (though not all languages have continuous integration in place.

Thanks for taking the time to improve the exercise.

PR #60, opened 2015-10-15
Change test order for a better flow

MarcosX:
While doing the word count exercise I noticed that the test `normalize case` comes only after the `German` one, but the `German` test also requires a case normalization.

So in this pull request I've changed the order of those two tests to make sure that when we get to the `German` test we already have covered the case normalization.


kytrinyx:
That's a great observation, thanks for catching that.

PR #61, opened 2015-10-18
Add pending tag to focus on a single test case at a time

MarcosX:
The `@tag :pending` was missing in this example and all tests were running at the same time, which I find hard to read and focus on compared to the precious exercises.


rubysolo:
:+1: thanks!

PR #62, opened 2015-10-20
clarify that the Russian string is in upper-case

digitalronin:
This makes it clear that the exercise is to alter the code to
match unicode upper-case strings.


kytrinyx:
That's a great observation, thanks!

PR #63, opened 2015-10-30
added @tag :pending to all test files' tests except the first test of…

jinyeow:
… each test file for test files that did not have this. I found that running the test files for new exercises and seeing a bunch of failed tests especially when there are a lot of them was particularly unhelpful.

So I added
    ExUnit.configure exclude: :pending

to the test files under ExUnit.start and added '@tag :pending' to all the test files except the first test of each test file so that the tests can be easily run and the output easily read.

### DIFF
--- a/allergies/allergies_test.exs
+++ b/allergies/allergies_test.exs
@@ -6,56 +6,69 @@ end
 
 
 ExUnit.start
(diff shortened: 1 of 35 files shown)


rubysolo:
Nice, thanks!

PR #65, opened 2015-11-17
Set ExUnit into trace mode

bordeltabernacle:
In response to Issue#47.  Add 'trace: true' to Elixir tests,
providing a more verbose output that includes indicating what
tests are skipped.


rubysolo:
:+1: thanks!

PR #70, opened 2015-12-01
Adds one more test to catch double counting multiples of both 3 and 5

CoderDennis:
This one tripped me up. Hopefully the test to 20 will make finding the double counting error a little more intuitive.

### DIFF
--- a/sum-of-multiples/sum_of_multiples_test.exs
+++ b/sum-of-multiples/sum_of_multiples_test.exs
@@ -26,6 +26,11 @@ defmodule SumOfMultiplesTest do
   end
 
   @tag :pending
+  test "sum to 20" do
+    assert SumOfMultiples.to(20) == 78
+  end
+
+  @tag :pending
   test "sum to 1000" do
     assert SumOfMultiples.to(1000) == 233168
   end


kytrinyx:
Thanks!

kytrinyx:
@parkerl what do you think about this? It seems harmless to add the extra test, but you might see some gotchas that I haven't thought through.

parkerl:
@kytrinyx and @CoderDennisI think this is fine. My only question is would it be better to test up to `16` to make it more clear that the problem is with double counting 15?

CoderDennis:
@parkerl I don't think it matters. If the difference between the expected result and the double counted result is 15, that should be enough to figure it out. I thought about naming the test something like `"sum to 20 doesn't double count multiples of both 3 and 5"`, but I wasn't sure how explicit it should be. Is the goal to lead someone directly through all the steps or just give them enough clues to figure it out?

When I was working on this exercise I had no idea why my to 1000 result was so far off the expected number. @rubysolo on the elixir #exercism slack channel helped me figure it out. That's also where the suggestion to send a pull request was made.

kytrinyx:
I think clues are better than explicit instructions. I think the suggestion for testing up to 16 is interesting. Would that be a stronger clue?

parkerl:
I'm went ahead and merged this. Thanks @CoderDennis! :tada:

kytrinyx:
:100:

PR #72, opened 2015-12-04
Adding test cases to Sum of Multiples

krishandley:
Adding test cases to Sum of Multiples as per exercism/exercism.io#2341


parkerl:
Thank you! :tada:

PR #75, opened 2015-12-10
Update gigasecond per issue #35

parkerl:
@kytrinyx I think this brings the Elixir gigasecond into line with the Ruby version.

//cc @rubysolo


kytrinyx:
Sweet!

PR #77, opened 2015-12-17
Update nucleotide count

toriejw:
Working with @parkerl to bring this exercise more in line with Ruby version


parkerl:
We know there are a couple whitespace issues. Any other issues?

PR #79, opened 2015-12-19
Hello world argument number

Egogre:
Being brand new to elixir, it was confusing why the function def hello(name) was not found when running the test calling HelloWorld.hello() and get this error.
** (UndefinedFunctionError) undefined function: HelloWorld.hello/0

the "Your implementation goes here" also makes it seem like I should only have to change the inside of the function in order to get this to pass. 

Instead of making a new empty argument function to show the passing test and then making another hello function that takes one argument, I changed the file to have comments to help guide a new user through their first elixir exercism.

Thanks also to @hectorhuertas for helping me through this!


kytrinyx:
Would you mind rebasing onto master?

Egogre:
I think the idea of rebasing is to keep the history linear and easy to track and avoid unnecessary merging commits. I am not rebasing correctly I think. It keeps wanting me to resolve the conflicts and create another commit before rebasing. I am going to pull this PR and submit a new one with master rebased using a method I am more familiar with.

PR #80, opened 2015-12-19
Proper russian translation for "Get out"

vladimir-tikhonov:
"Get out" is actually "Уходи" in russian, not "Уходить" :ru:


kytrinyx:
Haha, thank you! I was wondering about that.

PR #82, opened 2015-12-21
added in comments for beginning user clarity

Egogre:
Being brand new to elixir, it was confusing why the function def hello(name) was not found when running the test calling HelloWorld.hello() and get this error.
** (UndefinedFunctionError) undefined function: HelloWorld.hello/0

the "Your implementation goes here" also makes it seem like I should only have to change the inside of the function in order to get this to pass.

Instead of making a new empty argument function to show the passing test and then making another hello function that takes one argument, I changed the file to have comments to help guide a new user through their first elixir exercism.

Thanks also to @hectorhuertas for helping me through this!

### DIFF
--- a/hello-world/hello_world.exs
+++ b/hello-world/hello_world.exs
@@ -1,4 +1,20 @@
 defmodule HelloWorld do
+
+  # Elixir counts the number of arguments as part of the function name.
+  # For instance,
+
+    #def hello() do
+    #end
+
+  # would be a completely different function from
+
+    #def hello(name) do
+    #end
+
+  # Can you find a way to make all the tests pass with just one function?
+  # Hint: look into argument defaults here:
+  # http://elixir-lang.org/getting-started/modules.html#default-arguments
+
   def hello(name) do
 
     "Your implementation goes here"


parkerl:
@kytrinyx I'm ok with this for the hello world. You?

kytrinyx:
How about sticking this in a file named `HINTS.md`?

kytrinyx:
Actually, I am OK with this for Hello World.

Egogre:
@parkerl I am fine with putting the comments above def hello() or with making the hints.md file. Let me know what you would prefer!

parkerl:
Please put them above. Thanks

Egogre:
This has been done. I also squashed the commits.

parkerl:
Hey @Egogre sorry for the delay...holidays. :blush: 

I don't see your updated commits. Did you push them?

Egogre:
@parkerl Ah! Thanks! I had squashed them but didn't pay attention when I pushed. I am new to squashing so didn't realize it set my commit history back and then thought the origin branch was newer. I think it is ready now. Happy Holidays!

parkerl:
Thanks! :tada:

PR #83, opened 2015-12-26
Rename outdated method name

MarcosX:
The tests and `example.exs` reference the method as `histogram` but in the `dna.exs` it's still called `nucleotide_count`.


parkerl:
Thanks for updating that! :tada:

PR #84, opened 2015-12-31
Style modifications

ggpasqualino:
* Replace `assert !` by `refute`
* Replace word list literals by sigils


rubysolo:
Thank you!

PR #86, opened 2016-01-05
Include 'pending' specs in CI build

parkerl:
It seems that CI should run all tests against the example.exs files. Is that what other language tracks do? I assume the examples should always pass all tests which doesn't look like the case for us right now. 

//cc @kytrinyx @rubysolo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,5 +4,5 @@ script:
 - bin/fetch-configlet
 - bin/configlet .
 - mix deps.get
-- EXERCISM_TEST_EXAMPLES=true mix test
+- EXERCISM_TEST_EXAMPLES=true mix test --include pending
 sudo: false


parkerl:
My machine is on Elixir 1.2 and this fails:

```
 1) test inspect (CustomSetTest)
     custom-set/custom-set_test.exs:123
     Assertion with == failed
     code: inspect(CustomSet.new()) == "#<CustomSet []>"
     lhs:  "%CustomSet{list: []}"
     rhs:  "#<CustomSet []>"
     stacktrace:
       custom-set_test.exs:124
```

Travis is on 1.0.2. looks like we should think about updating as we are two minors behind?

rubysolo:
Yes, we should update -- I vote for 1.2, making sure that the documentation communicates that requirement.  Thoughts?

kytrinyx:
> Is that what other language tracks do?

Yeah, the ones that actually run the tests on CI generally do some form of copy into a temp dir, rewrite the files to have the correct name (if loading depends on the file name), and then removes pending directives, and finally runs the tests.

parkerl:
@kytrinyx phew glad we don't have to jump through those hoops. Merging this and opening an issue to target 1.2.

PR #88, opened 2016-01-06
Elixir is now installable directly via pacman

nickpellant:


nickpellant:
There also now exists a detailed "Installing Elixir" page on the Elixir website:

http://elixir-lang.org/install.html

Might this be worth linking to  here instead?

kytrinyx:
I think linking to elixir-lang's install page would be good--less of a maintenance burden on our end.

parkerl:
@kytrinyx sounds good.

PR #90, opened 2016-01-08
Added missing test that made checking super lists necessary

jason-kerney:
I was able to complete all the tests without checking the values in a super list. All I needed to do was check the size.

I added a test case that makes it more explicit.

### DIFF
--- a/sublist/sublist_test.exs
+++ b/sublist/sublist_test.exs
@@ -83,6 +83,11 @@ defmodule SublistTest do
   end
 
   @tag :pending
+  test "1 and 2 does not contain 3" do
+    assert Sublist.compare([1,2], [3]) == :unequal
+  end
+
+  @tag :pending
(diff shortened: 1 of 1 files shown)


parkerl:
Thanks! :tada:

jason-kerney:
Why was the test not added to the tests file? There is no test that checks that condition and super lists can be defined as any list with a larger size without it.

Take a look at my solution that shows it:
http://exercism.io/submissions/d8d9c3d8efc64423b4fee36c7e2b64b4

Thank you,
Jason Kerney

parkerl:
@jason-kerney so sorry my bad. I must have accidentally closed this without merging. Thanks for catching my mistake. :disappointed:

jason-kerney:
No problem. Just wanted to make an awesome project better. ;-)

PR #93, opened 2016-01-18
Added 2 invalid input testcases

NobbZ:
This is according to exercism/x-common#95

Since I'm only on mobile I need to have travis do the test run for me.


NobbZ:
I'm trying to fix the build locally right now, I'm back at my laptop.

kytrinyx:
+1 for these tests.

rubysolo:
Thanks!

PR #98, opened 2016-01-21
Replace map with keyword list

bordeltabernacle:
Maps are unsorted, but they currently have a strange quirk due to
an Erlang implementation detail which means that maps with less than
33 keys are actually sorted.
This means the "sort school by grade and by student name" test passes
when it shouldn't.
Unlike maps, keyword lists are ordered, as specified by the developer.

http://exercism.io/submissions/2a370c046f5449519e3c976fcaf58e68

### DIFF
--- a/grade-school/example.exs
+++ b/grade-school/example.exs
@@ -1,16 +1,15 @@
 defmodule School do
-  def add(db, student, grade) do
-    Dict.update db, grade, [student], &([ student | &1 ])
+  def add(db, name, grade) do
+    Keyword.update(db, grade, [name], &[name|&1])
   end
 
   def grade(db, grade) do
-    Dict.get db, grade, []
+    Keyword.get(db, grade, [])
   end
 
   def sort(db) do
     db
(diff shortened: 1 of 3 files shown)


parkerl:
@bordeltabernacle thanks for the PR! I'd like to review it more detail but wanted to let you know we have not forgotten about it.

bordeltabernacle:
No worries, thanks @parkerl .
I'm interested to hear your take on it, I'm sure there's things I've not considered, but it was a fairly simple change to make, hence a PR instead of an issue.

rubysolo:
LGTM, can you rebase with master and squash?

bordeltabernacle:
I *think* I've rebased & squashed appropriately, let me know if I've not!
Thanks :smile:

rubysolo:
Hmm, I was expecting the commits that are already in xelixir/master to go away.  Maybe you need to update your local master to HEAD of xelixir/master first?

bordeltabernacle:
right, that's better I think....

rubysolo:
:tada:

bordeltabernacle:
:+1:

PR #99, opened 2016-01-22
dries up the code loading logic for each test file

mindreframer:
problem: 
- each test file has some boilerplate code for loading either example or the users implementation  + starting ExUnit
- running all tests with test/test_helper via `mix test` did not work

solution: 
- provide a file_loader, that has the logic bits which file to load
- it is also starting the ExUnit and centralizes the configuration for it
- update the test/test_helper to do a bit less
- update SETUP.md with instructions how to run test for all examples 

Gotchas: 
- you will see warnings due to redefined FileLoader module: 
`file_loader.ex:1: warning: redefining module FileLoader`

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -32,5 +32,17 @@ Or, you can enable all the tests by commenting out the
 # ExUnit.configure exclude: :pending, trace: true
 ```
 
+
+
+## Run all tests
+```bash
+
+# with your implementation
+$ mix run test/test_helper.exs
+
+# with example implementation
+$ EXERCISM_TEST_EXAMPLES=true mix run test/test_helper.exs
+```
+
 For more detailed information about the Elixir track, please
 see the [help page](http://exercism.io/languages/elixir).
--- a/allergies/allergies_test.exs
+++ b/allergies/allergies_test.exs
@@ -1,12 +1,5 @@
-if System.get_env("EXERCISM_TEST_EXAMPLES") do
-  Code.load_file("example.exs")
-else
-  Code.load_file("allergies.exs")
-end
-
-
-ExUnit.start
-ExUnit.configure exclude: :pending, trace: true
+Code.load_file("../file_loader.ex", __DIR__)
+FileLoader.load("allergies", __DIR__)
 
 defmodule AllergiesTest do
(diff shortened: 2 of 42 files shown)


mindreframer:
I'm stupid... Just seen the travis.yml command... 
`EXERCISM_TEST_EXAMPLES=true mix test --include pending **/*_test.exs`

But still, it's less noise in each test file, so maybe useful...

parkerl:
@mindreframer thanks for your PR! As you said above, the expected way to run all tests is in the travis.yml. We do need to document that in the README asap. 

I have mixed feelings about merging this PR. I appreciate your hard work but this is a rather sweeping change that was not previously discussed. I humbly suggest that in the future you open an issue before composing such a PR. 

@rubysolo what are your thoughts on this one?

mindreframer:
@parkerl You're right about opening an issue for discussion first... I did all the work upfront, only to find the command to run all tests when the pull request was marked green by travis... then I hit me to just look into the travis.yml file... really stupid. 

On the other side it reduces the boilerplate in each testfile from 7 to 2 lines of code, maybe that's a small win. Overall it saves  328 - 113 => 215 lines of code and puts that logic behind an API. 

I understand your mixed feelings and hope you'll decide what's more maintainable for xelixir in the long run. 

Cheers, 
Roman

rubysolo:
I think we should merge it.  We don't have these exercises set up as mix projects, so this isn't any more non-standard, and it does reduce some boilerplate.

kytrinyx:
At the moment each exercise has to be self-contained. I think this would work for .travis.yml, but not if you pull down an individual exercise, since the user wouldn't have the top-level file.

parkerl:
Thanks @kytrinyx, of course. Not sure how we all missed that. Thanks @mindreframer for your work on this but unfortunately the duplication in the exercises is something we have to deal with as each must stand on its own.  If you are looking for another way to help, we could really use to clean up the pipe warnings given by Elixir 1.2...see https://github.com/exercism/xelixir/issues/87.

mindreframer:
@kytrinyx @parkerl  Oh, missed that! Well then it's moot and can be closed. Sorry for confusion.

PR #102, opened 2016-01-26
Fix broken link

smcabrera:
Current link for Getting Started appears to be broken. This is probably the current equivalent: http://elixir-lang.org/getting-started/introduction.html


rubysolo:
Thanks!  :tada:

PR #105, opened 2016-01-29
Fix apparent rename missed in test description

iamvery:
:abc:


kytrinyx:
Good catch, thanks!

iamvery:
You're welcome. Happy to help where I can!

PR #107, opened 2016-02-01
Remove `async: true` from tests as it is redundant.

bordeltabernacle:
When trace is set to true, as it is in all tests to
show more detailed information, tests are run
synchronously, therefore `async: true` has no effect.
`async: false` is the default.


rubysolo:
TIL!  :+1:

PR #108, opened 2016-02-01
Add typespecs to Hello World exercise, move help into moduledoc

bordeltabernacle:


parkerl:
Yes that is much nicer. Thanks! :tada:

Issue #87, opened 2016-01-05
Update the Elixir track to target version 1.2


parkerl:
We should update our target version to Elixir 1.2. 
 - [x] Update CI
 - [x] Fix test failures
 - [x] Update documentation
 - [x] Clean up pipe warnings
```
#We are getting quite a few of these in the examples after the 1.2 upgrade

warning: you are piping into a function call without parentheses, which may be ambiguous. Please wrap the function you are piping into in parentheses. For example:

    foo 1 |> bar 2 |> baz 3

Should be written as:

    foo(1) |> bar(2) |> baz(3)
```

bordeltabernacle:
Sorry for the extra commits, shoulda squashed them! https://github.com/exercism/xelixir/pull/111

bordeltabernacle:
Sorry about all the extra commits, shoulda squashed! https://github.com/exercism/xelixir/pull/112

PR #112, opened 2016-02-02
Clean up pipe warnings, re. issue #87.

bordeltabernacle:
Also remove whitespace & rename files to use camel_case


parkerl:
:heart: this @bordeltabernacle ! However, it looks like there are a bunch of whitespace changes (which I think we should include). Would you mind if I split these changes into two commits?

parkerl:
@bordeltabernacle thanks for this! :tada:

bordeltabernacle:
:+!: No worries, sorry I did't reply sooner, hope it's ok as is.

PR #126, opened 2016-02-19
Pattern match in the test for account

dalexj:
This confused me for a moment when I first opened the test file, so I think this change will be a bit more helpful to the reader.


rubysolo:
:+1: Thanks!

PR #129, opened 2016-02-22
Bank-account: test that amount is added to balance

martinsvalin:
As of now, the tests for bank-account pass, even if the implementation replaces balance, rather than adding amount to the existing balance.


parkerl:
Looks good to me. @rubysolo ?

rubysolo:
:+1:

PR #133, opened 2016-02-23
Add tests for `up` in zipper exercism

### LINKED ISSUE #131: `up` function not tested in zipper exercism
`Zipper.up` is never used in the `zipper_text.exs` but the stub for it in the example is still there. I can write a couple tests for it and send those in, unless somebody else wants to that knows more about the exercises than me

dalexj:
closes #131

### DIFF
--- a/exercises/zipper/zipper_test.exs
+++ b/exercises/zipper/zipper_test.exs
@@ -78,7 +78,6 @@ defmodule ZipperTest do
     assert (t1 |> from_tree |> left |> right |> up |> set_value(5) |> to_tree) == t2
   end
 
-
   @tag :pending
   test "set_left with leaf" do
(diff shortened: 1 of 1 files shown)


rubysolo:
:+1: Thanks!

parkerl:
Nice. :tada: 

@rubysolo do you think this test file should be more idiomatically formatted? Each pipe on a new line?

dalexj:
from the perspective of somebody that just solved it: I read it just fine how it is and it all made sense in the tests.

rubysolo:
I think these are fine.  The style guide discussions I have seen do indicate the left-alignment for multi-line pipelines, but also allow for single-line pipelines. (see https://github.com/niftyn8/elixir_style_guide/pull/33#issuecomment-72760481)

PR #140, opened 2016-03-03
Fix string warning messages in tests

dalexj:
switched all `@spec` definitions with `string` to `String.t` to fix warnings when tests are run. also stripped some whitespace in the files I was in.

The only other warnings are that `DNA`/`DNATest` are being defined 3 times for `hamming`, `rna_transcription`, and `nucleotide_count` ( dont know if this wants to be changed), and that the inspect protocol already been consolidated (which comes from the way the code is loaded after mix compiles the project)

### DIFF
--- a/exercises/acronym/example.exs
+++ b/exercises/acronym/example.exs
@@ -1,6 +1,6 @@
 defmodule Acronym do
-  
-  @spec abbreviate(string) :: String.t()
+
+  @spec abbreviate(String.t()) :: String.t()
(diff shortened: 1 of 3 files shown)


parkerl:
Sweet! :tada:

kytrinyx:
@parkerl I love the enthusiasm. Did you mean to merge this?

parkerl:
@kytrinyx doh! Yes but was waiting for CI and then I saw a squirrel.

PR #141, opened 2016-03-03
`mix test` runs on its own with no arguments

dalexj:
#### What's this PR do?
lets you run `mix test` instead of `EXERCISM_TEST_EXAMPLES=true mix test --include pending exercises/**/*_test.exs` and reduces boilerplate code inside of test files

(also loads code for tests ~3 seconds faster)

my editor also trimmed whitespace in every file I opened so a few extra changes in some files

#### Where should the reviewer start?
look at the `test_helper` changes and file relocation, and the `mix.exs` changes

#### Any background context you want to provide?
`test_helper` needed to be moved into the `exercises` directory so that `mix test` would auto look into the exercises for test files with the configuration inside `mix.exs`


parkerl:
@dalexj I think this is great. It is nice now that we have moved all the execises to `./exercises` that we can do this.

PR #144, opened 2016-03-12
Add hamming test

dalexj:
adding in a test to make sure people have implemented hamming correctly, since the current tests can pass even if you aren't comparing per index


parkerl:
Cool. :tada:

PR #148, opened 2016-03-21
Fix saddle points test

dalexj:
This saddle point test should not have the random space after a newline character.

I originally made this exercise by taking the ruby version's tests, which means this error is most likely on the ruby version as well, although string splitting is treated differently between ruby/elixir so it's not a problem there


parkerl:
Great thanks! :tada:

PR #149, opened 2016-03-22
Tests don’t test the order of keyword arguments

andrewtimberlake:
In Elixir keyword arguments are order dependent `Queen.new(white: {1,2}, black: {3,4})` is not the same as `Queen.new(black: {3,4}, white: {1,2})`.
Many solutions will assume they can pattern match on the keyword arguments.

### DIFF
--- a/exercises/queen-attack/example.exs
+++ b/exercises/queen-attack/example.exs
@@ -6,8 +6,12 @@ defmodule Queens do
   Creates a new set of Queens
   """
   @spec new(nil | list) :: Queens.t()
-  def new(white: same_loc, black: same_loc), do: raise ArgumentError
-  def new(white: white, black: black), do: %Queens{white: white, black: black}
+  def new(placement) do
+    white = placement[:white]
+    black = placement[:black]
+    if white == black, do: raise ArgumentError
+    %Queens{white: white, black: black}
+  end
   def new, do: %Queens{}
 
   @doc """
--- a/exercises/queen-attack/example.exs
+++ b/exercises/queen-attack/example.exs
@@ -5,7 +5,7 @@ defmodule Queens do
   @doc """
   Creates a new set of Queens
   """
-  @spec new(nil | list, nil | list) :: Queens.t()
+  @spec new(nil | tuple, nil | tuple) :: Queens.t()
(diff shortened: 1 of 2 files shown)


dalexj:
this seems correct, can you change the example to match so the tests pass?

something like:

```elixir
  def new(locations) do
    if locations[:black] == locations[:white], do: raise ArgumentError
    %Queens{white: locations[:white], black: locations[:black]}
  end
```

andrewtimberlake:
Sorry, I have updated the example and run the tests.

parkerl:
@andrewtimberlake thanks for raising this issue. I can get behind this change. However, if we are looking for order independence then I wonder if a `Map` might be a better choice as an argument? Or perhaps we should just pass two arguments?

```
Many solutions will assume they can pattern match on the keyword arguments.
```

They would be matching on the keyword list that is the single argument to the function.  We should not call it 'keyword arguments'. The test description needs to change.

Thanks

dalexj:
I like either 2 separate arguments/map

andrewtimberlake:
Keyword arguments seem to be the expected way of passing variable arguments to a function in Elixir
That being said, I think this function has two expected arguments so two separate named arguments would probably be better.
I'll make that change now.

andrewtimberlake:
I’ve made the change but not sure that `Queen.new({1,2}, {3,4})` is as expressive as `Quen.new(white: {1,2}, black: {3,4})`

parkerl:
@andrewtimberlake I'm going to merge this. I still think you could use a map and return some of the expressiveness you feel is lost. Feel free to submit another PR.

```
Keyword arguments seem to be the expected way of passing variable arguments to a function in Elixir
```

I feel I must point out that Elixir does not technically have keyword (otherwise known as "named") arguments. Rather, when the last argument to a function is a keyword list then you can omit the square brackets. Sorry, I am truly not trying to be harsh. I just thought you should know in case it comes up in other contexts. 

Thanks for the PR! :tada:

andrewtimberlake:
Thanks Lew. I know they're not named arguments and I do understand how they work. I should probably be more semantically correct when I describe things.

PR #151, opened 2016-03-23
Get rid of sum of multiples default arguments

### LINKED ISSUE #150: Pass explicit list of multiples in "Sum of Multiples" exercise rather than defaulting to 3 and 5
Hello, as part of https://github.com/exercism/x-common/issues/198 we'd like to make the sum of multiples exercise less confusing. Currently, the [README](https://github.com/exercism/x-common/blob/master/sum-of-multiples.md) specifies that if no multiples are given it should default to 3 and 5.

We'd like to remove this default, so that a list of multiples will always be specified by the caller. This makes the behavior explicit, avoiding surprising behavior and simplifying the problem.

**Please make sure this track's tests for the sum-of-multiples problem do not expect such a default**. Any tests that want to test behavior for multiples of [3, 5] should explicitly pass [3, 5] as the list of multiples.

After all tracks have completed this change, then https://github.com/exercism/x-common/pull/209 can be merged to remove the defaults from the README.

The reason we'd like this change to happen before changing the README is that it was very confusing for students to figure out the default behavior. It wasn't clear from simply looking at the tests that the default should be 3 and 5, as seen in https://github.com/exercism/exercism.io/issues/2654, so some had to resort to looking at the example solutions (which aren't served by `exercism fetch`, so they have to find it on GitHub). It was added to the README to fix this confusion, but now we'd like to be explicit so we can remove the default line from the README.

You can find the common test data at https://github.com/exercism/x-common/blob/master/sum-of-multiples.json, in case that is helpful.

dalexj:
closes #150

### DIFF
--- a/exercises/sum-of-multiples/example.exs
+++ b/exercises/sum-of-multiples/example.exs
@@ -2,11 +2,9 @@ defmodule SumOfMultiples do
 
   @doc """
   Adds up all numbers from 1 to a given end number that are multiples of the factors provided.
-
-  The default factors are 3 and 5.
   """
   @spec to(non_neg_integer, [non_neg_integer]) :: non_neg_integer
-  def to(limit, factors \\ [3, 5]) do
+  def to(limit, factors) do
     Enum.reduce(1..limit-1, 0, fn(n, acc) ->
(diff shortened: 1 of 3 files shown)


parkerl:
Great :tada:

PR #152, opened 2016-03-23
Fix map type and make it consistent

dantswain:
A few places were using `Map` as a type, which technically means the
atom `:'Elixir.Map'`.  The type should be `map`.  There was also some
inconsistency of `map` vs `map()`.  It seems that for the most part the
exercises aren't using `()` when unnecessary, so I changed the existing
instances of `map()` to just `map`.


parkerl:
Very nice! Thanks :tada:

PR #154, opened 2016-03-24
Add rail fence cipher exercise

jimmbraddock:
I don't know where place this exercise in config.json. Therefore put the exercise near atbash cipher.

### DIFF
--- a/exercises/rail-fence-cipher/example.exs
+++ b/exercises/rail-fence-cipher/example.exs
@@ -7,7 +7,7 @@ defmodule RailFenceCipher do
   def encode(str, rails) do
     str
     |> String.codepoints
-    |> fence(rails)
+    |> fill_fence(rails)
     |> Enum.join
   end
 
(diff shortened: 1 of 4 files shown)


parkerl:
@jimmbraddock very nice! I just have a couple small nitpicks.

exercises/rail-fence-cipher/rail_fence_cipher_test.exs
```diff
@@ -0,0 +1,91 @@
...
+defmodule RailFenceCipherTest do
+  use ExUnit.Case
+
+  #@tag :pending
+  test "test to encode with two rails" do
+    assert RailFenceCipher.encode("XOXOXOXOXOXOXOXOXO", 2) == "XXXXXXXXXOOOOOOOOO"
+  end
+
+  @tag :pending
+  test "test to encode with three rails" do
```
parkerl: All of the test descriptions begin with `test` which makes the test read "test test to encode with three rails". Remove `test` from the descriptions and perhaps consider rewording. (test "encoding with three rails"?)

exercises/rail-fence-cipher/example.exs
```diff
@@ -0,0 +1,47 @@
...
+         [String.at(str, Enum.find_index(pos, &(&1 == i))) | acc]
+       end)
+    |> Enum.reverse
+    |> Enum.join
+  end
+
+  defp fence(list_chars, rails) do
+    rail_indexes = Enum.concat(Enum.to_list(0..rails - 1), Enum.to_list(rails - 2..1))
+    fence = 0..rails - 1
+      |> Enum.reduce([], fn _, acc ->
```
parkerl: You are reducing into a list here. Could this be `Enum.map`?

exercises/rail-fence-cipher/example.exs
```diff
@@ -0,0 +1,47 @@
...
+  @doc """
+  Decode a given rail fence ciphertext to the corresponding plaintext
+  """
+  @spec decode(String.t, pos_integer) :: String.t
+  def decode(str, 1), do: str
+  def decode(str, rails) do
+    range = 0..String.length(str) - 1
+    pos = range |> Enum.to_list |> fence(rails)
+    range
+    |> Enum.reduce([], fn i, acc ->
```
parkerl: Like below, you are reducing into a list here. Could this be `Enum.map`?

parkerl:
@jimmbraddock this is a great exercise! Thanks you for submitting it. As I said earlier, there are just a few issues. First, can you scrub the trailing whitespace from all the files? Please update the test descriptions so they read more smoothly. Finally, `example.exs` could use just a little refactoring. In addition to my comments about `Enum.map`, perhaps consider pulling out a few more functions to help dry things up and clarify intent?

jimmbraddock:
I have tried clarify and divided on more functions, but just the same it is looking so-so.
Thank you for `Enum.map`. You really opened the my eyes of to this simple solution instead of bigger `Enum.reduce`.

dalexj:
I think the tests here should be consistent with which side has the expected value and which is the encode/decode function, some tests at the end swap the two

parkerl:
Yes as @dalexj points out, those last two tests should have the call to the function under test on the left and the expected output on the right. Thanks!

jimmbraddock:
Fixed.

exercises/rail-fence-cipher/rail_fence_cipher_test.exs
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  @tag :pending
+  test "decode with one rail" do
+    assert RailFenceCipher.decode("ABCDEFGHIJKLMNOP", 1) == "ABCDEFGHIJKLMNOP"
+  end
+
+  @tag :pending
+  test "decode letters of less than rails" do
+    assert "More rails than letters" == RailFenceCipher.decode("More rails than letters", 24)
```
dalexj: swap the order of these?

exercises/rail-fence-cipher/rail_fence_cipher_test.exs
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  @tag :pending
+  test "decode a more diverse text" do
+    msg = "The quick brown fox jumps over the lazy dog."
+    cipher = RailFenceCipher.encode(msg, 4)
+    assert msg == RailFenceCipher.decode(cipher, 4)
```
dalexj: same here

parkerl:
Thanks @jimmbraddock this looks good. 🎉 

Sorry for the delay on this.

PR #157, opened 2016-03-25
Add connect exercise

dalexj:


exercises/connect/example.exs
```diff
@@ -0,0 +1,71 @@
...
+      {x + 1, y},
+      {x - 1, y},
+      {x + 1, y - 1},
+      {x - 1, y + 1}
+    ]
+    |> Enum.filter(&valid_next_loc(&1, history, board))
+  end
+
+  defp valid_next_loc({x, y}, history, board) do
+    x >= 0 && y >= 0 && x < length(board) && y < byte_size(hd board) && !({x, y} in history)
```
parkerl: This is a long chain of conditions. Maybe reformat a bit? 

Also, no tests fail when I remove ` y < byte_size(hd board)`. Missing test?

FYI, I sometimes like to do these like this...

```
    Enum.all? [
      x >= 0,
      y >= 0,
      x < length(board),
      y < byte_size(hd board),
      !({x, y} in history)
    ]
```

It doesn't short circuit but I think it reads kinda nice. Not suggesting a change, just sharing.

parkerl:
@dalexj great job as usual. Just one comment for you to respond to. 

Would you mind if I email you about another matter?

Thanks!

exercises/connect/connect_test.exs
```diff
@@ -0,0 +1,106 @@
...
+ExUnit.configure exclude: :pending, trace: true
+
+defmodule ConnectTest do
+  use ExUnit.Case
+
+  def remove_spaces(rows) do
+    Enum.map(rows, &String.replace(&1, " ", ""))
+  end
+
+  @tag :pending
```
parkerl: Oh, we should also have one test where `@tag :pending` is commented out so the user has a place to start.

dalexj:
I've addressed the comments here, good catch on the first `@tag :pending`

feel free to email me - <96dalex@gmail.com>

parkerl:
Great! :tada:

PR #158, opened 2016-03-26
adds test for deep focus set_value to Zipper

CoderDennis:
In my second iteration of the Zipper exercise, I implemented it with only the `trail` property. This required me to update the tree within the set functions. Without this deep set_value test, I may not have realized that my `update_node` function needed to recursively travel up the tree.

http://exercism.io/submissions/271485ec894e41269b347db8ee162b0f

All of the other set_* tests are only modifying a 2nd level node, so adding one that modifies a deeper node seems valuable.


parkerl:
Thanks! :tada:

PR #160, opened 2016-03-26
Add diamond exercise

jimmbraddock:
I can't come up normal tests. Easest is to check summary diamond shape.

### DIFF
--- a/exercises/diamond/diamond.exs
+++ b/exercises/diamond/diamond.exs
@@ -1,6 +1,7 @@
 defmodule Diamond do
   @doc """
-  List the allergies for which the corresponding flag bit is true.
+  Given a letter, it prints a diamond starting with 'A',
+  with the supplied letter at the widest point.
   """
   @spec build_shape(char) :: String.t
   def build_shape(letter) do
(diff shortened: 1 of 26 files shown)


exercises/diamond/diamond.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Diamond do
+  @doc """
+  List the allergies for which the corresponding flag bit is true.
```
dalexj: forgot to change this

dalexj:
I think for tests it would be best to have only 3-4 tests that look like this one, but for different letters:

```elixir
  @tag :pending
  test "letters form a diamond shape" do
    shape = Diamond.build_shape(?E)
    assert shape == """
\s   A    
\s  B B   
\s C   C  
\sD     D 
 E       E
\sD     D 
\s C   C  
\s  B B   
\s   A    
    """
  end
```

jimmbraddock:
I used `\s`, because leading whitespace disappear and was `...A...\n` instead of `...A...\n`. Even Markdown removes the leading whitespaces. Or I can replace whitespace to another symbol. Or this problem of my editor.

parkerl:
@jimmbraddock could you rebase master onto your branch? Thanks!

jimmbraddock:
@parkerl , I rebased.

parkerl:
Thanks 🎉

PR #161, opened 2016-03-28
Comment out first pending tag in test

andrewtimberlake:
The first test in the test file should have the pending tag commented out.


parkerl:
@andrewtimberlake great catch! Thanks! :tada:

PR #167, opened 2016-04-07
Fix stubbed custom exceptions in forth.exs

### LINKED ISSUE #135: forth.exs template custom errors stubbed incorrectly
the `def exception` functions seem like they should be `def message` functions, which is the function used in the example

I dont know a ton on the working of defining custom errors in elixir so I cant give much more information.

dalexj:
closes #135


parkerl:
Great. :tada:

PR #170, opened 2016-04-09
Update queen attack spec for `new`

### LINKED ISSUE #169: queen-attack: Provided stub file doesn't match usage from test
The stub file only does provide a `Queens.new/1` which has `@spec new(nil | list)`, while the tests do call it either without any arguments or with two seperate arguments but not a list.

dalexj:
closes #169


parkerl:
:tada: Thanks!!!

PR #171, opened 2016-04-22
Add accumulate exercise

bernardoamc:


exercises/accumulate/accumulate.exs
```diff
@@ -0,0 +1,21 @@
...
+      iex> Accumulate.accumulate(fn(x) -> x * 2 end, [])
+      []
+
+      iex> Accumulate.accumulate(fn(x) -> x * 2 end, [1, 2, 3])
+      [2, 4, 6]
+
+  """
+
+  @spec accumulate(Enumerable.t, (any -> any)) :: list
+  def accumulate(fun, list) do
```
dalexj: most elixir functions like this are built to take in `list` as the first argument, like your `@spec` shows

bernardoamc: I agree, but I think it is a good exercise for newcomers to try to implement `map`. What do you think?

martinsvalin: `map` is implemented in ListOps. It also has a notice about not using any of the built-in Enum or List functions, which this exercise would also need, so that folks don't simply submit `def accumulate(list, fun), do: Enum.map(list, fun)`

bernardoamc: I totally misunderstood what @dalexj said. Should I reverse the args? ;)

bernardoamc:
@dalexj Sorry for the confusion, should be fixed now.

dalexj:
sweet, everything looks good to me

I'd say tiny anonymous functions like this are easier to read as `&( &1 * &1)` rather than `fn x -> x * x end` but both work

bernardoamc:
Yeah, I was tempted to write these functions with this notation, but since it is an exercise for beginners I avoided it. Don't know if I should though.

parkerl:
Great! Thanks 🎉

PR #172, opened 2016-04-23
Extra queen attack tests for drawing the board.

brianvanburken:
This adds two new tests for queens that are placed on the edges of the board. This would test scenarios that the other representation tests don't cover.

### DIFF
--- a/exercises/queen-attack/queen_attack_test.exs
+++ b/exercises/queen-attack/queen_attack_test.exs
@@ -78,6 +78,38 @@ _ _ _ _ _ _ _ _
   end
 
   @tag :pending
+  test "queen placed on the bottom right corner" do
+    queens = Queens.new({4, 3}, {7, 7})
+    board = String.strip """
+_ _ _ _ _ _ _ _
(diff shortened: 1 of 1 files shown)


parkerl:
@brianvanburken thanks for this PR. Can you help me understand how these two new specs help with this exercise? I'm trying to figure out how I could implement this exercise in such a way that the existing specs would pass while these new specs fail.

brianvanburken:
@parkerl sorry for the late response. I followed the exercise using the TDD principle and made the tests pass by outputting the queens only `"B "` and `"W "` (notice the spaces after the letter). This does not make the exercise complete. After adding these two tests I had to rethink my solution and came with a better one. Adding this PR would make sure other won't skip these cases. 

My solution was improved with these tests and can be seen in this commit: https://github.com/brianvanburken/exercism/commit/f1e62d9844ad766cf9420b0f0e2a10c27d8be5c4

Hopefully I've clarified my reason for this PR :)

parkerl:
@brianvanburken Thanks for the response! Merging 🎉

PR #173, opened 2016-04-23
Add binary-search exercise

bernardoamc:


exercises/binary-search/binary_search.exs
```diff
@@ -0,0 +1,24 @@
...
+      iex> BinarySearch.search([1, 3, 5], 2)
+      :not_found
+
+      iex> BinarySearch.search([1, 3, 5], 5)
+      {:ok, 2}
+
+  """
+
+  @spec search(Enumerable.t, integer) :: {:ok, integer} | :not_found
+  def search([], key) do
```
dalexj: probably make this first argument a variable name, leaving implementation up to the solver

exercises/binary-search/example.exs
```diff
@@ -0,0 +1,35 @@
...
+      {:ok, 2}
+
+  """
+
+  @spec search(Enumerable.t, integer) :: {:ok, integer} | :not_found
+  def search([], _key), do: :not_found
+  def search(list, key), do: _search(list, key, 0, length(list) - 1)
+
+  defp _search(list, _key, low, high) when high < low, do: :not_found
+  defp _search(list, key, low, high) do
```
dalexj: elixir functions like your `_search` are commonly named `do_{name}`

exercises/binary-search/binary_search.exs
bernardoamc: Makes sense, my mistake. ;)

exercises/binary-search/example.exs
bernardoamc: Really? Didn't know about this, thanks for the tip.

dalexj: you'll see it a lot if you read some of the elixir source, especially in the `Enum` module if you want to take a look: https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/enum.ex

bernardoamc: 👍 ❤️

bernardoamc:
@dalexj Done, thank you for reviewing the code.

parkerl:
@bernardoamc thanks for this! We probably need a test for an unsorted list. Perhaps raise an `ArgumentError`. Also, what about al list with duplicates? I'm not sure if that should raise or just return the index of the first one found. Probably the later, but we should still have the test I think. @dalexj thoughts?

dalexj:
looking at the ruby exercism, there's an `ArgumentError` for unsorted lists, and nothing regarding duplicates in the list.

duplicates are a little strange as they would return the first number that falls on a `length/2^n` index instead of the first one in the list. example: `[2, 2, 2]` would return index `1` since it looks at the middle first

bernardoamc:
@parkerl No problem, glad to help! I think it would be unexpected to `raise` in case of duplicates since the search will work as expected. I have mixed feelings regarding enforcing that the list should be sorted, but since the `ruby` exercise has this maybe we should implement it.

Something like: `raise ArgumentError, "expected list to be sorted"`

bernardoamc:
@parkerl @dalexj I implemented the check for unsorted lists, suggestions are welcome. :)

parkerl:
Thanks @bernardoamc. I'm good with this s it is. @dalexj any other comments?

dalexj:
:+1: looks good to me

PR #174, opened 2016-04-29
Add "change" exercise

bernardoamc:


bernardoamc:
The problem does not mention that we need to accept coins in any order, but since the [lua example](https://github.com/exercism/xlua/blob/master/exercises/change/change_spec.lua#L20) does it I chose to implement this feature.

dalexj:
I feel like it would be much nicer for this exercism to return a map of `coin_value` => `amount`

I know this is the version described in x-common, but maybe I can make an issue there and see what some ideas are?

bernardoamc:
I think they specified it this way so people would not be guided towards a solution involving a map, hash, table, or something similar. They can still use a map in their implementation though, so maybe this is more flexible?

dalexj:
it seems very strange for me to put the same data in different order to a function and get different results, when the order of the data isn't important (only to specify which coins exist)

bernardoamc:
I agree, if it was me I would sort the values and always return the coins in the same order, but it is better to keep the solution/tests compatible with other languages, right?

dalexj:
yes it would be best to keep consistent with other languages. that's why I had suggesting creating an issue inside x-common for it.

otherwise, looks good

kotp:
I think the tests should match, for the most part, but only because tests should cover the solution (much less so an implementation).  The returns should be exactly how you would expect to see them in Elixer.  The returns in Lua would be exactly what should be expected in Lua.  Perhaps the problem is in Lua, but that problem shouldn't carry over here, you are introducing the exercise here.  Let's get it right so that Lua can learn from it.

I did not see anything in the README that dictated this kind of return.  I don't think it is a x-common problem yet.

bernardoamc:
Looks like the tests with `arrays/lists` are the ones that don't drive exercists to a specific solution. If we use `hash/map/table`(pick your language) the implementation is pretty much specified.

kytrinyx:
> it seems very strange for me to put the same data in different order to a function and get different results, when the order of the data isn't important (only to specify which coins exist)

The order isn't important. If you give me a handful of coins in some order, I could probably end up giving you the change in any different order--but it's still the same change.

I think that the tests could be written to make sure that whatever the order people return, it doesn't matter.

bernardoamc:
> I think that the tests could be written to make sure that whatever the order people return, it doesn't matter.

Something like this?

```elixir
test "generates the same change given any coin order" do
  assert Change.generate(18, [1, 5, 10]) == {:ok, [1, 1, 3]}
  assert Change.generate(18, [10, 5, 1]) == {:ok, [1, 1, 3]}
end
```

kytrinyx:
That would assume that `Change.generate` always gives back an array that is sorted lowest to highest.

I don't know the language but something like

```
test "generates the same change given any coin order" do
  assert Change.generate(18, [1, 5, 10]) == {:ok, sort([1, 1, 3])}
  assert Change.generate(18, [10, 5, 1]) == {:ok, sort([1, 1, 3])}
end
```

bernardoamc:
Done.

The solution was simplified with these constraints.

dalexj:
I really think the return value of the generate here should be a map. a list sorted by number of coins gives us even less information about which coins are being used than before as well

kytrinyx:
I think a map would make sense also. That would remove the ordering issue, too (unless elixir maps are ordered).

bernardoamc:
Now it's returning a map, is this better than the previous solution?

kytrinyx:
Yeah, I like this better. @dalexj?

dalexj:
like it now :+1:. thanks for working with all the discussion

bernardoamc:
No problem, thank you for the feedback @dalexj @kytrinyx and @kotp ❤️

exercises/change/change_test.exs
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  @tag :pending
+  test "generates the correct change when multiple coin types are needed" do
+    change = %{1 => 3, 5 => 1, 10 => 1}
+    assert Change.generate(18, [1, 5, 10]) == {:ok, change}
+  end
+
+  @tag :pending
+  test "returns error when it is not possible to generate change" do
```
parkerl: Should be "returns :error when it is not possible to generate change" to be consistent with the first test.

parkerl:
Great discussion. I added one small nitpick, but other than that this looks good to me.

bernardoamc:
Thanks @parkerl ! Should be fixed now. :)

parkerl:
Thanks :tada:

PR #177, opened 2016-04-30
accumulate: Fix typespec for accumulate/2 function.

### LINKED ISSUE #175: [accumulate]: typespec vs. example implementation
According to the typespec given in the exercise-stub, I'd expect `Accumulate.accumulate %{foo: :bar}, &(&1)` beeing a valid call of the function, but such is not caovered by the tests, as well as the example implementation does as well only handle lists.

So either the `@spec` is wrong and should be changed towards a more proper `@spec accumulate(list, (any -> any)) :: list` or the example is wrong and another clause should be added (`def accumulate(anything), fun) when not is_list(anything), do: anything |> Enum.into([]) |> accumulate(fun)`) as well as some tests that do test some other `Enumable.t`s that are not lists.

bernardoamc:
fixes #175.

### DIFF
--- a/exercises/accumulate/accumulate.exs
+++ b/exercises/accumulate/accumulate.exs
@@ -15,7 +15,7 @@ defmodule Accumulate do
 
   """
 
-  @spec accumulate(Enumerable.t, (any -> any)) :: list
+  @spec accumulate(list, (any -> any)) :: list
   def accumulate(list, fun) do
   end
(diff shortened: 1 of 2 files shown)


bernardoamc:
@NobbZ Thank you for catching this.

parkerl:
Thanks. 🎉

PR #179, opened 2016-05-13
add luhn exercise

dalexj:


exercises/luhn/luhn_test.exs
```diff
@@ -0,0 +1,44 @@
...
+
+defmodule LuhnTest do
+  use ExUnit.Case
+
+  test "checksum" do
+    assert Luhn.checksum("4913") == 22
+  end
+
+  @tag :pending
+  test "checksum again" do
```
bernardoamc: Should we move these "again" tests into a single test or there is some rule in exercism that we should not have multiple assertions per test? I'm asking this because we are not testing a new scenario here, so it would make sense to have just a `checksum` test with both assertions. What do you think? :)

bernardoamc:
Nice @dalexj ! ❤️

exercises/luhn/luhn_test.exs
dalexj: I'm a fan of having a single case per `test`, which is typically equal to having 1 single assert per test, and see no improvement to compact them at all other than just to have them compacted

I know I've caught myself with errors on the `again` test from some exercisms that never showed up on the first test

bernardoamc: No strong feelings regarding merging tests, just discussing possibilities. 👍 
 Nowadays I'm experimenting with this approach when the new test is not about a new scenario or edge case.

bernardoamc:
Looks good to me.

parkerl:
Nice. 

Is there a reason to pass strings rather than integers?

Do we need a test with an empty string as input? A test with bad input such as a string with characters that are non-numeric?

dalexj:
using a string for input I think it made more sense for things like the `create` function that append to a string rather than appending to a number, which would seem strange.

parkerl:
@dalexj cool. I guess I would ask that if you were to pass numbers in the tests then the submitter could decide if they want to cast to a string in their implementation or something else. For instance, you might be able to break the number into a list of digits to make a good solution. Have you seen `Integer.digits/2` and `Integer.undigits/2`?  Just a thought.

dalexj:
I did not know about `undigits`, has to be one of the coolest function names I've seen.

anyway, I'm down for either a string input or integer input. I'll leave the final say to you and I can swap out the tests/example to match.

parkerl:
Ok @dalexj I'm good with this as is. Thanks!

PR #182, opened 2016-05-23
Fix deprecation of \x{H*} in ForthTest

martinsvalin:
I'm running my Exercism tests on Elixir 1.3-dev, and got a deprecation warning in ForthTest:

```
warning: \x{H*} inside strings/sigils/chars is deprecated, please use \xHH (byte) or \uHHHH (codepoint) instead
```

The fix isn't exactly an improvement in terms of readability, but it satisfies the compiler.
I've tested this with Elixir 1.3-dev (current master) and 1.2.5.


parkerl:
Thanks :tads:

PR #184, opened 2016-05-25
add beer-song

devonestes:
I'm just dibsing this exercise for now and I'll add the implementation and test suite soon.

### DIFF
--- a/config.json
+++ b/config.json
@@ -18,6 +18,7 @@
     "accumulate",
     "hamming",
     "phone-number",
+    "beer-song",
     "grade-school",
     "leap",
     "etl",
--- a/exercises/beer-song/beer_song.exs
+++ b/exercises/beer-song/beer_song.exs
@@ -1,0 +1,17 @@
+defmodule BeerSong do
+  @doc """
+  Get a single verse of the beer song
+  """
+  @spec verse(integer) :: String.t
+  def verse(number) do
+    # Your implementation here...
+  end
+
+  @doc """
+  Get the entire beer song for a given range of numbers of bottles.
+  """
+  @spec lyrics(Range.t) :: String.t
+  def lyrics(range) do
+    # Your implementation here...
+  end
+end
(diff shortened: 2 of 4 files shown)


config.json
```diff
@@ -56,7 +56,8 @@
     "dot-dsl",
     "custom-set",
     "nth-prime",
-    "forth"
+    "forth",
+    "beer-song"
```
dalexj: I dont think this exercise belongs at the end of this list

exercises/beer-song/beer_song.exs
```diff
@@ -0,0 +1,9 @@
+defmodule BeerSong do
+  def verse(number) do
+    # Your implementation here...
+  end
+
+  def lyrics(range) do
+    # Your implementation here...
+  end
```
dalexj: can you add some typespecs and docs?

exercises/beer-song/example.exs
```diff
@@ -0,0 +1,23 @@
+defmodule BeerSong do
+  def verse(number) when number == 1 do
```
dalexj: here's a simpler way to do this:

```elixir
def verse(1) do
  ..
end
```

config.json
devonestes: Does that list determine the order in which people get the exercises when the call `exercism fetch elixir`? If so, it totally doesn't!

exercises/beer-song/beer_song.exs
devonestes: Sure thing! Should they be doc tests, or just plan ol' docs?

exercises/beer-song/example.exs
devonestes: Brilliant - thanks for this!

config.json
kytrinyx: Yeah, the `config.json` determines the order in which people receive exercises by default. (We really need to make that easier to understand/discover!)

parkerl:
Thanks! 🎉

PR #185, opened 2016-05-28
Add flatten-array exercise

bernardoamc:


config.json
```diff
@@ -11,6 +11,7 @@
     "run-length-encoding",
     "list-ops",
     "sublist",
+    "flatten-array",
```
dalexj: I think maybe putting this before sublist makes slightly more sense, it'd be a good warmup for sublist

thoughts?

bernardoamc: Makes sense since this is easier than the sublist problem.

exercises/flatten-array/flatten_array_test.exs
```diff
@@ -0,0 +1,34 @@
...
+    assert Flattener.flatten([1, 2, 3]) ==  [1, 2, 3]
+  end
+
+  @tag :pending
+  test "flattens an empty nested list" do
+    assert Flattener.flatten([[]]) ==  []
+  end
+
+  @tag :pending
+  test "flattens a not empty nested list" do
```
parkerl: This is probably just "flattens a nested list" as this is kind of the happy path test.

exercises/flatten-array/flatten_array_test.exs
```diff
@@ -0,0 +1,34 @@
...
+    assert Flattener.flatten([1,[2,[3],4],5,[6,[7,8]]]) == [1, 2, 3, 4, 5, 6, 7, 8]
+  end
+
+  @tag :pending
+  test "removes nil from list" do
+    assert Flattener.flatten([1, nil, 2]) ==  [1, 2]
+  end
+
+  @tag :pending
```
parkerl: Do we need a test showing that it removes `nil` from a nested list? I see the following test where all entries are removed but it might be nice to show where a sublist is compacted as well.

exercises/flatten-array/flatten_array_test.exs
bernardoamc: Makes sense. 👍

exercises/flatten-array/flatten_array_test.exs
bernardoamc: The line `assert Flattener.flatten([1,[2,[3],4],5,[6,[7,8]]]) == [1, 2, 3, 4, 5, 6, 7, 8]` doesn't cover this case?  I think I didn't get what you wanted to say.

parkerl: I was thinking something like `[1,[2,nil,4],5]`. It might be helpful for folks to have something between `[1, nil, 2]` and the next test where everything is nil all the way down.

bernardoamc: I see! Good idea, going to add this one.

bernardoamc: @parkerl Done, thanks for the feedback. =]

martinsvalin:
Sometimes, newcomers to Elixir see the `[]` syntax and think it's the Array data type from their previous language, when in fact it's called a List and is a linked list data type. To avoid this confusion, I think we should avoid using "Array" to refer to a List. 

The implementation and the tests only mention lists, but the file name is `flatten_array`. Can we rename the files?

bernardoamc:
@martinsvalin Good question, I really don't know! I think the name must match the exercise from the [x-common](https://github.com/exercism/x-common) repository. I don't know if it is possible to create an alias for an exercise.

parkerl:
@bernardoamc this looks good. Sorry for the delay. However, @martinsvalin makes a very good point. @kytrinyx thoughts on this?

kytrinyx:
I think we should probably come up with a name that is implementation-unspecific so that we can use the exercise either with an array or a list depending on what makes sense for the language.

Any suggestions for a better exercise name?

bernardoamc:
> Any suggestions for a better exercise name?

I think `flatten_list` would be good enough since people know what a list is and how to implement it in his/her language. `flatten_array` as pointed above is language specific.

martinsvalin:
I think it's hard to find a word for a collection/list/array/enumerable/vector/whatever that isn't somehow tied up in at least one language's standard vocabulary, and thus given a more implementation specific meaning in that language. I'd say `flatten-list` is pretty general. It's used as the implementation name for the linked-list data structure in a lot of languages, and most people practicing those languages know it's not synonymous with the data structure.

I'd also consider just `flatten` or `flattener` or some variation.

But just to be clear: do we intentionally want to have the same names on exercises implemented for different languages? Since a language can have language-specific exercises, I'm guessing there's no technical limitation on naming a common exercise different in one language track.

kytrinyx:
> I'd also consider just flatten or flattener or some variation.

Agreed. I'd like to avoid data structure names where possible.

> But just to be clear: do we intentionally want to have the same names on exercises implemented for different languages? 

If it's basically the same exercise, then I think it's worth having the same name. This makes it easier to compare implementations (and update them when someone discovers a good edge case in one language that applies to a lot of other languages).

> I'm guessing there's no technical limitation on naming a common exercise different in one language track.

No, not directly, though we'd end up with copy/paste stuff in x-common, since the name of the exercise in the language track needs to have a corresponding thing in x-common with the same name, otherwise it barfs.

kytrinyx:
So: `flattener`?

parkerl:
+1

kytrinyx:
OK, cool. So I think we need to

- [ ] duplicate the exercise files here in x-common (so we can migrate individual language tracks independently of one another
- [ ]  create a blazon issue that can be submitted to all the languages that have implemented `flatten_array` (javascript, kotlin, ruby) to describe how to migrate the problem. I think this requires
    - renaming the problem directory
    - renaming the files
    - updating any code that refers to the old name
    - updating any comments or documentation that refer to the old name
- [ ] make sure we have a rake task that can be run in production to migrate existing solutions to the new name as soon as we deploy the migrated (see below)

We used to have a rake task that kind of did this, but it was all hard coded:

```

    desc "migrate deprecated problems"
    task :deprecated_problems do
      require 'bundler'
      Bundler.require
      require_relative '../exercism'
      {
        'point-mutations' => 'hamming'
      }.each do |deprecated, replacement|
        UserExercise.where(slug: deprecated).each do |exercise|
          if UserExercise.where(slug: replacement, user_id: exercise.user_id, language: exercise.language).count > 0
            next
          end

          exercise.slug = replacement
          exercise.save
          exercise.submissions.each do |submission|
            submission.slug = replacement
            submission.save
          end
          ACL.authorize(exercise.user, exercise.problem)
        end
      end
    end
```

It would be nice if we could have this task, but it would take two command-line arguments, one for the track_id and two from the problem slugs (src, dest).

parkerl:
@kytrinyx should this be merged and a separate issue opened for the rename? This PR has been siting idle as none of us have been able to get to the work you outlined above.

kytrinyx:
Ah, yeah, I think you're right: let's get this merged, and all the fixes can come later.

bernardoamc:
Ouch, sorry guys, totally forgot about this exercise.

Let me resolve the conflicts.

parkerl:
Thanks 🎉

PR #186, opened 2016-05-28
Correcting Luhn test case to be a valid number.

jtharris:
The test case for a valid number was incorrect (the checksum == 41)


kytrinyx:
Did this turn out to not be a bug after all?

jtharris:
Silly mistake on my part.  The exercise is correct as it is - sorry for the
confusion!

kytrinyx:
It's all good, just making sure we didn't miss something (or scare you off or anything)!

PR #187, opened 2016-05-29
Adding more documentation for contributers

devonestes:
There was some question about how to make the `config.json` file a bit more
discoverable, so I figured documentation was a good place to start with that.
This way people know what that file is for, know that the order matters, and
also know to add their new exercise to that file when they are adding a new one
to the repo.

This addresses [this comment](https://github.com/exercism/xelixir/pull/184#discussion_r64992425) by @kytrinyx.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,11 +11,11 @@ The exercises currently target Elixir 1.2. Detailed installation instructions ca
 
 Thank you so much for contributing! :tada:
 
-Please start by reading the general Exercism [contributing guide](https://github.com/exercism/x-api/blob/master/CONTRIBUTING.md#the-exercise-data).
-
 We welcome pull requests that provide fixes and improvements to existing exercises. If you're unsure, then go ahead and open a GitHub issue, and we'll discuss the change.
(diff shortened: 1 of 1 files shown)


kytrinyx:
We have a Contributor guide that we try to point people to. What do you think we can do to make the guide easier to understand / the config thing more discoverable there? https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md

Is the problem that people don't read the guide (perhaps because it's so long)?

devonestes:
I think the problem is that before now I didn't know that guide existed. In looking at the README for this repo, my eye jumped to the bulleted points on how to contribute, and the link to that file wasn't in there. I'm thinking maybe if that's just the first bullet point then it would be easier to see? Once I saw that file then I saw the bit about the `config.json`, but I think the problem is getting folks to see that file.

kytrinyx:
Yeah, making it the first bullet point would probably be a good idea!

parkerl:
Thanks 🎉

PR #188, opened 2016-05-29
adding bowling

devonestes:
Dibsing another new exercise - bowling!

### DIFF
--- a/exercises/bowling/bowling_test.exs
+++ b/exercises/bowling/bowling_test.exs
@@ -1,0 +1,283 @@
+if !System.get_env("EXERCISM_TEST_EXAMPLES") do
+  Code.load_file("bowling.exs")
+end
+
+ExUnit.start
+ExUnit.configure exclude: :pending, trace: true
+
+defmodule BowlingTest do
+  use ExUnit.Case
+
+  defp roll_reduce(game, rolls) do
+    Enum.reduce(rolls, game, fn(roll, game) -> Bowling.roll(game, roll) end)
+  end
+
+  test "can score all 0s" do
+    game = Bowling.start
+    rolls = [0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0]
+    game = roll_reduce(game, rolls)
+    assert Bowling.score(game) == 0
+  end
+
+  @tag :pending
+  test "can score an open frame" do
+    game = Bowling.start
+    rolls = [3, 4,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0]
+    game = roll_reduce(game, rolls)
+    assert Bowling.score(game) == 7
+  end
+
+  @tag :pending
+  test "can score multiple frames" do
+    game = Bowling.start
+    rolls = [3, 4,
+             2, 3,
+             5, 2,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0]
+    game = roll_reduce(game, rolls)
+    assert Bowling.score(game) == 19
+  end
+
+  @tag :pending
+  test "can score a game with all single pin rolls" do
+    game = Bowling.start
+    rolls = [1, 1,
+             1, 1,
+             1, 1,
+             1, 1,
+             1, 1,
+             1, 1,
+             1, 1,
+             1, 1,
+             1, 1,
+             1, 1]
+    game = roll_reduce(game, rolls)
+    assert Bowling.score(game) == 20
+  end
+
+  @tag :pending
+  test "can score a game with all open frames" do
+    game = Bowling.start
+    rolls = [3, 6,
+             3, 6,
+             3, 6,
+             3, 6,
+             3, 6,
+             3, 6,
+             3, 6,
+             3, 6,
+             3, 6,
+             3, 6]
+    game = roll_reduce(game, rolls)
+    assert Bowling.score(game) == 90
+  end
+
+  @tag :pending
+  test "can score a game with a strike not in the last frame" do
+    game = Bowling.start
+    rolls = [10,
+             5, 3,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0]
+    game = roll_reduce(game, rolls)
+    assert Bowling.score(game) == 26
+  end
+
+  @tag :pending
+  test "can score a game with a spare not in the last frame" do
+    game = Bowling.start
+    rolls = [5, 5,
+             3, 4,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0]
+    game = roll_reduce(game, rolls)
+    assert Bowling.score(game) == 20
+  end
+
+  @tag :pending
+  test "can score a game with multiple strikes in a row" do
+    game = Bowling.start
+    rolls = [10,
+             10,
+             10,
+             5, 3,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0]
+    game = roll_reduce(game, rolls)
+    assert Bowling.score(game) == 81
+  end
+
+  @tag :pending
+  test "can score a game with multiple spares in a row" do
+    game = Bowling.start
+    rolls = [5, 5,
+             3, 7,
+             4, 1,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0]
+    game = roll_reduce(game, rolls)
+    assert Bowling.score(game) == 32
+  end
+  
+  @tag :pending
+  test "fills out the last frame when there is a strike" do
+    game = Bowling.start
+    rolls = [0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             10,
+             7, 1]
+    game = roll_reduce(game, rolls)
+    assert Bowling.score(game) == 18
+  end
+
+  @tag :pending
+  test "fills out the last frame when there is a spare" do
+    game = Bowling.start
+    rolls = [0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             9, 1,
+             7]
+    game = roll_reduce(game, rolls)
+    assert Bowling.score(game) == 17
+  end
+
+  @tag :pending
+  test "allows filled balls to be strikes" do
+    game = Bowling.start
+    rolls = [0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             0, 0,
+             10,
+             10,
+             10]
+    game = roll_reduce(game, rolls)
+    assert Bowling.score(game) == 30
+  end
+
+  @tag :pending
+  test "scores a perfect game correctly" do
+    game = Bowling.start
+    rolls = [10,
+             10,
+             10,
+             10,
+             10,
+             10,
+             10,
+             10,
+             10,
+             10,
+             10,
+             10]
+    game = roll_reduce(game, rolls)
+    assert Bowling.score(game) == 300
+  end
+
+  @tag :pending
+  test "does not allow negative rolls" do
+    game = Bowling.start
+    assert Bowling.roll(game, -1) == {:error, "Pins must have a value from 0 to 10"}
+  end
+
+  @tag :pending
+  test "does not allow rolls greater than a strike" do
+    game = Bowling.start
+    assert Bowling.roll(game, 11) == {:error, "Pins must have a value from 0 to 10"}
+  end
+  
+  @tag :pending
+  test "does not allow consecutive rolls to be greater than a strike" do
+    game = Bowling.start
+    game = Bowling.roll(game, 5)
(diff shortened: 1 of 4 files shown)


devonestes:
This is ready for review now. I feel pretty good about the tests, but I feel like my example implementation is terrible! I'd love any feedback that anyone might have on how to make it better (or would it be better to save that for the site itself?)

Specifically, I'm using `cond` in several places, and I know that's a sort of anti-pattern in Elixir, but coming from Ruby the Elixir best practices still aren't really solid for me. Any guidance on how to refactor that would be much appreciated!

dalexj:
I don't think you really need the entire agent/pid piece, makes this much more complicated in my opinion

exercises/bowling/bowling_test.exs
```diff
@@ -0,0 +1,151 @@
...
+    game = Bowling.new
+    rolls = [10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10]
+    roll_from_array(game, rolls)
+    assert Bowling.score(game) == 300
+  end
+
+  @tag :pending
+  test "does not allow negative rolls" do
+    game = Bowling.new
+    assert Bowling.roll(game, -1) == "Pins must have a value from 0 to 10"
```
dalexj: I'd have these errors returned as `{:error, "message"}`

exercises/bowling/example.exs
```diff
@@ -0,0 +1,106 @@
...
+        spare(current_frame) ->
+          10 + Enum.at(next_frame, 0)
+        true ->
+          sum_list(current_frame)
+      end
+    end)
+  end
+
+  defp strike(frame) do
+    Enum.at(frame, 0) == 10
```
dalexj: would this work? `frame == [10]`

exercises/bowling/example.exs
```diff
@@ -0,0 +1,106 @@
...
+  end
+
+  defp spare(frame) do
+    sum_list(frame) == 10
+  end
+
+  defp sum_list(frame) do
+    Enum.reduce(frame, fn(score, acc) -> acc + score end)
```
dalexj: `Enum.sum`

exercises/bowling/example.exs
```diff
@@ -0,0 +1,106 @@
...
+      |> score_frames
+      |> sum_list
+  end
+
+  defp score_frames(scores) do
+    Enum.map((0..9), fn(idx) ->
+      current_frame = Enum.at(scores, idx)
+      next_frame = Enum.at(scores, idx + 1, [0,0])
+      cond do
+        strike(current_frame) ->
```
dalexj: probably make these boolean functions with a `?` at the end

exercises/bowling/example.exs
```diff
@@ -0,0 +1,106 @@
...
+          if strike(next_frame) do
+            sum_list(current_frame) + sum_list(next_frame) + (Enum.at(scores, idx + 2, [0, 0]) |> Enum.at(0))
+          else
+            sum_list(Enum.at(scores, idx)) + sum_list(Enum.at(scores, idx + 1, [0,0]))
+          end
+        spare(current_frame) ->
+          10 + Enum.at(next_frame, 0)
```
dalexj: `hd(list)` will return the first element

exercises/bowling/example.exs
```diff
@@ -0,0 +1,106 @@
...
+        true ->
+          sum_list(current_frame)
+      end
+    end)
+  end
```
dalexj: I'd definitely split this out into several functions, probably one to calculate strike/spare/other

exercises/bowling/example.exs
```diff
@@ -0,0 +1,106 @@
...
+
+  defp get_game_info(game) do
+    Agent.get(game, fn(game_struct) -> game_struct end)
+  end
+
+  defp parse_scores(scores) do
+    scores
+      |> score_frames
+      |> sum_list
```
dalexj: I've always seen `|>` indented the same level of the beginning of the pipeline unless theres a variable being set

```elixir
var
|> fun
|> fun

result = var
  |> fun
  |> fun
```

exercises/bowling/example.exs
```diff
@@ -0,0 +1,106 @@
+defmodule BowlingGame do
+  defstruct scores: [[0,0], [0,0], [0,0], [0,0], [0,0], [0,0], [0,0], [0,0], [0,0], [0,0], [0,0], [0,0]], current_frame: 1, roll_in_frame: 1
```
dalexj: `List.duplicate`

devonestes:
@dalexj Thanks for the feedback! Do you think we can ditch the Agent and just pass around the struct instead? It felt like a nice API for the tests to not have to pass around the struct, but I might be over thinking it. I mean, no one's really going to use this program after all!

kotp:
The example file is just proof of concept, so doesn't have to be very solid at all.

The tests on the other hand are presented and the README is presented, so these are the things that should be exemplar.  Like you said @devonestes the example is not really even going to be looked at, except to know that it is solvable in the language.

dalexj:
I'd take a look at the zipper exercise tests. It's a very functional way to write these kinds of "objects", which doesn't use the `Agent` module at all. I think you should try to write this without `Agent`

devonestes:
@dalexj After refactoring it, I agree, the `Agent` was unnecessary. I like how it looks now(ish). If there's anything else you think needs to be updated, let me know!

exercises/bowling/bowling_test.exs
```diff
@@ -0,0 +1,152 @@
...
+  Code.load_file("bowling.exs")
+end
+
+ExUnit.start
+ExUnit.configure exclude: :pending, trace: true
+
+defmodule BowlingTest do
+  use ExUnit.Case
+
+  def roll_from_array(game, []) do game end
```
bernardoamc: Usually we put helper functions at the end of the test file (in Ruby), I don't know if Elixir has the same practice.

parkerl: Actually in ExUnit I more often see helper functions at the top. See https://github.com/elixir-lang/ecto/blob/master/test/ecto/adapters/postgres_test.exs and other ecto tests for examples.

parkerl: That said, the helper functions are usually private.

bernardoamc: Good to know, thanks @parkerl ! =]

devonestes: Thanks for the knowledge @parkerl! I've just updated the test suite to make those helper functions private.

exercises/bowling/bowling_test.exs
```diff
@@ -0,0 +1,152 @@
...
+ExUnit.start
+ExUnit.configure exclude: :pending, trace: true
+
+defmodule BowlingTest do
+  use ExUnit.Case
+
+  defp roll_from_array(game, []) do game end
+
+  defp roll_from_array(game, rolls) do
+    roll_from_array(Bowling.roll(game, hd(rolls)), tl(rolls))
```
dalexj: pattern matching for rolls here would be nice: `[h | t]` instead of `rollls` in the argument

devonestes: good point! I've made the change.

parkerl:
@devonestes thanks for all your hard work on this. I have a number of questions.

First, I'm wondering if we should really have `new/0`? While there are a few examples of this function being used to return a struct from the module in Elixir libraries, they are few and far between. It tends to pull one toward an OO leaning implementation. I might feel differently if `new` took arguments that were needed to construct the struct as in the Queen Attack exercise. I suggest either simply passing `%BowlingGame{}` directly to the relevant functions or perhaps defining `roll/1` that takes the score of a roll and returns the game struct to be used on subsequent rolls.

I also wonder if we don't need more direct testing of `roll/2`? It is only tested indirectly through `score/1` with the exception of the error cases. I know if I were working on this exercise the first thing I would do is write tests for `roll/2` on my own so that I could more incrementally drive out the exercise.

That brings me to `roll_from_array/2`. First off, what you are passing to this function is a List so the function is misnamed. Second, I think it obfuscates what is going on in the tests. Not only do I have to figure out what each test is doing, but I also have to figure this function when doing the exercise. It looks like your implementation always returns the updated game struct from the happy path of `roll/2`? This is where you could make nice use of the `|` operator.  Something like...

```
game
|> Bowling.roll(10)
|> Bowling.roll(10)
#...etc
```

You will would have to work on handing the error cases a bit. Perhaps they should raise?


Thanks again! We look forward to your next iteration.

devonestes:
@parkerl Thanks for the great feedback! My Elixir code is definitely still rather tinged with Ruby OO tendencies, so I'm always happy to hear of other, more idiomatic ways of solving problems in Elixir.

The reason that I went with `Bowling.new` was to leave it up to the implementer how they wanted to manage the state here. For example, in my first implementation of this I was using `Agent` and the `game` variable referred to a `PID`. I think if we just give the user a struct with set fields it boxes them into a rather specific implementation and kind of gives them too much rather than letting them come up with a good implementation on their own (or through feedback on their iterations on the site). Basically the tradeoff I made was flexibility of implementation for handling the state of the module versus having a more directly functional approach. I think implementing `roll/1` will still give that flexibility without having to use `new`. I'll play around with that and see how I like it. I haven't really read enough Elixir code to know if that's idiomatic, though - is that a common practice?

Re: testing `roll/2`, I think there's the same tradeoff to be made. To accurately test that I think we'd have to infer a rather specific implementation. My first thought on testing it would look something like the following:

```elixir
test "can score a single roll" do
  game = Bowling.new
  game = Bowling.roll(game, 1)
  assert game.rolls == [1]
end
```

That would significantly reduce the number of possible implementations that a user can have since it locks them into a bunch of decisions. Is there a more flexible way that you can think of to test that function more directly?

Re: the error cases, I waffled back and forth a bunch on if they should raise or if they should return a tuple. What would your preference be? I see the benefits for both, but I thought that if someone did try and retrieve the score too early then it shouldn't warrant raising, and thus for consistency the errors for `roll/2` shouldn't raise either. After the feedback from @dalexj, I rather liked returning the tuple of `{:error, msg}`. That felt really idiomatic to Elixir for me, but I'm happy to change it to raising if you like!

You're totally right that `roll_from_array/2` is misnamed - more Ruby sneaking in! Thanks for keeping me honest 😉 . I avoided the pipe operator because some of the lists of rolls are rather long, and I didn't really want to have 20 lines in each test with all the rolls. That function came mostly out of laziness on my part, but if you think that it would be more helpful for the user to see each roll written out with the pipe then I'm happy to change it.

Thanks again for the great feedback - it's super helpful and much appreciated!

parkerl:
@devonestes been thinking quite a bit about this one (while out sick). I think you make good points above. Let me gather my thoughts a little more and get back to you. Sorry for the delay.

devonestes:
@parkerl No problem - take your time! Sorry to hear you're not feeling well 😷

parkerl:
I find the README for this exercise a bit over prescriptive. Here are my thoughts given those constraints. Now that I look at it, I think `roll/1` was a bad idea on my part. It throws off the list of rolls so it is always `n-1` for the number of rolls needed for a complete game. What if you went with `start/0` instead of `new/0`? That is perhaps a bit more erlang otp/elixir style. 

Also, `roll_from_list/2` is better, though putting the data type in the name of a function is a tiny code smell. This operation is also a reduction and could be handled with `Enum.reduce/3` rather than two function bodies. 

I will put a couple more housekeeping items inline. 

Thanks! I think this is close.

exercises/bowling/bowling_test.exs
```diff
@@ -0,0 +1,132 @@
...
+  end
+
+  test "can score all 0s" do
+    game = Bowling.roll(0)
+    rolls = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
+    game = roll_from_list(game, rolls)
+    assert Bowling.score(game) == 0
+  end
+
+  test "can score an open frame" do
```
parkerl: This and subsequent tests should have `@tag :pending`

exercises/bowling/bowling_test.exs
```diff
@@ -0,0 +1,132 @@
...
+
+  defp roll_from_list(game, []) do game end
+
+  defp roll_from_list(game, [roll | rolls]) do
+    roll_from_list(Bowling.roll(game, roll), rolls)
+  end
+
+  test "can score all 0s" do
+    game = Bowling.roll(0)
+    rolls = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```
parkerl: Wondering if it would be nice/helpful to format this to display the frames? 

(Assuming `roll/1` changes per my comment it could look like this?)
```
rolls = [
 0, 0,
 0, 0,
 0, 0,
 0, 0,
 0, 0,
 0, 0,
 0, 0,
 0, 0,
 0, 0,
 0, 0
]
```

kytrinyx:
> I find the README for this exercise a bit over prescriptive.

@parkerl I've opened an issue about this: https://github.com/exercism/discussions/issues/44

devonestes:
@parkerl Good call about `start/0` - I like it! Thanks for the input about `roll_from_list` as well - I didn't know that was a code smell in Elixir, but I do now!

I'll address those points and the other ones you inlined and get something back up today.

devonestes:
OK, I think I got the point about `roll_from_array`, but if not please let me know!

exercises/bowling/bowling_test.exs
```diff
@@ -0,0 +1,283 @@
...
+end
+
+ExUnit.start
+ExUnit.configure exclude: :pending, trace: true
+
+defmodule BowlingTest do
+  use ExUnit.Case
+
+  defp roll_reduce(game, rolls) do
+    Enum.reduce(rolls, game, fn(roll, game) -> Bowling.roll(game, roll) end)
```
parkerl: Nailed it!

parkerl:
I think this looks really nice now! Great job. I will wait a few to merge to see if others have any final comments.

parkerl:
Thanks @kytrinyx I will take a look.

parkerl:
Thanks! :tada:

kotp:
This is in, and I would say you scored a three hundred and one!  (I have never heard of anyone scoring a three hundred and lost!)

PR #196, opened 2016-06-14
Fix Wordy typespec

davinlagerroos:
Sorry, I had the wrong typespec in the wordy.exs and example.exs files when I submitted the exercise. Some answers are negative.


parkerl:
Good catch! Thanks! 🎉

PR #199, opened 2016-06-19
acronym.exs whitespace

henrik:
Remove trailing whitespace; make method body be two-space indent.


parkerl:
Thanks! Nice catch! 🎉

PR #200, opened 2016-06-20
Suggesting a different API for the Hamming exercise

jedschneider:
While doing this problem, I originally missed 'un-pending' the final test. When I did, I realized that there were two types (Integer and Nil) possibly returning from the `hamming_distance/2` function. I'd like to propose a single type return with the idiomatic `:ok` `:error` tuple pattern for handling un-resolved use cases. If a hamming distance is undefined for strands of different lengths (and not considered `0`), it seems returning the error tuple is more informative than a `nil` return and ensures a consistent tuple data type being returned from the query.

This Pull Request:

- includes an updated example solution using the same method solutions as the original exercise
- re-organizes the tests and adds an additional test to direct the solution earlier in the exercise

I'm uncertain if this type of PR is accepted, so I am perfectly fine with a close, but for whatever reason the API in this particular exercise seemed flawed. 

As a relative new comer to the Elixir language, I'm unsure if there is a better pattern for tuple checks than I used in the tests. It seems a little redundant to create a local from the return to pull the value out of the tuple but I don't know a better baked in solution to ExUnit. I'd be happy to update the tests to use a better matcher, or whatever would be most idiomatic for the language. 

Thanks for the hard work and the opportunity to submit a contribution.

### DIFF
--- a/exercises/hamming/example.exs
+++ b/exercises/hamming/example.exs
@@ -1,16 +1,17 @@
 defmodule DNA do
   @doc """
-  Compute the hamming distance between two character lists.
+  Returns number of differences between two strands of DNA, known as the Hamming Distance.
 
   ## Examples
 
-  iex> DNA.hamming_distance('ACCAGGG', 'ACTATGG')
-  2
+  iex> DNA.hamming_distance('AAGTCATA', 'TAGCGATC')
+  4
   """
-  def hamming_distance(strand1, strand2) when length(strand1) == length(strand2) do
-    pairs(strand1, strand2) |> count_mismatched
+  def hamming_distance(strand1, strand2) when length(strand1) === length(strand2) do
+    distance = pairs(strand1, strand2) |> count_mismatched
+    {:ok, distance}
   end
-  def hamming_distance(_, _), do: nil
+  def hamming_distance(_, _), do: {:error, "Lists must be the same length"}
 
   defp pairs(s1, s2),           do: Enum.zip(s1, s2)
   defp count_mismatched(pairs), do: Enum.count(pairs, fn({c1, c2}) -> c2 && (c1 != c2) end)
--- a/exercises/hamming/example.exs
+++ b/exercises/hamming/example.exs
@@ -5,7 +5,7 @@ defmodule DNA do
   ## Examples
 
   iex> DNA.hamming_distance('AAGTCATA', 'TAGCGATC')
-  4
+  {:ok, 4}
   """
   def hamming_distance(strand1, strand2) when length(strand1) === length(strand2) do
     distance = pairs(strand1, strand2) |> count_mismatched
--- a/exercises/hamming/hamming_test.exs
+++ b/exercises/hamming/hamming_test.exs
@@ -9,45 +9,38 @@ defmodule DNATest do
   use ExUnit.Case
 
 #   @tag :pending
-  test "hamming distance for strands of same length returns :ok" do
-    {code, _} = DNA.hamming_distance('', '')
-    assert code == :ok
+  test "hamming distance for strands of the same length are ok" do
+    assert {:ok, _} = DNA.hamming_distance('', '')
   end
 
   @tag :pending
   test "hamming distance is undefined for strands of different lengths" do
-    assert DNA.hamming_distance('AAAC', 'TAGGGGAGGCTAGCGGTAGGAC') == {:error, "Lists must be the same length."}
-    assert DNA.hamming_distance('GACTACGGACAGGACACC', 'GACATCGC') == {:error, "Lists must be the same length."}
+    assert {:error, "Lists must be the same length."} = DNA.hamming_distance('AAAC', 'TAGGGGAGGCTAGCGGTAGGAC')
+    assert {:error, "Lists must be the same length."} = DNA.hamming_distance('GACTACGGACAGGACACC', 'GACATCGC')
   end
 
   @tag :pending
   test "no difference between empty strands" do
-    {:ok, distance} = DNA.hamming_distance('', '')
-    assert distance == 0
+    assert {:ok, 0} = DNA.hamming_distance('', '')
   end
 
   @tag :pending
   test "no difference between identical strands" do
-    {:ok, distance} = DNA.hamming_distance('GGACTGA', 'GGACTGA')
-    assert distance == 0
+    assert {:ok, 0} = DNA.hamming_distance('GGACTGA', 'GGACTGA')
   end
 
   @tag :pending
   test "small hamming distance in middle somewhere" do
-    {:ok, distance} = DNA.hamming_distance('GGACG', 'GGTCG')
-    assert distance == 1
+    assert {:ok, 1} = DNA.hamming_distance('GGACG', 'GGTCG')
   end
 
   @tag :pending
   test "distance with same nucleotides in different locations" do
-    {:ok, distance} = DNA.hamming_distance('TAG', 'GAT')
-    assert distance == 2
+    assert {:ok, 2} = DNA.hamming_distance('TAG', 'GAT')
   end
 
   @tag :pending
   test "larger distance" do
-    {:ok, distance} = DNA.hamming_distance('ACCAGGG', 'ACTATGG')
-    assert distance == 2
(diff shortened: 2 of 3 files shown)


exercises/hamming/hamming_test.exs
```diff
@@ -8,34 +8,46 @@ ExUnit.configure exclude: :pending, trace: true
...
+  test "hamming distance is undefined for strands of different lengths" do
+    assert DNA.hamming_distance('AAAC', 'TAGGGGAGGCTAGCGGTAGGAC') == {:error, "Lists must be the same length."}
+    assert DNA.hamming_distance('GACTACGGACAGGACACC', 'GACATCGC') == {:error, "Lists must be the same length."}
+  end
+
+  @tag :pending
   test "no difference between empty strands" do
-    assert DNA.hamming_distance('', '') == 0
+    {:ok, distance} = DNA.hamming_distance('', '')
+    assert distance == 0
```
martinsvalin: You can also assert on a match. `assert {:ok, 0} = DNA.hamming_distance('','')`. The test failure message will be slightly different.
Or, maybe extract to a custom assertion:

    def assert_distance(one, two, distance) do
      assert {:ok, ^distance} = DNA.hamming_distance(one, two)
    end

    test "no difference between empty strands" do
      assert_distance('', '', 0)
    end

exercises/hamming/hamming_test.exs
```diff
@@ -8,34 +8,46 @@ ExUnit.configure exclude: :pending, trace: true
 defmodule DNATest do
   use ExUnit.Case
 
-  # @tag :pending
+#   @tag :pending
+  test "hamming distance for strands of same length returns :ok" do
+    {code, _} = DNA.hamming_distance('', '')
+    assert code == :ok
```
martinsvalin: This would probably be more idiomatic as `assert {:ok, _} = DNA.hamming_distance('', '')`.

I'd describe the test slightly different, as it doesn't just return `:ok`. I'd say "hamming distance for strands of the same length are ok"

exercises/hamming/hamming_test.exs
jedschneider: my first inclination was to pull out a helper assertion, I just wasn't sure that exercism community wanted the complication in an effort to keep the tests as "Plain Jane" as possible. I'll take a stab with the suggestions and see what you guys think! I'll probably keep the expected / actual on the same `lhs` `rhs` of the assertion, since that is what users are used to through the previous exercises.

jedschneider:
feedback changes finished. From my side, this is GTM unless a rebase/squash is preferred. The assert on pattern match suggested by @martinsvalin ended up really nice I think and gives good error messaging (see below). I'll use that pattern again in my own work. Good call.

```
  1) test hamming distance is undefined for strands of different lengths (DNATest)
     hamming_test.exs:17
     match (=) failed
     code: {:error, "Lists must be the same length."} = DNA.hamming_distance('AAAC', 'TAGGGGAGGCTAGCGGTAGGAC')
     rhs:  {:error, "Lists must be the same length"}
     stacktrace:
       hamming_test.exs:18
```

exercises/hamming/hamming_test.exs
```diff
@@ -8,34 +8,39 @@ ExUnit.configure exclude: :pending, trace: true
 defmodule DNATest do
   use ExUnit.Case
 
-  # @tag :pending
+#   @tag :pending
+  test "hamming distance for strands of the same length are ok" do
+    assert {:ok, _} = DNA.hamming_distance('', '')
+  end
```
dalexj: this test seems really redundant actually. we'll be checking for `{:ok, _}` in about every test already

jedschneider: I think the intent was to guide the solution without a major jump, but I'm
also indifferent about it.

On Mon, Jun 20, 2016 at 5:48 PM, Alex <notifications@github.com> wrote:

> In exercises/hamming/hamming_test.exs
> <https://github.com/exercism/xelixir/pull/200#discussion_r67792318>:
>
> > @@ -8,34 +8,39 @@ ExUnit.configure exclude: :pending, trace: true
> >  defmodule DNATest do
> >    use ExUnit.Case
> >
> > -  # @tag :pending
> > +#   @tag :pending
> > +  test "hamming distance for strands of the same length are ok" do
> > +    assert {:ok, _} = DNA.hamming_distance('', '')
> > +  end
>
> this test seems really redundant actually. we'll be checking for {:ok, _}
> in about every test already
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/exercism/xelixir/pull/200/files/51d04cb635c50aca5844604c9af14f475d86b501#r67792318>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAFBHCW8x1ni8ihNIKAfl-ECgE6yIst4ks5qNzTVgaJpZM4I5STM>
> .
>



-- 
*MODE SET**.*
_

jed schneider
developer.

706.207.4458
<http://bittheory.com/>modeset.com <http://www.modeset.com/>

exercises/hamming/example.exs
```diff
@@ -1,16 +1,17 @@
 defmodule DNA do
   @doc """
-  Compute the hamming distance between two character lists.
+  Returns number of differences between two strands of DNA, known as the Hamming Distance.
```
parkerl: Good catch!

exercises/hamming/hamming_test.exs
parkerl: I agree with @dalexj on this one.

exercises/hamming/hamming_test.exs
```diff
@@ -8,34 +8,39 @@ ExUnit.configure exclude: :pending, trace: true
...
+
+  @tag :pending
+  test "hamming distance is undefined for strands of different lengths" do
+    assert {:error, "Lists must be the same length."} = DNA.hamming_distance('AAAC', 'TAGGGGAGGCTAGCGGTAGGAC')
+    assert {:error, "Lists must be the same length."} = DNA.hamming_distance('GACTACGGACAGGACACC', 'GACATCGC')
+  end
+
+  @tag :pending
   test "no difference between empty strands" do
```
parkerl: Consider making this the first test.

exercises/hamming/hamming_test.exs
```diff
@@ -8,34 +8,39 @@ ExUnit.configure exclude: :pending, trace: true
...
   use ExUnit.Case
 
-  # @tag :pending
+#   @tag :pending
+  test "hamming distance for strands of the same length are ok" do
+    assert {:ok, _} = DNA.hamming_distance('', '')
+  end
+
+  @tag :pending
+  test "hamming distance is undefined for strands of different lengths" do
```
parkerl: Consider making this the last test. We don't really need the error case until the happy path is fully driven out.

exercises/hamming/hamming_test.exs
```diff
@@ -8,34 +8,39 @@ ExUnit.configure exclude: :pending, trace: true
...
+    assert {:error, "Lists must be the same length."} = DNA.hamming_distance('AAAC', 'TAGGGGAGGCTAGCGGTAGGAC')
+    assert {:error, "Lists must be the same length."} = DNA.hamming_distance('GACTACGGACAGGACACC', 'GACATCGC')
+  end
+
+  @tag :pending
   test "no difference between empty strands" do
-    assert DNA.hamming_distance('', '') == 0
+    assert {:ok, 0} = DNA.hamming_distance('', '')
```
parkerl: With the exception of that first test the rest of these can be `DNA.hamming_distance('', '') == {:ok, 0}` which is consistent with the other exercises.

jedschneider:
I feel like there are competing opinions here on how the tests should be structured. That's fine, and I'm glad there has been so much feedback, but maybe to reduce churn could you guys thumbsup the comments you want to see in the final PR? @dalexj @martinsvalin ? I don't really have an opinion, other than tightening up the return data type. Thanks!

jedschneider:
OK. I've updated those specs based on feedback. Thanks.

parkerl:
Great! Thanks so much! :tada:

PR #201, opened 2016-06-24
Add 'kindergarten-garden'

devonestes:
This adds the kindergarten garden exercise. There wasn't a set of example data
that we had in the x-common repo, so I chose to mimic the tests from the Ruby
implementation (i.e. adding the ability to pass in custom names, which wasn't
expressly stated in the Readme).

I'm hoping that the tests that I set up are generic enough to offer various
ways of solving this problem and that they're not too prescriptive.

### DIFF
--- a/config.json
+++ b/config.json
@@ -28,6 +28,7 @@
     "grade-school",
     "list-ops",
     "leap",
+    "kindergarten-garden",
     "etl",
     "meetup",
     "grains",
--- a/exercises/kindergarten-garden/example.exs
+++ b/exercises/kindergarten-garden/example.exs
@@ -1,0 +1,53 @@
+defmodule Garden do
+  @default_names [:alice, :bob, :charlie, :david, :eve, :fred, :ginny, :harriet,
+                  :ileana, :joseph, :kincaid, :larry]
+  @letter_map %{"V" => :violets, "R" => :radishes, "C" => :clover, "G" => :grass}
+
+  @doc """
+    Accepts a string representing the arrangement of cups on a windowsill and a
+    list with names of students in the class. The student names list does not
+    have to be in alphabetical order.
+
+    It decodes that string into the various gardens for each student and returns
+    that information in a map.
+  """
+
+  @spec info(String.t(), list) :: map
+  def info(info_string, student_names \\ @default_names) do
+    student_names = Enum.sort(student_names)
+    map = create_map(student_names)
+    [first_row, second_row] = String.split(info_string, "\n")
+    populate_map(first_row, second_row, map, student_names)
+  end
+
+  defp create_map(student_names) do
+    create_map(student_names, %{})
+  end
+
+  defp create_map([], map), do: map
+
+  defp create_map([h | t], map) do
+    map = Map.put(map, h, nil)
+    create_map(t, map)
+  end
+
+  defp populate_map(_, _, map, []), do: map
+
+  defp populate_map(first_row, second_row, map, [current_name | student_names]) do
+    first_row_letters = String.slice(first_row, 0..1)
+    second_row_letters = String.slice(second_row, 0..1)
+    letters = "#{first_row_letters}#{second_row_letters}"
+    map = add_to_map(map, current_name, letters)
+
+    first_row = String.slice(first_row, 2..-1)
+    second_row = String.slice(second_row, 2..-1)
+    populate_map(first_row, second_row, map, student_names)
+  end
+
+  defp add_to_map(map, current_name, letters) do
+    letters = String.codepoints(letters)
+    translated_list = Enum.map(letters, fn(letter) -> @letter_map[letter] end)
+    tuple = List.to_tuple(translated_list) 
+    Map.put(map, current_name, tuple)
+  end
+end
--- a/exercises/kindergarten-garden/garden.exs
+++ b/exercises/kindergarten-garden/garden.exs
@@ -1,0 +1,14 @@
+defmodule Garden do
+  @doc """
+    Accepts a string representing the arrangement of cups on a windowsill and a
+    list with names of students in the class. The student names list does not
+    have to be in alphabetical order.
+
+    It decodes that string into the various gardens for each student and returns
+    that information in a map.
+  """
+
+  @spec info(String.t(), list) :: map
+  def info(info_string, student_names) do
+  end
+end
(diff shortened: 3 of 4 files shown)


dalexj:
the return value of the full garden map already calculated for each student seems a little strange to me but I don't have any better ideas

are there any argument error cases we want to handle here?

devonestes:
Argument errors are an interesting idea. I went off of the Ruby implementation of this exercise, and I didn't see any [there](https://github.com/exercism/xruby/blob/master/exercises/kindergarten-garden/kindergarten_garden_test.rb), so I figured I'd try and stay consistent with the current implementations of this exercise.

When you say:

> the return value of the full garden map already calculated for each student seems a little strange to me but I don't have any better ideas

Do you mean that it might make sense to only calculate the garden for a given student when we get the data for each student, kind of in a lazy evaluation? I guess I don't totally know what you mean there.

dalexj:
you nailed exactly what I meant there.

devonestes:
well, I could refactor it to do that, but I don't think it would really need to be done. It would make the code _very_ complex. I can see a way (I think) of using closures and anonymous functions to do that for us, but I don't think the benefit would be worth the added complexity - at least not for now. If someone were to explore doing that in their solution that the submit to exercism.io, then that would be cool to see, but for this we're more just proving that the test suite works and that the problem can be solved as described.

dalexj:
I think we're talking about 2 different cases here.

I mean something more like this:

```elixir
# current
Garden.info("RC\nGG") # => %{name: {flowers} }
# my suggestion
Garden.info("RC\nGG", name) # => flowers
```

devonestes:
I get what you're saying now.

I don't think that would work since you won't know what flowers belong to which child in isolation. Because they're arranged alphabetically, you need the names of all the children to do any sort of decoding of the initial string.

```elixir
# can't do this
Garden.info("RC\nGC", name) # => we won't know which flowers go to that student without knowing the other students in the class

# We could do this, though
Garden.info("RG\nGC", all_names, name) # => flowers
```

I think the strategy I have now of "decoding" the data into a map works pretty well, but we could also implement `info\3` if you think the API above is worthwhile. Personally I don't see the benefit in it, but I'm happy to be convinced!

dalexj:
I'm not sure

something about this just makes sense to me in an OO language like ruby, but seems very strange in a functional language as elixir

I'm ok with both, but thought I'd bring it up

devonestes:
I totally see how it might look that way.

The way I chose to look at it, though, was essentially as a map/encryption operation. We have two datasets that we're using, one is kind of "encrypted" data (the string with the info), and the other is kind of the "key" to decrypt the data (the names of the students). We map over the encrypted data and use the key to decrypt it, giving us the decrypted data (the output map).

Maybe there's a better way to illustrate the problem in this way so it seems more idiomatic to Elixir/FP?

martinsvalin:
You could take inspiration from other FP language tracks that implement the exercise. [Haskell](https://github.com/exercism/xhaskell/blob/master/exercises/kindergarten-garden/src/Example.hs) has it, for example. It has separate functions for producing a garden (the map, in our case), and looking up a student in that garden. [Clojure](https://github.com/exercism/xclojure/blob/master/exercises/kindergarten-garden/src/example.clj) seems to simply produce a map, like this one.

I do think the naming is better in those two examples. `info` seems a bit generic. It's a map from students to plants. I'd suggest `students_plants` (as in the possessive "students' plants"), but maybe someone has a better idea. The name `info` bothers me a bit, though, and the test suite enforces it (except for people who've figured out they're allowed to change the tests).

devonestes:
How about `Garden.decode_string` or `Garden.decrypt_string`? Since that's kind of what we're doing, and it also makes the exercise sound a lot cooler than it seems at first, which is a _big_ plus in my book 😉

exercises/kindergarten-garden/kindergarten_garden_test.exs
```diff
@@ -0,0 +1,77 @@
+if !System.get_env("EXERCISM_TEST_EXAMPLES") do
+  Code.load_file("kindergarten_garden.exs")
```
parkerl: Usually the file name matches the module contained inside. This should be `garden.exs` or the module should be renamed.

parkerl:
@devonestes this looks good. One question. Why return tuples?

devonestes:
@parkerl Good question! I felt that it should either be a list or a tuple, and for some reason a tuple just made more sense to me. I'm not entirely sure why (maybe it's because I'm used to seeing atoms in tuples?), but I can't give a better reason than that. Is there some benefit to using something other than a tuple?

parkerl:
@devonestes as I expect, this now has a conflict with your other PR. Can you put it in its proper place in the new lineup? 

As for the tuples, it causes the submitter to take an extra step. They probably will end up with a list  in end, as you did and have to tupleize it at the last. 

IRL you would have to ask what operations this result will be fed into. If it is going to be iterated over down the line, a list is more appropriate. If it is going into pattern matches, then maybe a tuple is better. Food for thought. Ultimately it is a small issue. 

Can you address the filename/module name mismatch?

Thanks

devonestes:
@parkerl Thanks for the info about lists vs. tuples. I guess in my mind I would envision pattern matching, but that's probably just cause it's a collection of atoms, and I see that so frequently in pattern matching.

Anyway, the other stuff is all done now!

parkerl:
Thanks! 🎉

stosorio:
Hi!
Could someone look at the test "gets the garden for all students with custom child names"?
I have the impression that the order is a bit mixed up there. Or am i missing something?

```
    names = [:maggie, :nate, :xander, ***:ophelia***, :pete, :reggie, :sylvia,
             :tanner, :ursula, :victor, :winnie, :ynold]
    garden_string = "VRCGVV***RV***CGGCCGVRGCVCGCGV\nVRCCCG***CR***RGVCGCRVVCVGCGCV"
    [...]
    assert garden_info.ophelia == {:violets, :violets, :clover, :grass}
```
(*** to highlight ophelias part of the garden)

devonestes:
@stosorio The problem there is that the names aren't in alphabetical order, and as the problem description says, the garden is arranged in alphabetical order 😉  This is kind of a "trick question" to ensure that your implementation can handle names that aren't input in alphabetical order.

stosorio:
Damn.
Thanks, i'll read the readmes more carefully now :smile:

PR #203, opened 2016-07-01
Require SUT files relative to test file directory

waiting-for-dev:
This way you are not forced to run each test within test file directory, even
if you still can.

Useful to be able to run each test from the parent directory, which
makes workflow easier mainly if you are running everything inside, for
example, a docker container.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ The example solution should be named `example.exs`.
 
 ```elixir
 if !System.get_env("EXERCISM_TEST_EXAMPLES") do
-  Code.load_file("bob.exs")
+  Code.load_file("bob.exs", __DIR__)
 end
 
 ExUnit.start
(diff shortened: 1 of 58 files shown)


gmile:
Plus one on this feature. I see a good use case for it in future.

I keep all my exercises in a single folder, `~/exercism/elixir`. Given the feature this PR proposes, I will now be possible to easily **run tests for all exercises**.

This may be useful when tests for exercises get updated (unfortunately this process is not automated by exercism cli yet): one may want to check that new/updated tests still pass given the solution codes. Similarly, if a new version of elixir comes out with some methods removed/deprecated – one can run all tests and see if something fails.

This would allow a sort of a home grown CI for exercises.

parkerl:
Nice improvement. Thanks! 🎉

PR #207, opened 2016-07-30
Normalize file names for `nucleotide-count`

devonestes:
The file names for the test and the exercise didn't match the pattern that we
have for the rest of the exercises, so this commit changes that name and
updates the test so everything is consistent.


parkerl:
Thanks 🎉

PR #208, opened 2016-07-30
Add `it` to test case

jwworth:
Adds `it` to match other tests in the file.

Thank you for this project!


parkerl:
Awesome! Thanks 🎉

PR #210, opened 2016-08-01
Remove deprecated `HashSet` from allergies

### LINKED ISSUE #209: Remove deprecated HashSet from `allergies` tests
I was just looking through the tests for another thing I'm doing, and I saw that we're still using `HashSet` in `allergies` here: https://github.com/exercism/xelixir/blob/master/exercises/allergies/allergies_test.exs#L75

That's now deprecated, so we'll have to update it to use MapSet. I'll probably take care of that today, but I wanted to write it down so I don't forget (or so someone else can take care of it).

devonestes:
We still had a reference to `HashSet`, which is now deprecated, in the
`allergies` test. This replaces that with the new `MapSet` which replaced the
old module.

This fixes #209

### DIFF
--- a/exercises/allergies/allergies_test.exs
+++ b/exercises/allergies/allergies_test.exs
@@ -72,9 +72,9 @@ defmodule AllergiesTest do
   end
 
   defp assert_is_a_set_containing(list, to_contain) do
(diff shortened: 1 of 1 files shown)


parkerl:
Great! 🎉

Issue #190, opened 2016-06-07
Elixir track difficulty curve of early exercises.


Billiam:
I'm currently making my way through the elixir track slowly, but some of the exercises (so far, as a newcomer) seem to occur later or earlier than I would expect.

A couple of examples:

Exercise 5: The list-ops exercise, using core language features to re-implement reduce, append, filter, count, concat, and **map** occurs before:
Exercise 12: Accumulate (`Given a list and a function, apply the function to each list item and replace it with the function's return value.`)

Exercise 10: Nucleotide-count (Counting occurrences of a character in a character list, raising exceptions based on invalid arguments, creating histogram of preset characters), occurs before
Exercise 11: rna-transcription (simple letter substitution).

I didn't see any open (or closed) issues about this, so thought I'd throw this out there for discussion. Thanks!

kytrinyx:
I think we always end up with things in the wrong places for a number of reasons. First, it's never quite clear when we're making new exercises how difficult they actually will be compared to others. Second, we're always implementing exercises spread out across time, so we don't always remember everything about the existing exercises when we add new ones. Third, sometimes it just takes time and feedback to really understand which exercises are harder when you're new to the language (some exercises aren't conceptually difficult, but might introduce several new language features at once).

This is how it was tackled in the Rust track: https://github.com/exercism/xrust/issues/127

jedschneider:
I also noted the discrepancy [here](https://github.com/exercism/x-common/pull/274).

devonestes:
Ok, I've gone through and done a brief analysis of all the exercises in the track so far and made notes about them. Here's that initial analysis:

Exercise  | Concepts
----------|--------------
hello-world | basics
bob | control flow w/ cond, String processing (should be lower)
word-count | maps, lists, reduce
acronym | String processing, map
run-length-encoding | String processing, map_join
list-ops | compound, pattern matching, recursion, lists
sublist | pattern matching, recursion, lists
anagram | filter, String processing
bracket-push | control flow w/ cond, String processing
nucleotide-count | maps, String processing (should be _very_ early, very simple)
rna-transcription | map
accumulate | reduce, recursion (should be before list ops)
hamming | String processing, tuples
phone-number | String processing, Regex, pattern matching
beer-song | Ranges, default values, pattern matching, edge cases
grade-school | state, maps, lists, tuples
leap | algorithms, conditionals
etl | maps, algorithms
meetup | dates, algorithms
[...]

PR #211, opened 2016-08-01
Reorder exercises for Elixir track

devonestes:
There were some exercises that were rather easy that were somewhat far down the
track, and some that were pretty hard that were early on in the track. With
this we're attempting to find a better order of exercises for the track. I've
also tried to space out similar exercises so there aren't too many that have
common themes next to each other.

This addresses #190


parkerl:
Awesome! Thanks for tackling this! 🎉

Issue #159, opened 2016-03-26
Fail the build on trailing whitespace?


kytrinyx:
There's one exercise (OCR numbers) where trailing whitespace is necessary for the test data, but (a) I don't know if that exercise is implemented, and (b) maybe we could put the actual test data in external files with a different file extension.

If we make sure not to screw up any exercises, though, we could fail the build if there's trailing whitespace by sticking this in the travis file.

```
- "! git grep ' $' -- \\*.exs"
```

dalexj:
OCR is not on the elixir track right now

kytrinyx:
Cool, so technically we could add this and then figure out how to deal with the whitespace issue in the OCR test data later (as my partner says: "This is a problem for future Katrina").

parkerl:
@dalexj PR?

PR #213, opened 2016-08-04
Update Elixir and Erlang versions, add check for trailing whitespace

devonestes:
Issue #159 was about failing the build based on the presence of trailing
whitespace. This adds that, and while I was in the `.travis.yml` file I updated
our versions for Elixir and Erlang so we're up to the latest.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,14 @@
 ---
 language: elixir
 elixir:
-  - 1.2.0-rc.0
+  - 1.3.2
 otp_release:
-  - 18.1
+  - 19.0
 
 script:
 - bin/fetch-configlet
 - bin/configlet .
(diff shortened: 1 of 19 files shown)


parkerl:
Great! 🎉

DoggettCK:
Cool, good to know. Rubocop at work might have me a bit spoiled. Hope Elixir has something similar.

devonestes:
@DoggettCK It does - two at the moment actually! There's Credo (which I prefer): https://github.com/rrrene/credo and there's Dogma: https://github.com/lpil/dogma

I thought about adding those, but for this single check I didn't want to add a whole other dependency to the project. For now this is enough. We'll see if we need to enforce some further guidelines in the future, though!

Issue #206, opened 2016-07-27
Add test case for Prime Factors


sunboshan:
Consider add one more test case for [Prime Factors](https://github.com/exercism/xelixir/blob/master/exercises/prime-factors/prime_factors_test.exs)
```
test "10000000055" do
  assert PrimeFactors.factors_for(10000000055) == [5, 2000000011]
end
```

With my [first iteration](http://exercism.io/submissions/164bc866d4dd4ed989f2cf321b4658c0), it took 44189.1ms to pass this test. So need to tweak the solution to get this test pass in a decent time. I think it would serve a good test case to make your solution more optimized.

kytrinyx:
I don't necessarily want to force people to consider optimization as a requirement, but thinking about trade-offs makes the exercise more interesting.

How about making it an optional test (as in "here's something to think about")? Do we have a mechanism for that @exercism/elixir?

martinsvalin:
I think it'd be fine to simply say so in a code comment by the `@tag :pending`. Or you could be fancy and configure ExUnit to `exclude: [:pending, :extra_credit]`, and mark the test `@tag :extra_credit` instead of `:pending`.

PR #214, opened 2016-08-04
Add additional test case for `prime-factors`

devonestes:
This commit adds an additional test case for the `prime-factors` exercise. The
idea was to add in optional performance tests to see if your solution was
quick. I don't quite know what we're going to call "quick" for the purposes of
this test, but since the person who posted the issue said their initial
implementation took 44 seconds, I chose 20 seconds as a starting point.

This addresses Issue #206.

### DIFF
--- a/exercises/prime-factors/prime_factors_test.exs
+++ b/exercises/prime-factors/prime_factors_test.exs
@@ -62,4 +62,14 @@ defmodule PrimeFactorsTest do
   test "93819012551" do
     assert PrimeFactors.factors_for(93819012551) == [11, 9539, 894119]
   end
+
+  @tag :pending
+  # @tag timeout: 2000
+  #
+  # The timeout tag above will set the below test to fail unless it completes
+  # in under two sconds. Uncomment it if you want to test the efficiency of your
+  # solution.
+  test "10000000055" do
+    assert PrimeFactors.factors_for(10000000055) == [5, 2000000011]
+  end
 end
--- a/exercises/prime-factors/example.exs
+++ b/exercises/prime-factors/example.exs
@@ -1,18 +1,16 @@
 defmodule PrimeFactors do
-  def factors_for(n) when n < 2, do: []
+  @spec factors_for(pos_integer) :: [pos_integer]
   def factors_for(number) do
-    case least_divisor(number) do
-      nil    -> [number]
-      factor -> [factor, factors_for(round(number/factor))]
-    end |> List.flatten
+    do_factors(number)
   end
 
-  defp least_divisor(number) do
-    number
-      |> factor_range
-      |> Stream.filter(&(rem(number, &1) == 0))
-      |> Enum.at(0)
-  end
-
-  defp factor_range(number), do: (2..round(number/2))
+  defp do_factors(_, i \\ 2, acc \\ [])
+  defp do_factors(1, _, acc),
+    do: Enum.reverse(acc)
+  defp do_factors(n, i, acc) when n < i * i,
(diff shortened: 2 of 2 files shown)


exercises/prime-factors/prime_factors_test.exs
```diff
@@ -62,4 +62,16 @@ defmodule PrimeFactorsTest do
   test "93819012551" do
     assert PrimeFactors.factors_for(93819012551) == [11, 9539, 894119]
   end
+
+  @tag :optional
+  @tag timeout: 120000
+  test "performance optimizations" do
+    %{minute: original_minutes, second: original_seconds} = DateTime.utc_now
+    PrimeFactors.factors_for(10000000055) == [5, 2000000011]
```
dalexj: does this need `assert`?

devonestes: Good point - I'm not sure we care about the result here, but it couldn't hurt to add an assertion there. Or maybe split it into two tests - one to get the value, and the second to ensure that it finishes in a reasonable amount of time?

devonestes: @dalexj I ended up splitting it up into two tests - that made the most sense to me. I also changed the assertion as to how fast that function returns, since I saw on the second iteration from @sunboshan that he got it to finish in 2.4ms, so I figured under 2 seconds would be a safe max to make that test pass.

sunboshan: @devonestes Thanks for the PR! My original thought was just to add this test case. The non optimized solution would still pass the test, but very slow. This will make the users to try to optimized solution. But again, it's optional, so personally I don't feel the need to add the time constrain. Because for different computer cpus, the time may vary.

dalexj: I don't think this needs to be 2 test cases, code written up to that point should have been tested enough that it will get the correct answer. Even then, the order of the assertions will give you what you want anyway. These are optional tests so there's no value I see from splitting them

parkerl:
I'm not a huge fan of adding an `optional` tag. 

I suggest we keep this simple. The default timeout for ExUnit specs is 1 minute (60_000 ms). We could add above the existing test a commented out `timeout` tag with a less forgiving amount of time. The comment should include an explanation of how and why it can be used.

devonestes:
@parkerl I'm ok with that in this case. I guess I can see a case for an `optional` tag later on if we want to add additional types of optional tests, but there's no need to go down that road now.

exercises/prime-factors/prime_factors_test.exs
```diff
@@ -3,7 +3,7 @@ if !System.get_env("EXERCISM_TEST_EXAMPLES") do
 end
 
 ExUnit.start
-ExUnit.configure exclude: :pending, trace: true
+ExUnit.configure exclude: [:pending], trace: true
```
parkerl: Revert?

exercises/prime-factors/prime_factors_test.exs
```diff
@@ -62,4 +62,14 @@ defmodule PrimeFactorsTest do
   test "93819012551" do
     assert PrimeFactors.factors_for(93819012551) == [11, 9539, 894119]
   end
+
+  @tag :pending
+  # @tag timeout: 2000
+  #
+  # The above tag will set the below test to fail unless it completes in under
```
parkerl: Sorry to nitpick.

Perhaps `The 'timeout' tag above will set the test below to`?

parkerl:
Awesome!!! 🎉

Issue #204, opened 2016-07-06
typespecs and dialyzer


NobbZ:
I really like the typespecs which are provided with the module skeletons.

But sometimes they don't get updated in the example and/or the skelleton when tests are updated to expect new types, also I do often forget to update them when I catch for the newest version of the tests.

Is there an easy and hassle free way to actually do a check against the specs without wrapping everything into a mix-project?

And if so, could this be used in CI also?

Can we mention it in the README as a part of running the tests?

kytrinyx:
Yes, we should definitely get this sort of check in place. I don't know what the hassle-free approach would be, though!

lpil:
Would be interested to see how you do this. I've not found any examples of anyone running dialyzer on CI, probably due to the faff of generating and managing the PLT.

devonestes:
So after some research there definitely isn't an "easy and hassle free" way of getting dialyzer set up without at least some more structure to the project (and by structure, I mean like Mix structure). Theoretically we could do something to make it work, but I question if the cost would be worth the benefit. I'm going to explore other ways that we can get the same benefit that wouldn't require such a gnarly change to the current codebase.

@lipil - yeah, the PLT takes forever and is a bummer, but luckily there's open source to the rescue! Check out this repo for pre-built PLTs for pretty much every combo of Erlang and Elixir versions that you can download (and info on how to make it all work): https://github.com/danielberkompas/travis_elixir_plts

lpil:
This feels like yet another reason to adopt the standard Elixir project structure.

devonestes:
@lpil - yeah, that definitely crossed my mind as I was working on this yesterday. I'm just not sure how we'll be able to do that within the structure that we need to keep to stay compliant with the x-exercises API. I'm going to do some research on what _exactly_ we need to implement to make sure the API works.

Part of me actually thinks we might be able to structure this as an umbrella app, but that's for further investigation. Of course, if you have any input, I'd be happy to hear it!

lpil:
What is the part of the x-exercises API we would be unable to comply to?

parkerl:
Closed by #258 and #237

PR #215, opened 2016-08-05
[WIP] Add Dialyzer to travis.yml file

devonestes:
Apparently some of our typespecs have been getting out of date, so with this
commit we're adding Dialyzer to our CI process to fail the build if that
happens.

This addresses Issue #204

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,17 @@
 ---
 language: elixir
 elixir:
-  - 1.2.0-rc.0
+  - 1.3.0
 otp_release:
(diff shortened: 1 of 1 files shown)


NobbZ:
Is this a WIP?

Right now dialyzer simply fails because there are no compiled modules...

devonestes:
Yeah, WIP for right now. I'll change the title to make that clear. It's my first time integrating dialyzer into CI, so it might take me a little bit to figure out!

devonestes:
So

PR #228, opened 2016-08-22
Change build to fail on compile warnings

devonestes:
We had accumulated several compiler warnings throughout this project, and in
this PR I'm adding some configuration that will fail the CI build if there are
any compiler warnings by turning them into errors thanks to the
`--warnings-as-errors` flag.


dalexj:
sounds good

parkerl:
Thanks. 🎉

PR #230, opened 2016-08-23
Update tests and example for NucleotideCount

### LINKED ISSUE #224: nucleotide: raising errors?
It's not very Elixir-ish to raise an error. The preferred pattern is to return a tuple like `{:error, :invalid_strand}`. Is this only there as a learning tool to learn how to raise errors?

[Edit: adding link to  exercise https://github.com/exercism/xelixir/tree/master/exercises/nucleotide-count ]

devonestes:
There was a good point raised about some conventions in Elixir around raising
errors vs. returning tuples with error messages, so we've made some revisions
to the tests and example implementation for this exercise. We decided that
since there are examples of raising exceptions in the Elixir standard library,
that it wouldn't be a bad thing to include that functionality, but that we'd
_also_ include the more common pattern of returing tuples with error messages.

Closes #224

### DIFF
--- a/exercises/nucleotide-count/nucleotide_count_test.exs
+++ b/exercises/nucleotide-count/nucleotide_count_test.exs
@@ -41,25 +41,4 @@ defmodule NucleotideCountTest do
     expected = %{?A => 20, ?T => 21, ?C => 12, ?G => 17}
     assert NucleotideCount.histogram(s) == expected
   end
-
-  @tag :pending
-  test "histogram validates the strand" do
-    assert_raise ArgumentError, fn ->
-      NucleotideCount.histogram('JOHNNYAPPLESEED')
-    end
-  end
-
-  @tag :pending
-  test "count validates the nucleotide" do
-    assert_raise ArgumentError, fn ->
-      NucleotideCount.count('', ?U)
-    end
-  end
-
-  @tag :pending
-  test "count validates the strand" do
-    assert_raise ArgumentError, fn ->
-      NucleotideCount.count('JOHNNYAPPLESEED', ?A)
-    end
-  end
 end


parkerl:
@devonestes I'm really on the fence about this change. I still find it difficult to decide which style of error would be better or more "idiomatic" in this case. Is it worth making such a large change to this exercise and by doing so negating all existing submissions? Perhaps the least contentious thing to do is simply remove the `assert_raise ArgumentError` test. This would leave existing submissions mostly valid with specs green. 

In any case, I really don't like that we have complete duplication of all the tests and suggest that one side or the other simply have a single happy path test verifying it works.

devonestes:
@parkerl I've been thinking about this, and I think you're mostly right - this is taking it a little too far. However, I think the issue that was raised is still pretty valid. While there are examples of raising exceptions in the StdLib, they're almost always `!` functions, and right now we're raising errors from non `!` functions, which isn't really good, ideomatic Elixir.

What do you think - should we change the API of the module to remove `count/2` and have `count!/2` only, or should we change what `count/2` returns so it's a tuple instead? I don't think there's a good way we can fix what we have here in a non-breaking fashion, though... It feels kind of wrong to me to just not include the error handling tests, but maybe I'm not giving appropriate weight to the breakage that would occur. Can you explain a bit more about why that would be a serious issue?

As I mentioned before, there are ways that other language tracks manage the "versioning" of tests - see here for the Golang track, which uses a constant to keep track of the test versions so that users know when their implementation is no longer up to date with the current tests: https://github.com/exercism/xgo/blob/master/exercises/bracket-push/bracket_push_test.go#L7

Do you think something like this could be helpful?

parkerl:
Thanks @devonestes 

With nucleotide count sitting at number two in the exercise lineup, we feel that it might be too soon  to introduce either error paradigm to new users of Elixir. There is also the issue of invalidating many existing submissions. Therefore we are going to make the least invasive change by removing the errors cases altogether.

PR #229, opened 2016-08-23
Load correct files

veelenga:
```sh
❯ elixir rna-transcription/rna_transcription_test.exs
** (Code.LoadError) could not load [...]/elixir/rna-transcription/dna.exs
    (elixir) lib/code.ex:657: Code.find_file/2
    (elixir) lib/code.ex:319: Code.load_file/2
    rna-transcription/rna_transcription_test.exs:2: (file)
    (elixir) lib/code.ex:363: Code.require_file/2

❯ elixir hamming/hamming_test.exs 

HammingTest
  * test no difference between empty strands (6.5ms)

  1) test no difference between empty strands (HammingTest)
     hamming/hamming_test.exs:11
     ** (UndefinedFunctionError) function Hamming.hamming_distance/2 is undefined (module Hamming is not available)
     stacktrace:
       Hamming.hamming_distance([], [])
       hamming/hamming_test.exs:12: (test)

Finished in 0.03 seconds (0.03s on load, 0.00s on tests)
1 test, 1 failure

```

PR #232, opened 2016-08-25
Fix typo in module name for hamming distance exercise

wiserfirst:
There is a typo in the module name for the hamming distance exercise and this PR fixes it


veelenga:
@wiserfirst i'have also did that fix in #229 :)

wiserfirst:
Oh. I'll close mine. Thanks!

PR #234, opened 2016-08-30
Fix path to rna_transcription.exs

nikitachernov:
It seems that test file attempts to load a file by wrong name.


parkerl:
Thanks for the PR! I think this was fixed in #229. My bad for not merging it sooner. Sorry!

PR #237, opened 2016-09-07
Add dialyzer to ci

parkerl:
@devonestes I think you were very close on this. I have built upon your fine work.

The biggest problem is that out of the box mix is not going to compile `.exs` files to `.beam` files that can be analyzed by dialyzer. We can force that with `elixirc`. 

The next issue is that all the files we want to analyze are called `example.exs` which does not give the best feedback as you can see in this build https://travis-ci.org/parkerl/xelixir/builds/157473389 I've done some shell scripting here to help with that. 

There are at least two issues left.

First, I have hard coded this to the 1.3.0 PLT version which seems to work on Travis but not for me locally. 

Second, this is a bit rough to run locally when the PLT is not hard coded as you have to set the Travis specific environment variables. We should be able to let contributors run this as smoothly as possible on their PRs.

P.S. There was a great Dialyzer talk at ElixirConf. I recommend a watch when the vids come out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,5 +16,7 @@ script:
 - EXERCISM_TEST_EXAMPLES=true elixirc --warnings-as-errors **/example.exs **/*_test.exs -o ./tmp > /dev/null
 - rm -rf ./tmp
 
+- dialyzer --no_check_plt --plt $PLT_LOCATION
+
 - "! git grep ' $' -- \\*.exs | grep -v -e 'exercises/diamond/diamond_test.exs'"
 sudo: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ script:
 - EXERCISM_TEST_EXAMPLES=true elixirc --warnings-as-errors **/example.exs **/*_test.exs -o ./tmp > /dev/null
 - rm -rf ./tmp
 
-- dialyzer --no_check_plt --plt $PLT_LOCATION
+- bin/dialyzer_check.sh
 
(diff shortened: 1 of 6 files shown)


devonestes:
Sorry it's taken me so long to get to this! Been a busy couple of weeks...

This is a great step forward! I think we can take care of those travis-specific environment variables by using conditional assignment in the bash script. Something like this, maybe?

```bash
: ${TRAVIS_ELIXIR_VERSION=1.3}
```

I wouldn't say I'm a bash scripting expert so that might not be exactly proper syntax, but the idea is saying "if this variable isn't set, set it to this, otherwise take what's set". That way we can have values in there for a user to use on their local machine just fine, but still take whatever's set by Travis without overwriting it.

The unhelpful errors thing is rough, though. Honestly, I don't know enough about Dialyzer to know if we can change the output at all to get a more complete file path. The other alternative that I think we could do is to only use Dialyzer on one file at a time (like, in the loop rather than outside of it) so we can print something to STDOUT telling the user which file we're checking - that way they'll know which `example.exs` is in need of some love. I'm not entirely sure that's a great idea, but just wanted to throw it out there in case it helps folks come up with something good to solve that issue.

I'll keep thinking about this, though, and play around a bit myself as well!

parkerl:
@devonestes thanks. Actually I have figured out the file name issue. I will push up my latest WIP so you can take a look.

parkerl:
@devonestes w00t! This is finally 🍏 ! 

What do you think? Good to go?

devonestes:
Awesome! This is ready to rock 'n roll! 👍

PR #242, opened 2016-09-10
Add pangram exercise

tejasbubane:


config.json
```diff
@@ -64,7 +64,8 @@
     "bowling",
     "dot-dsl",
     "custom-set",
-    "forth"
+    "forth",
+    "pangram"
```
tejasbubane: Putting this at the end for now since reordering will be done after adding `topics` and `difficulties` for all exercises.

config.json
```diff
@@ -432,6 +433,12 @@
       "difficulty": 1,
       "topics": [
       ]
+    },
+    {
+      "slug": "pangram",
+      "difficulty": 1,
+      "topics": [
+      ]
```
tejasbubane: Should we add topics now itself or keep it for later when all others are added?

dalexj:
I've gotten into the thinking of using `assert` for true/false gives better error messages and reads more easily when using `assert something == true` or `assert something == false` rather than assert/refute.

Just my thoughts, not end all be all.

exercises/pangram/pangram_test.exs
```diff
@@ -0,0 +1,60 @@
...
+    assert Pangram.pangram?("the_quick_brown_fox_jumps_over_the_lazy_dog")
+  end
+
+  @tag :pending
+  test "pangram with numbers" do
+    assert Pangram.pangram?("the 1 quick brown fox jumps over the 2 lazy dogs")
+  end
+
+  @tag :pending
+  test "missing letters replaces by numbers" do
```
parkerl: replaces should be `replaced`

exercises/pangram/pangram_test.exs
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  @tag :pending
+  test "pangram with mixed case and punctuation" do
+    assert Pangram.pangram?("Five quacking Zephyrs jolt my wax bed.")
+  end
+
+  @tag :pending
+  test "pangram with non ascii characters" do
+    assert Pangram.pangram?("Five quacking Zephyrs jolt my wax bed.")
```
parkerl: Looks like the input for this test should be 

`Victor jagt zwölf Boxkämpfer quer über den großen Sylter Deich.`

parkerl:
@tejasbubane thanks for contributing this. There are just a few small issues to resolve. Thanks.

config.json
devonestes: I don't think it would hurt to add topics now - worst thing that could happen is they change later on, but that wouldn't be all that bad!

tejasbubane:
@parkerl Fixed as per your comments and added a couple of topics to config.

exercises/pangram/pangram.exs
```diff
@@ -0,0 +1,18 @@
+defmodule Pangram do
+  @doc """
+  Determines if a word or sentence is an pangram.
```
devonestes: Small typo here - "Determines if a word or sentence is __a__ pangram"

tejasbubane: Thanks @devonestes 👍  Fixed the typo.

parkerl:
👍 looks great! 🎉 @devonestes if you are good care to do the honors?

PR #243, opened 2016-09-13
Add clock exercise

tejasbubane:
Need your thoughts on the APIs. Most object oriented languages tend to store `minutes` in object state. I used plain data transformations.

### DIFF
--- a/config.json
+++ b/config.json
@@ -65,7 +65,8 @@
     "dot-dsl",
     "custom-set",
     "forth",
-    "pangram"
+    "pangram",
+    "clock"
   ],
   "exercises": [
     {
@@ -440,6 +441,17 @@
       "topics": [
         "Strings",
         "Enumerations"
+      ]
+    },
+    {
+      "slug": "clock",
+      "difficulty": 1,
+      "topics": [
+        "Protocols",
+        "Structs",
+        "Logic",
+        "Transforming",
+        "Integers"
       ]
     }
   ],
--- a/exercises/clock/clock.exs
+++ b/exercises/clock/clock.exs
@@ -1,0 +1,28 @@
+defmodule Clock do
+  defstruct hour: 0, minute: 0
+
+  @doc """
+  Returns a string representation of a clock:
+
+      iex> Clock.new(8, 9) |> to_string
+      "08:09"
+  """
+  @spec new(integer, integer) :: Clock
+  def new(hour, minute) do
+  end
+
+  @doc """
+  Adds two clock times:
+
+      iex> Clock.add(10, 0) |> Clock.add(3) |> to_string
+      "10:03"
+  """
+  @spec add(Clock, integer) :: Clock
+  def add(%Clock{hour: hour, minute: minute}, add_minute) do
+  end
+
+  defimpl String.Chars, for: Clock do
+    def to_string(%Clock{hour: hour, minute: minute}) do
+    end
+  end
+end
--- a/exercises/clock/clock_test.exs
+++ b/exercises/clock/clock_test.exs
@@ -1,0 +1,266 @@
+if !System.get_env("EXERCISM_TEST_EXAMPLES") do
+  Code.load_file("clock.exs", __DIR__)
+end
+
+ExUnit.start
+ExUnit.configure exclude: :pending, trace: true
+
+defmodule ClockTest do
+  use ExUnit.Case
+
+  describe "create" do
+    # @tag :pending
+    test "on the hour" do
+      assert Clock.new(8, 0) |> to_string == "08:00"
+    end
+
+    @tag :pending
+    test "past the hour" do
+      assert Clock.new(11, 9) |> to_string == "11:09"
+    end
+
+    @tag :pending
+    test "midnight is zero hours" do
+      assert Clock.new(24, 0) |> to_string == "00:00"
+    end
+
+    @tag :pending
+    test "hour rolls over" do
+      assert Clock.new(25, 0) |> to_string == "01:00"
+    end
+
+    @tag :pending
+    test "hour rolls over continuously" do
+      assert Clock.new(100, 0) |> to_string == "04:00"
+    end
+
+    @tag :pending
+    test "sixty minutes is next hour" do
+      assert Clock.new(1, 60) |> to_string == "02:00"
+    end
+
+    @tag :pending
+    test "minutes roll over" do
+      assert Clock.new(0, 160) |> to_string == "02:40"
+    end
+
+    @tag :pending
+    test "minutes roll over continuously" do
+      assert Clock.new(0, 1723) |> to_string == "04:43"
+    end
+
+    @tag :pending
+    test "hour and minutes roll over" do
+      assert Clock.new(25, 160) |> to_string == "03:40"
+    end
+
+    @tag :pending
+    test "hour and minutes roll over continuously" do
+      assert Clock.new(201, 3001) |> to_string == "11:01"
+    end
+
+    @tag :pending
+    test "hour and minutes roll over to exactly midnight" do
+      assert Clock.new(72, 8640) |> to_string == "00:00"
+    end
+
+    @tag :pending
+    test "negative hour" do
+      assert Clock.new(-1, 15) |> to_string == "23:15"
+    end
+
+    @tag :pending
+    test "negative hour rolls over" do
+      assert Clock.new(-25, 0) |> to_string == "23:00"
+    end
+
+    @tag :pending
+    test "negative hour rolls over continuously" do
+      assert Clock.new(-91, 0) |> to_string == "05:00"
+    end
+
+    @tag :pending
+    test "negative minutes" do
+      assert Clock.new(1, -40) |> to_string == "00:20"
+    end
+
+    @tag :pending
+    test "negative minutes roll over" do
+      assert Clock.new(1, -160) |> to_string == "22:20"
+    end
+
+    @tag :pending
+    test "negative minutes roll over continuously" do
+      assert Clock.new(1, -4820) |> to_string == "16:40"
+    end
+
+    @tag :pending
+    test "negative hour and minutes roll over" do
+      assert Clock.new(-25, -160) |> to_string == "20:20"
+    end
+
+    @tag :pending
+    test "negative hour and minutes roll over continuously" do
+      assert Clock.new(-121, -5810) |> to_string == "22:10"
+    end
+  end
+
+  describe "add" do
+    @tag :pending
+    test "add minutes" do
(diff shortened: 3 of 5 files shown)


dalexj:
take a look at maybe the bowling exercise to get a feel of the API of the module: https://github.com/exercism/xelixir/blob/master/exercises/bowling/bowling_test.exs

I'd suggest having something like this:

```elixir
clock = Clock.new(hour, minutes)
clock = Clock.add(clock, more_minutes)
"3:15" == Clock.to_s(clock) # possible way of checking this
```

martinsvalin:
I agree, a `Clock` struct, with a `new` function for creating clocks would be better.

I'd also suggest implementing the `String.Chars` protocol, instead of `Clock.to_s`. Afaik no other exercise implements it, and I think this could be a good exercise to expose people to protocols.

```
Clock.new(hour, minute) |> Clock.add(more_minutes) |> to_string
```

Reads quite nicely, imho.

parkerl:
I would prefer `start` to `new` but I agree with that design. I find that `Something.new` tends to lead those new to Elixir to think in terms of objects. (Raises eyebrow at several stdlib functions 😁) 

@martinsvalin can you help explain how we might gracefully lead the user into implementing that protocol? Just call it out explicitly in the first test?

@tejasbubane thanks for the great start!

martinsvalin:
@parkerl - I agree, `Something.new` leads the OOP programmer to think of objects, but it's pretty common in Elixir's standard library as well for structs that should be constructed with a helper function, instead of a raw `%Something(foo: "foo")` etc. The new Calendar structs use `new`, for example.

Naming the function `start` could lead someone to believe it starts a process.

As to leading someone to implement a protocol…

If you try to do string conversion on a struct that doesn't implement `String.Chars`, you'll get a relevant error, so that helps:

```
iex> to_string Clock.new(0,0)
** (Protocol.UndefinedError) protocol String.Chars not implemented for %Clock{hour: 0, minute: 0}
```

So a plain `assert to_string(Clock.new(0,0)) == "00:00"` would at least give a reasonably actionable failure message.

If we wanted to have a more helpful, guiding test failure message, since this might be the first time someone even heard of Elixir protocols, you could control the error message like this:

```
test "to_string" do
  try do
    to_string(Clock.new(0,0))
  rescue
    Protocol.UndefinedError ->
        refute(true, "Can't convert Clock to string.\n" <>
          "Hint: implement the String.Chars protocol for Clock.\n" <>
          "http://elixir-lang.org/getting-started/protocols.html\n" <>
          "http://elixir-lang.org/docs/stable/elixir/String.Chars.html")
  end
end
```

tejasbubane:
> guiding test failure message, since this might be the first time someone even heard of Elixir protocols, you could control the error message

This would be of great help for beginners 👍  
Should we also give a stub for protocol like just like we give it for all public API methods?

```elixir
defimpl String.Chars, for: Clock do
  def to_string(%Clock{hr: hr, min: min}) do
  end
end
```

martinsvalin:
Oh right. We usually have stubs for functions we expect them to implement. Yes, we should have that, and I think maybe the stub and the default error message `… protocol String.Chars not implemented for …` is enough.

But we could go the extra mile with the `try … rescue` test.

config.json
```diff
@@ -441,6 +442,16 @@
...
       ]
+    },
+    {
+      "slug": "clock",
+      "difficulty": 1,
+      "topics": [
+        "Logic",
+        "Transforming",
+        "Tuples",
+        "Integers"
```
parkerl: Add protocols

exercises/clock/clock.exs
```diff
@@ -0,0 +1,28 @@
+defmodule Clock do
+  defstruct hr: 0, min: 0
+
+  @doc """
+  Returns a string representation of a clock:
+
+      iex> Clock.new(8, 9) |> to_s
```
parkerl: `to_s` should be `to_string` I believe.

exercises/clock/clock.exs
```diff
@@ -0,0 +1,28 @@
...
+      "08:09"
+  """
+  @spec new(integer, integer) :: Clock
+  def new(hr, min) do
+  end
+
+  @doc """
+  Adds two clock times:
+
+      iex> Clock.add(10, 0, 3) |> to_s
```
parkerl: This example needs updating and `to_s` should be `to_string` I believe.

exercises/clock/clock.exs
```diff
@@ -0,0 +1,28 @@
+defmodule Clock do
+  defstruct hr: 0, min: 0
```
parkerl: Would you mind fully spelling out `hour` and `minute` as in done in the new `Calendar` module?

tejasbubane:
Thanks for your suggestions everyone! I have made the changes accordingly. New APis seem to be much better than the previous implementation. Loved the fact that using structs completely eliminated the need to test equality. Just using plain old `==` works now! 😄 

@martinsvalin Using just `defimpl` stub for now and I think this and the default error should be sufficient. 

PS: I had to disable `protocol_consolidation` for `mix test` to pass.

exercises/clock/clock.exs
```diff
@@ -0,0 +1,28 @@
...
+  Adds two clock times:
+
+      iex> Clock.add(10, 0, 3) |> to_s
+      "10:03"
+  """
+  @spec add(Clock, integer) :: Clock
+  def add(%Clock{hr: hr, min: min}, add_min) do
+  end
+
+  defimpl String.Chars, for: Clock do
```
parkerl: Ok, by stubbing out `String.Chars` we have done all the hard work for the user. However, does this give the introduction to protocols we are looking for? We no longer get the protocol not implemented error.

exercises/clock/clock_test.exs
```diff
@@ -0,0 +1,266 @@
...
+      assert Clock.new(5, 32) |> Clock.add(-1500) |> to_string == "04:32"
+    end
+
+    @tag :pending
+    test "subtract more than two days" do
+      assert Clock.new(2, 20) |> Clock.add(-3000) |> to_string == "00:20"
+    end
+  end
+
+  describe "equal?" do
```
parkerl: since you are not implementing an `equal?` function this describe should be renamed. `describe '=='`

parkerl:
@tejasbubane I think this is looking great! 

@martinsvalin `new` seems good in this instance where the user is required to use a struct due to the protocol implementation.

exercises/clock/clock.exs
tejasbubane: Ahh, the ruby muscle memory 😄

exercises/clock/clock.exs
tejasbubane: Hmm right. This does not give that error we were looking for. I guess I should remove the stub and go the rescue way giving explicit message to the user.

parkerl:
Lets see what others feel first?

exercises/clock/clock.exs
```diff
@@ -0,0 +1,28 @@
...
+
+      iex> Clock.add(10, 0) |> Clock.add(3) |> to_string
+      "10:03"
+  """
+  @spec add(Clock, integer) :: Clock
+  def add(%Clock{hour: hour, minute: minute}, add_minute) do
+  end
+
+  defimpl String.Chars, for: Clock do
+    def to_string(%Clock{hour: hour, minute: minute}) do
```
tejasbubane: I am in double minds regarding the usage of `struct` everywhere in the stub file. 
I think we should not influence the thinking of users towards usage of `struct` since this is an implementation detail. The problem can be implemented using some other data structure (eg. tuples) and maybe someone might come up with a much more elegant solution without the use of struct.

parkerl: I like your thinking and normally would totally agree. However the protocol forces the use of a struct so we might as well document it as such.

tejasbubane: @parkerl `Protocol` does not force to use a `struct`. One can always implement the protocol on some other built-in data type (which I don't think is a good practice in this case but can be done). 
Here is an example of using `tuple` instead of `struct` keeping the public API exactly the same: https://gist.github.com/tejasbubane/9e7ae3253756edf093dbc98b4e8e65ff
All the test-cases pass without any change.

parkerl:
Hey all, can we get some feedback regarding the stubbing out of the protocol? Should it already be there for the user to simply fill in or is it a better experience to have to stumble through implementing it completely?

martinsvalin:
Good catch, that the mere presence of the stub removes the failure message that drives the implementation of the protocol. I'm leaning towards removing the stub, to drive home the test-driven aspect and to have people learn a little about protocols.

It makes the exercise a bit harder, especially since it breaks the norm of having stubs. Perhaps that is an argument for the try-rescue style error, with additional pointers to the protocol docs. Personally, that is what I would prefer: no stub, but a helpful message.

parkerl:
@tejasbubane let's pull off the stub and go with the try/catch test. Thanks.

exercises/clock/clock.exs
lpil: What is the reason for using `to_string`? This seems like it should be `Clock.format` or similar to me.

martinsvalin: @lpil - Using the String.Chars protocol to give a string representation of a data structure is very common in the Elixir standard library, and generally a good thing to learn about. We did a quick review of other exercises in this track, and found that nowhere did we teach protocols. So it seemed appropriate.

exercises/clock/clock.exs
```diff
@@ -0,0 +1,28 @@
+defmodule Clock do
+  defstruct hour: 0, minute: 0
+
+  @doc """
+  Returns a string representation of a clock:
+
+      iex> Clock.new(8, 9) |> to_string
+      "08:09"
+  """
+  @spec new(integer, integer) :: Clock
```
devonestes: I'm not 100% sure this typespec is valid without defining a type for this module - is it? I'd be happy to see `Clock.t`, but I could also just be wrong on this. I just haven't seen this type of syntax in the standard library, so it struck me as something we might want to look at further.

NobbZ: This spec is totally valid, but it fixes the return value to the atom `:"Elixir.Clock"`.

devonestes: Ahh, yeah, that's kind of what I was saying... @tejasbubane - want to take a swing at implementing at type for this module so these typespecs are doing what we expect them to do?

parkerl:
@tejasbubane any chance you have time to finish this one off? We'd love to merge it.

parkerl:
Thanks. :tada: Merging in order to update.

tejasbubane:
Ahh damn, I missed out on this one. I should open a new PR to fix this.

@parkerl I fear this exercise would get served to users if the submodule in [x-api](https://github.com/exercism/x-api) gets updated.

PR #244, opened 2016-09-13
Use nested accumulate instead of recursive

### LINKED ISSUE #216: Accumulate's recursive function does not actually test recursive mapping
The test here:
https://github.com/exercism/xelixir/blob/master/exercises/accumulate/accumulate_test.exs#L35

Looks like its attempting to check that the `accumulate/2` function will continue mapping if it encounters a list as one of the items in the list. This however is not true since the 2nd level of lists is just created in the anonymous function.

The test IMO should be more like:

```elixir
chars = ~w(a b c)
nums  = ~w(1 2 3)
list = for c <- chars, num <- nums, do: c <> num
expected = [["A1", "A2", "A3"], ["B1", "B2", "B3"], ["C1", "C2", "C3"]]
assert Accumulate.accumulate(chars, &String.upcase/1) == expected
```

@jeg2 I belive you submitted this exercism

tejasbubane:
Fixes #216

Recursive `accumulate` was not actually recursive.
Hence renamed the test description to use `nested` instead and change
test to use `accumulate` over 2 lists.

### DIFF
--- a/exercises/accumulate/accumulate_test.exs
+++ b/exercises/accumulate/accumulate_test.exs
@@ -32,10 +32,10 @@ defmodule AccumulateTest do
   end
 
   @tag :pending
-  test "accumulate recursively" do
+  test "nested accumulate" do
     chars = ~w(a b c)
     nums  = ~w(1 2 3)
-    fun = fn(c) -> for num <- nums, do: c <> num end
+    fun = fn(c) -> Accumulate.accumulate(nums, &(c <> &1)) end
     expected = [["a1", "a2", "a3"], ["b1", "b2", "b3"], ["c1", "c2", "c3"]]
     assert Accumulate.accumulate(chars, fun) == expected
   end


parkerl:
Thanks @tejasbubane, let us take a good look at this one.

dalexj:
looks good

lpil:
:+1:

devonestes:
👍

parkerl:
I'm not convinced this change adds anything to the exercise. To follow up my comments in the original issue, this test should simply pass if the user has the previous tests passing. Given that it will simply go green I'm not sure they will read/understand this test.

devonestes:
@parkerl Are you thinking it would be best to delete that test rather than rename it?

parkerl:
@devonestes I don't know. Honestly, if you look at the tests for this exercise, it could be argued that the third test is fairly meaningless. The fourth and fifth tests will not add anything at all that I can tell. They are not driving out any new behavior. I think the vast majority of users will un-skip these tests, see that they are green, and move on without reading any more deeply into the tests.

We can merge this change but I don't think it addresses these issues.

parkerl:
Ok, merging this as it clarifies this test and my other issues with this exercise are tangential. Thanks @tejasbubane 🎉

Issue #176, opened 2016-04-30
[binary-search]: Current implementation is much less efficient than a linear search


NobbZ:
The conclusion in the subject is from a very brief look at the code.

In general, you can't search a linked list, in a way that is more efficient than a linear search, due to its nature.

To access the element in the middle, you have to iterate the first half of the list completely, while dooing this recursively, you might even read over the searched element multiple times.

Also when splitting the list you do again fly over the first half (even doing a copy when the element is there!)

I already had a similar discussion in a project which aims on implementing some prominent CS datastructures and algorithms over it, perhaps you want to take a read at sashaafm/exads#14 and then deprecate/remove this exercise from this track.

bernardoamc:
You are right, just saw the `fetch` implementation and it is indeed a linear scan. How to proceed so people will not be able to implement this exercise in this track again?

NobbZ:
IIRC there are keys in the JSON which mark exercises as deprecated when trying to move them out of the track or to mark them as not suitable for a track. But I don't remember their exact names. perhaps @kytrinyx can help out here?

NobbZ:
@kytrinyx as we are discussing it here right now, how was the link to check for other tracks having the exercise?  Then I could take a looking there are others around that implement the binary search on an inappropriate datastructure and perhaps file another issue over at x-common if necessary.

kytrinyx:
> IIRC there are keys in the JSON which mark exercises as deprecated when trying to move them out of the track or to mark them as not suitable for a track.

In the config.json there is a key `deprecated` (array of problem slugs) for problems that are already implemented but should not be delivered via `fetch`, and `foregone` for problems that we want to avoid implementing.

> how was the link to check for other tracks having the exercise?

If you look at the output of http://x.exercism.io/v3/problems/binary-search/todo look under the key 'implementations' which lists the links to the directory of the implementation on GitHub.

devonestes:
Can I ask a stupid question - why is the efficiency of the implementation important for our purposes? Personally, I don't see a huge problem in there being some less-than-optimal efficiency in the code that's use here. As long as it's part of a valuable learning experience I think it's just fine.

NobbZ:
to require the student to solve the binary search over a linked list does
imply that it would be in general efficient and the student might think it
were better than a linear search because it's written in the readme.

Devon Estes <notifications@github.com> schrieb am Fr., 5. Aug. 2016 00:22:

devonestes:
[...]

PR #251, opened 2016-09-20
Binary search on tuple input

martinsvalin:
A linked list is a poor data structure for binary search, as pointed out in #176. To do binary search in O(log n) time (which is the whole point of binary search), you need constant time access to the middle element.

To avoid teaching an anti-pattern, let's use tuples as input.

I've removed the test case and behaviour of checking that the input is sorted, raising ArgumentError otherwise. The example implementation compares the list to a sorted list, using `Enum.sort/1` in O(n log n) time. Adding to this, tuples aren't Enumerable, so I would have to convert to a list – O(n), I'd expect. Both steps blows O(log n) out of the water

It's conceptually better to have the user of BinarySearch ensure that the inputs are sorted. I think again it would be an anti-pattern to build checks into a binary search that sabotages the performance for valid input. Also, this check isn't essential to the algorithm this exercise teaches.

### DIFF
--- a/exercises/binary-search/binary_search.exs
+++ b/exercises/binary-search/binary_search.exs
@@ -1,24 +1,24 @@
 defmodule BinarySearch do
   @doc """
-    Searches for a key in the list using the binary search algorithm.
-    It returns :not_found if the key is not in the list.
-    Otherwise returns the tuple {:ok, index}.
+    Searches for a key in the tuple using the binary search algorithm.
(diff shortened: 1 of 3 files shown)


parkerl:
Awesome! ❤️

parkerl:
I am :+1: on this solution and feel this should be merged. @devonestes, @rubysolo any parting thoughts?

devonestes:
This is definitely better than what we had - I'm all for it! 👍

PR #254, opened 2016-09-25
Fix BeerSong string indentation

lpil:
In Elixir multi-line strings trim preceeding spaces, and thus should be
indented as per usual. :)


parkerl:
Great! Thanks :tada:

PR #255, opened 2016-09-25
Replace :pending test tag with idiomatic :skip tag

lpil:
Since Elixir v1.1 the tag `:skip` has been used to skip tests.

If we use the standard here the students will not have to re-learn how
to skip tests when they move onto real Elixir projects. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ The example solution should be named `example.exs`.
 
 - Use typespecs in the example and template files as described [here](http://elixir-lang.org/getting-started/typespecs-and-behaviours.html).
 
-- Each test file should have code like the following at the top of the file. This allows the tests to be run on CI and configures tests to be skipped with the `:pending` flag.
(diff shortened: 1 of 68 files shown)


devonestes:
I'm totally down with this - thanks for the great contribution! 👍 

Since it's a rather significant change I want to leave this open for a couple of days just to make sure no one raises any issues, but I can't see any myself.

parkerl:
I'm :+1: on this change. I think it is great. However, it points out that the Elixir version in mix.exs is out of sync with the README and Travis. Opening an issue on that.

lpil:
I can fix that. Would you like me to resolve it in this pull request or another?

parkerl:
Normally I'd prefer a separate PR but since you have to change it in this one anyway let's do it here. Sound good?

parkerl:
Thanks 🎉

PR #257, opened 2016-09-29
Incorrect markup in markdown exercise

DanCouper:
Currently, the tests and example both specify that the the HTML "bold" tag is `<em>`. This is incorrect; it would actually render as italic. `<strong>` renders as bold.

Markdown parsers render bold as `<strong>`, and italics as `<em>`; this PR brings the exercise in line with this - `<i>` has been changed to `<em>`, `<em>` has been changed to `<strong>`.


yeongsheng-tan:
👍LGTM

devonestes:
Thanks for the catch! This is actually in line with what the exercise describes in the tests in the `x-common` repo at the moment, so it looks like the data there is incorrect as well. @DanCouper - if you're up for it, it would be good to open a PR there as well so others don't make this same mistake when implementing this exercise! Here's a link to that info: https://github.com/exercism/x-common/blob/master/exercises/markdown/canonical-data.json

All looks good to me, but I'll leave it open for another day or so in case anyone has any other questions or comments 👍

DanCouper:
Ach, I forgot to check that, was half asleep this morning, I'll check and PR against that tomorrow when I've got time, then I'll update here depending on feedback. There are a few other things about the test conditions I'd say need changing - I know it's a simplified version of Markdown, but it seems a bit off to have it operate under different rules to the actual thing...I'll have a look at it tomorrow anyway.

devonestes:
@parkerl Yeah, we still need to update this to be a refactoring exercise, but we might as well have correct tests before we do that!

PR #258, opened 2016-10-01
Add documentation about running Dialyzer to TESTS.md

parkerl:
This is what I hope to be a good start on documenting how users can setup Dialyzer. This change is in response to #204. The setup is fairly simple. However, I feel it is important that those new to Elixir know that using Dialyzer is optional.

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -31,3 +31,84 @@ Or, you can enable all the tests by commenting out the
 ```elixir
 # ExUnit.configure trace: true
 ```
+
+### Typespecs and Dialyzer (DIscrepancy AnalYZer for ERlang programs)
+
+Elixir exercises include a skeleton implementation file. This
+file outlines the module and functions that you are expected
+to implement. In most exercises, you will find typespecs
+above the function declaration. These start with the `@spec` tag and
+typically follow the `@spec function_name(type1, type2) :: return_type`
+format. These are used in Elixir and Erlang as documentation and
+in conjunction with a tool called Dialyzer to find type inconsistencies
+and possible bugs. For more information see the
+[Elixir typespecs guide](http://elixir-lang.org/getting-started/typespecs-and-behaviours.html)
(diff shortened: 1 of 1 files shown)


parkerl:
@devonestes if you have a sec could you give this a once over?

docs/TESTS.md
```diff
@@ -31,3 +31,83 @@ Or, you can enable all the tests by commenting out the
...
+to implement. In most exercises, you will find typespecs
+above the function declaration. These start with the `@spec` tag and
+typically follow the `@spec function_name(type1, type2) :: return_type`
+format. These are used in Elixir and Erlang as documentation and
+in conjunction with a tool called Dialyzer to find type inconsistencies
+and possible bugs. For more information see the
+[Elixir typespecs guide](http://elixir-lang.org/getting-started/typespecs-and-behaviours.html)
+or the [typespecs documentation](http://elixir-lang.org/docs/stable/elixir/typespecs.html).
+
+If you are a more experienced Elixir user, you may want to check
```
devonestes: I'm not sure we want to create fear of Dialyzer being only for "more experienced Elixir user"s here. I think we can probably say that Dialyzer can be confusing, but totally learnable!

devonestes:
In general I think this is awesome! The one thing that might make it a little better is links to some more info about Dialyzer specifically. I know there's a link in the Elixir docs in the typespecs section, but maybe we'd want to also add a link to the Dialyzer docs as well (just in case folks don't click on that other link)?

docs/TESTS.md
parkerl: OK. What I feel we need to stress is that unlike the tests, running Dialyzer and getting rid of any warnings is optional. New users should not feel they have to take this step before submitting their solutions.

parkerl:
@devonestes added the link to the dialyzer docs. Also tried to clarify that running Dialyzer is optional. Better?

devonestes:
Looks great! I say 👍 , but I don't think we should merge this until we've merged #237, right?

parkerl:
@devonestes speaking of #237 I think it is close. I'd like to merge both ASAP.

PR #259, opened 2016-10-03
Revert "Replace :pending test tag with idiomatic :skip tag"

parkerl:
This reverts commit 2ae55a4723089984bf02c7e50a01e8ce97ca7e3c.

I'm sorry to say that we missed the fact that tests marked with `skip` are not being run despite configuring CI to include them. I have re-created this behavior in a plain elixir app. As far as I can tell there is no way to un-skip tests marked with `skip`. This seems to be a bug in ExUnit. I'm reverting this change until ExUnit changes or until we find a way to include these tests.

This was good lesson for me to carefully check the build whenever we make sweeping changes to how tests are run. 😞

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ The example solution should be named `example.exs`.
 
 - Use typespecs in the example and template files as described [here](http://elixir-lang.org/getting-started/typespecs-and-behaviours.html).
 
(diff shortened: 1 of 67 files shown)


parkerl:
Merged immediately as our tests are currently not running in CI.

PR #268, opened 2016-10-13
Update binary_search_test.exs : returns :not_found when key is too high

screamingjungle:
Many solutions I've seen fail on this test.


devonestes:
Looks good to me - thanks for the contribution! This case is actually already in the `cannonical_data.json` file and not implemented here, so we needed it!

It looks like you might have some trailing whitespace here, and that's what's failing the CI build. Once you take care of that then this should be good to merge.

parkerl:
Thanks :tada:

PR #271, opened 2016-10-21
Change `markdown` exercise to be a refactoring exercise

### LINKED ISSUE #252: `markdown` should be a refactoring exercise
The current implementation of the `markdown` exercise isn't exactly what is intended in the README, and I worry that folks will be confused. If you take a look at the README, it makes it clear that the user is given a very poorly factored version of the exercise and their job is to refactor it, but that's not really we have here.

For some background, here's the original issue that was raised about adding more refactoring exercises: https://github.com/exercism/discussions/issues/24

And here's the link to the README and such in the `x-common` repo: https://github.com/exercism/x-common/pull/262/files

@yeongsheng-tan, do you want to take a crack at this since you did the initial implementation? There isn't a whole lot more that you'd need to do - just take your existing `example.exs` and make it way worse, then put that in the `markdown.exs` file so that's what's served to users.

If you have any questions, let me know! I'd be happy to help whoever wants to take this on.

devonestes:
The implementation we had for `markdown` wasn't entirely in line with what was
set out in `x-common`, so I've changed it here. Basically I took the original
implemntation that we had in the `example.exs` file and made it a whole lot
worse! One of my favorite ways to make something really bad is to use single
letter variable names instead of descriptive variables, so I've done that. I
also removed a bunch of pipe operators, used inconsistent string interplation
versus string concatenation, and added in some gnarly `if` statements for fun.
The tests all still pass, though!

Let's hope we never see code like this in our "real" lives, but I have to say
it was kind of fun to do! Let me know if this is too cruel and if I should make it a
little easier to handle, cause right now it might be _too_ tough to refactor!

Closes #252

### DIFF
--- a/exercises/markdown/markdown.exs
+++ b/exercises/markdown/markdown.exs
@@ -10,9 +10,66 @@ defmodule Markdown do
     iex> Markdown.parse("#Header!\n* __Bold Item__\n* _Italic Item_")
     "<h1>Header!</h1><ul><li><em>Bold Item</em></li><li><i>Italic Item</i></li></ul>"
   """
-
   @spec parse(String.t) :: String.t
-  def parse(markdown) do
+  def parse(m) do
+    patch(Enum.join(Enum.map(String.split(m, "\n"), fn(t) -> process(t) end)))
+  end
 
+  defp process(t) do
+    if String.starts_with?(t, "#") || String.starts_with?(t, "*") do
+      if String.starts_with?(t, "#") do
+        enclose_with_header_tag(parse_header_md_level(t))
+      else
+        parse_list_md_level(t)
+      end
+    else
+      enclose_with_paragraph_tag(String.split(t))
+    end
+  end
+
+  defp parse_header_md_level(hwt) do
+    [h | t] = String.split(hwt)
+    {to_string(String.length(h)), Enum.join(t, " ")}
+  end
+
+  defp parse_list_md_level(l) do
+    t = String.split(String.trim_leading(l, "* "))
+    "<li>" <> join_words_with_tags(t) <> "</li>"
+  end
+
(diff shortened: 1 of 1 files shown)


yeongsheng-tan:
evil 😬  😰

exercises/markdown/markdown.exs
```diff
@@ -10,9 +10,66 @@ defmodule Markdown do
...
+  defp enclose_with_header_tag({hl, htl}) do
+    "<h" <> hl <> ">" <> htl <> "</h" <> hl <> ">"
+  end
+
+  defp enclose_with_paragraph_tag(t) do
+    "<p>#{join_words_with_tags(t)}</p>"
+  end
+
+  defp join_words_with_tags(t) do
+    Enum.map_join(t, " ", fn(w) -> replace_md_with_tag(w) end)
```
parkerl: Break this into separate `map` and `join` calls?

exercises/markdown/markdown.exs
```diff
@@ -10,9 +10,66 @@ defmodule Markdown do
...
+  end
+
+  defp replace_md_with_tag(w) do
+    replace_suffix_md(replace_prefix_md(w))
+  end
+
+  defp replace_prefix_md(w) do
+    cond do
+      w =~ ~r/^[#{"_"}{1}][^#{"_"}+]/ -> String.replace_prefix(w, "_", "<em>")
```
parkerl: Maybe use `replace/4` instead of giving them `replace_prefix/3` and `replace_suffix/3` right away? Also might give you a chance to duplicate the regular expressions?

exercises/markdown/markdown.exs
devonestes: Good call!

parkerl:
Great! 🎉 ...I mean horrible! 😈

PR #272, opened 2016-10-24
Replaced last trailing space with "\s"

davbre:
Some editors automatically trim whitespace so I replaced the last trailing space with "\s".

### DIFF
--- a/exercises/diamond/diamond_test.exs
+++ b/exercises/diamond/diamond_test.exs
@@ -38,7 +38,7 @@ defmodule DiamondTest do
 \sD     D\s
 \s C   C \s
 \s  B B  \s
-\s   A   \s  
+\s   A   \s
     """
   end
 end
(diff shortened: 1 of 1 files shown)


parkerl:
Nice, that is a good catch. Can you please squash your commits? Thanks!

davbre:
@parkerl - from what I've read you can squash the two commits if/when you decide to merge. Is that the case? See here: [http://stackoverflow.com/a/36377634/10021400](http://stackoverflow.com/a/36377634/1002140).

parkerl:
@davbre I can squash the commits. It does require me (or any maintainer) to fix up the commit message. (Sometimes I merge these simple PRs from a device where editing the message is a pain 😨 ) 

Thanks!
🎉

davbre:
@parkerl - ok, thanks for doing that :+1: (I edited it online so you saved me the pain of cloning the rep locally etc)

PR #274, opened 2016-10-25
Add simple linked list exercise

lpil:
Hey

Had a go at adding this exercise. I adapted the tests from the Haskell & Ruby implementations. What do you think?

I wasn't sure about the config.json bit, so I sorta winged it.

### DIFF
--- a/exercises/simple-linked-list/linked_list.exs
+++ b/exercises/simple-linked-list/linked_list.exs
@@ -1,0 +1,83 @@
+defmodule LinkedList do
+  @opaque t :: tuple()
+
+  @doc """
+  Construct a new LinkedList
+  """
+  @spec new() :: t
+  def new() do
+    # Your implementation here...
+  end
+
+  @doc """
+  Push an item onto a LinkedList
+  """
+  @spec push(t, any()) :: t
+  def push(list, elem) do
+    # Your implementation here...
+  end
+
+  @doc """
+  Calculate the length of a LinkedList
+  """
+  @spec length(t) :: non_neg_integer()
+  def length(list) do
+    # Your implementation here...
+  end
+
+  @doc """
+  Determine if a LinkedList is empty
+  """
+  @spec empty?(t) :: boolean()
+  def empty?(list) do
+    # Your implementation here...
+  end
+
+  @doc """
+  Get the value of a head of the LinkedList
+  """
+  @spec peek(t) :: {:ok, any()} | {:error, :empty_list}
+  def peek(list) do
+    # Your implementation here...
+  end
+
+  @doc """
+  Get tail of a LinkedList
+  """
+  @spec tail(t) :: {:ok, t} | {:error, :empty_list}
+  def tail(list) do
+    # Your implementation here...
+  end
+
+  @doc """
+  Remove the head from a LinkedList
+  """
+  @spec pop(t) :: {:ok, any(), t} | {:error, :empty_list}
+  def pop(list) do
+    # Your implementation here...
+  end
+
+  @doc """
+  Construct a LinkedList from a stdlib List
+  """
+  @spec from_list(list()) :: t
+  def from_list(list) do
+    # Your implementation here...
+  end
+
+  @doc """
+  Construct a stdlib List LinkedList from a LinkedList
+  """
+  @spec to_list(t) :: list()
+  def to_list(list) do
+    # Your implementation here...
+  end
+
+  @doc """
+  Reverse a LinkedList
+  """
+  @spec reverse(t) :: t
+  def reverse(list) do
+    # Your implementation here...
+  end
+end
--- a/exercises/simple-linked-list/linked_list_test.exs
+++ b/exercises/simple-linked-list/linked_list_test.exs
@@ -1,0 +1,154 @@
+if !System.get_env("EXERCISM_TEST_EXAMPLES") do
+  Code.load_file("linked_list.exs", __DIR__)
+end
+
+ExUnit.start
+ExUnit.configure exclude: :pending, trace: true
+
+defmodule LinkedListTest do
+  use ExUnit.Case
+
+  test "length/1 of new list" do
+    list = LinkedList.new()
+    assert LinkedList.length(list) == 0
+  end
+
+  @tag :pending
+  test "empty?/1 of new list" do
+    list = LinkedList.new()
+    assert LinkedList.empty?(list)
+  end
+
+  @tag :pending
+  test "length/1 of list of 1 datum" do
+    list = LinkedList.new() |> LinkedList.push(10)
+    assert LinkedList.length(list) == 1
+  end
+
+  @tag :pending
+  test "empty?/1 of list of 1 datum" do
+    list = LinkedList.new() |> LinkedList.push(20)
+    refute LinkedList.empty?(list)
+  end
+
+  @tag :pending
+  test "peek/1 of list of 1 datum" do
+    list = LinkedList.new() |> LinkedList.push(20)
+    assert LinkedList.peek(list) == {:ok, 20}
+  end
+
+  @tag :pending
+  test "peek/1 of list of empty list" do
+    list = LinkedList.new()
+    assert LinkedList.peek(list) == {:error, :empty_list}
+  end
+
+  @tag :pending
+  test "tail/1 of empty list" do
+    list = LinkedList.new()
+    assert {:error, :empty_list} = LinkedList.tail(list)
+  end
+
+  @tag :pending
+  test "tail/1 of list of 1 datum" do
+    list = LinkedList.new() |> LinkedList.push(:hello)
+    assert {:ok, tail} = LinkedList.tail(list)
+    assert LinkedList.peek(tail) == {:error, :empty_list}
+  end
+
+  @tag :pending
+  test "pushed items are stacked" do
+    list =
+      LinkedList.new()
+      |> LinkedList.push(:a)
+      |> LinkedList.push(:b)
+    assert LinkedList.peek(list) == {:ok, :b}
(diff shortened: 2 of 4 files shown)


exercises/simple-linked-list/linked_list_test.exs
```diff
@@ -0,0 +1,154 @@
...
+  end
+
+  @tag :pending
+  test "empty?/1 of list of 1 datum" do
+    list = LinkedList.new() |> LinkedList.push(20)
+    refute LinkedList.empty?(list)
+  end
+
+  @tag :pending
+  test "peek/1 of list of 1 datum" do
```
devonestes: I think one other assertion here - or maybe in a different test if you like - to make sure that `peek` isn't actually implemented like `pop`. Maybe something like:

```elixir
assert LinkedList.peek(list) == {:ok, 20}
assert LinkedList.length(list) == 1
```

devonestes:
Awesome - thanks for this!

For the `config.json`, I think what you have there is fine for now. If you really wanted to, you could add a difficulty rating (from 1-10) that you feel fits this exercise. For reference, check out #276 to see what I've considered other exercises (although those are just mine and I'd love other feedback).

I have a couple small nit picks, but I think this generally looks great!

exercises/simple-linked-list/linked_list_test.exs
```diff
@@ -0,0 +1,154 @@
...
+  @tag :pending
+  test "peek/1 of list of empty list" do
+    list = LinkedList.new()
+    assert LinkedList.peek(list) == {:error, :empty_list}
+  end
+
+  @tag :pending
+  test "tail/1 of empty list" do
+    list = LinkedList.new()
+    assert {:error, :empty_list} = LinkedList.tail(list)
```
devonestes: Why do you use this assertion style (relying on pattern matching rather than equality) here? It seemed oddly inconsistent to me, and I worry it might confuse others as well. I expected it to look pretty much the same as the assertion on line 43 above.

lpil: No reason that I can remember. I suspect this is a typo. :)

exercises/simple-linked-list/linked_list_test.exs
lpil: Ah yes, you're right. I'll do that.

exercises/simple-linked-list/linked_list.exs
```diff
@@ -0,0 +1,83 @@
+defmodule LinkedList do
+  @opaque t :: tuple()
+
+  @doc """
+  Construct a new LinkedList
+  """
+  @spec new() :: t
+  def new() do
+    # Your implementation here...
```
devonestes: We don't really put these comments in the functions to be implemented. For some reason many folks don't delete them when they get around to doing the implementations.

lpil: I don't like these at all. I would prefer not to supply the file at all so the user has to write their own module.

Shall I just remove the comments?

devonestes: Yeah, we remove the comments and leave the function that we expect them to implement just empty (along with the spec).

I mostly agree with you in general that it would be preferable to not give the user anything so they have to "completely" TDD their solution. I'm sure that somewhere before that discussion has been had and people came up with really good reasons for doing it this way, so I just roll with it for now 😉 

One reason that I think is really good - especially in Elixir - is that we can give them the typespec for each function that we expect them to implement when we give them the file. I think that alone makes it all worth it. I can see how getting a user to implement those on their own would be good, but I can't think of tests that would drive them to write those typespecs. Is there a way to test that a typespec is implemented for a function?

lpil:
I've updated the PR. How does it look now? Thanks

lpil:
I feel there is one problem though. In order to get CI to pass I had to add
specific type specs for the functions which forces them to implement this
using the same data structures that I have used.

Or perhaps in the template file I should change the type of the list to
any() ?

exercises/simple-linked-list/linked_list_test.exs
```diff
@@ -0,0 +1,155 @@
...
+  test "empty?/1 of list of 1 datum" do
+    list = LinkedList.new() |> LinkedList.push(20)
+    refute LinkedList.empty?(list)
+  end
+
+  @tag :pending
+  test "peek/1 of list of 1 datum" do
+    list = LinkedList.new() |> LinkedList.push(20)
+    assert LinkedList.peek(list) == {:ok, 20}
+    assert LinkedList.length(list) == 0
```
devonestes: Ok, you'll have to bear with me since my data structures may be a little weak, but isn't `peek` typically non-destructive in a linked list? So in this case, after peeking at the first element in the list, the length of that list should remain unchanged - in this case, 1?

I also might be totally wrong, so forgive me if that's the case!

devonestes:
@lpil Yeah, that's tension between a really specific typespec and it being too restrictive for the users is always a tricky line to walk! In this case, I'm cool with `any()` for the type since I tend to err on the side of giving users more flexibility in implementation, but I'd be curious to see what other folks think!

exercises/simple-linked-list/linked_list_test.exs
lpil: You're right. A mistake from trying to make these changes while in a rush! I'm a little confused that CI didn't error here. Did it run?

I'll fix this now.

lpil:
Also it could be interesting to see a version of this exercise where the use has to implement the `Enumberable` protocol for this exercise :)

exercises/simple-linked-list/linked_list_test.exs
lpil: Actually this test is pointless as the data structure is immutable.

exercises/simple-linked-list/linked_list_test.exs
devonestes: I don't think it did run... I've been noticing some odd occasions where a push doesn't trigger a CI run in GitHub recently, but only with open source projects. I'm not sure where the fault lies, but I've reported it to GitHub. Hopefully they can figure it out!

I first noticed it during the DDoS attack earlier and figured it was related to that, but it's still happening, so hopefully it is just a quick thing they can fix.

exercises/simple-linked-list/linked_list.exs
```diff
@@ -0,0 +1,73 @@
+defmodule LinkedList do
+  @opaque t :: any()
+
+  @doc """
+  Construct a new LinkedList
```
parkerl: I think using the module name is a bit confusing in this and the following comments. Perhaps, 

"Construct an empty linked list"

exercises/simple-linked-list/linked_list.exs
```diff
@@ -0,0 +1,73 @@
...
+  """
+  @spec new() :: t
+  def new() do
+  end
+
+  @doc """
+  Push an item onto a LinkedList
```
parkerl: "Push an item onto the front of a linked list" ???

exercises/simple-linked-list/linked_list.exs
```diff
@@ -0,0 +1,73 @@
...
+
+  @doc """
+  Determine if a LinkedList is empty
+  """
+  @spec empty?(t) :: boolean()
+  def empty?(list) do
+  end
+
+  @doc """
+  Get the value of a head of the LinkedList
```
parkerl: "...value of the head of..."???

exercises/simple-linked-list/linked_list.exs
```diff
@@ -0,0 +1,73 @@
...
+
+  @doc """
+  Get tail of a LinkedList
+  """
+  @spec tail(t) :: {:ok, t} | {:error, :empty_list}
+  def tail(list) do
+  end
+
+  @doc """
+  Remove the head from a LinkedList
```
parkerl: "Remove the head from a linked list, and return the value and the tail of the linked list"???

exercises/simple-linked-list/linked_list.exs
```diff
@@ -0,0 +1,73 @@
...
+
+  @doc """
+  Construct a LinkedList from a stdlib List
+  """
+  @spec from_list(list()) :: t
+  def from_list(list) do
+  end
+
+  @doc """
+  Construct a stdlib List LinkedList from a LinkedList
```
parkerl: "Construct a stdlib List from a LinkedList"

lpil:
@parkerl The names are taken more or less directly from the names in other courses. Happy to update them to be more descriptive :)

parkerl:
@lpil sorry...missed your response. Yes, I would prefer if they were more descriptive. How do you feel about wrapping them in `describe`? Like this exercise https://github.com/exercism/xelixir/blob/a327ebce7b51fdd3bc4cb468295510487ea9c150/exercises/custom-set/custom_set_test.exs 

I know @devonestes favors this style.

Thanks.

lpil:
Hi @parkerl , I've updated the PR.

martinsvalin:
This build is failing due to a warning (we're running the tests with `--warnings-as-errors`). The warning is:

    warning: opaque type t() is underspecified and therefore meaningless
    exercises/simple-linked-list/example.exs:2

The type in question is defined both in `example.exs` and the `linked_list.exs` stub:

    @opaque t :: any()

Maybe we could have a type that allows several possible implementations, but not `any()`? Like `tuple() | list() | map() | struct()`.

lpil:
Oops, didn't notice that. I'll have another look.

lpil:
Huh, I forgot to push this ~1 month ago. oops.

devonestes:
> Huh, I forgot to push this ~1 month ago. oops.

d'oh

I hate it when I do that!

parkerl:
Thanks @lpil I think we have nitpicked this enough. 🎉

PR #275, opened 2016-10-25
Remove obsolete problems key

ErikSchierboom:
The `problems` key has been obsoleted as per [this issue](https://github.com/exercism/x-api/pull/137). This has already been succesfully implemented by the [C#](https://github.com/exercism/xcsharp#130), [F#](https://github.com/exercism/xfsharp#224) and [Haskell](https://github.com/exercism/xhaskell/pull/399) tracks, so API-wise this is a safe action.

Removing the `problems` key also makes it very impossible for people to only add a new exercise in the old, obsolete `problems` key.


ErikSchierboom:
You're welcome!

PR #277, opened 2016-10-27
Update phone_number_test.exs

v-kolesnikov:
Add some asserts

### DIFF
--- a/exercises/phone-number/phone_number_test.exs
+++ b/exercises/phone-number/phone_number_test.exs
@@ -38,6 +38,11 @@ defmodule PhoneTest do
   end
 
   @tag :pending
+  test "invalid with correct number of characters but some are letters" do
+    assert Phone.number("1a2a3a4a5a") == "0000000000"
+  end
+
+  @tag :pending
   test "area code" do
     assert Phone.area_code("1234567890") == "123"
   end


devonestes:
Thanks for taking the time to submit this!

I have a feeling that once the tests before these are passing that these would also be green, so I don't think they would really add anything towards the user's implementation of this exercise. Can you provide some example where the previous tests might have passed but these would have failed?

v-kolesnikov:
http://exercism.io/submissions/b0863e23bbd448999229f3db41a38dee
http://exercism.io/submissions/da352d5dfe6441cda4da17437a225f4f
http://exercism.io/submissions/d2650f74b3b441f4a15e0286febf24d9
and more...

devonestes:
Thanks for the examples! The third one of those actually passes these new tests, but the first two do fail them, so I agree that this is a good addition since there are folks that are answering this in a way that fails this.

I have just a couple little nit picks and then this should be good to go!

exercises/phone-number/phone_number_test.exs
```diff
@@ -38,6 +38,16 @@ defmodule PhoneTest do
   end
 
   @tag :pending
+  test "invalid when proper number of digits but letters mixed in and length is 10" do
```
devonestes: This description and the next one are a little misleading since there aren't the `proper number of digits`. We'd need 10 digits, but there are only 5 here.

Something like `invalid with correct number of characters but some are letters` might be more clear.

exercises/phone-number/phone_number_test.exs
```diff
@@ -38,6 +38,16 @@ defmodule PhoneTest do
...
+    assert Phone.number("1a2a3a4a5a") == "0000000000"
+  end
+
+  @tag :pending
+  test "invalid when proper number of digits but letters mixed in and length is 11" do
```
devonestes: Sorry to be so questioning about this, but do you think it's possible that something could pass the test above but fail this test? We've been doing a lot of cleanup of the test suites recently, removing redundant tests, and I worry this might be one of them.

devonestes:
Looks good to me! I'm just going to leave it open a little longer to see if anyone has any other comments 👍

parkerl:
Thanks 🎉

PR #278, opened 2016-10-27
Reference latest book version: Programming Elixir 1.3

mariusbutuc:
The older versions of Programming Elixir are out of print. 

To save a couple of clicks and a few seconds of figuring out what's going on, reference the latest version directly.

/review @kytrinyx, @parkerl, @devonestes


devonestes:
Awesome - thanks! 👍

PR #279, opened 2016-10-27
Add topics for first couple of exercises

devonestes:
In the issue it recommended that we add topics for each exercise one by one,
but for these first ones I think we can chunk them up a little more, so I'm
doing like 5 or 6 at a time for each PR. I may even do more per PR if folks
feel that it's a lot to review so many PRs - just let me know!

### DIFF
--- a/config.json
+++ b/config.json
@@ -74,36 +74,49 @@
       "slug": "hello-world",
       "difficulty": 1,
       "topics": [
+        "strings"
       ]
     },
     {
       "slug": "nucleotide-count",
       "difficulty": 1,
       "topics": [
+        "maps",
+        "string processing",
+        "enumerables"
       ]
     },
     {
       "slug": "space-age",
       "difficulty": 1,
       "topics": [
+        "pattern matching",
+        "math"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -74,36 +74,49 @@
       "slug": "hello-world",
       "difficulty": 1,
       "topics": [
+        "IO",
```
parkerl: I don't think there is any IO in this one?

config.json
```diff
@@ -74,36 +74,49 @@
...
+        "strings"
       ]
     },
     {
       "slug": "nucleotide-count",
       "difficulty": 1,
       "topics": [
+        "maps",
+        "string processing"
```
parkerl: Also `enumerables`?

config.json
devonestes: Oh, yeah, in the Elixir version we don't actually output that to the console. Good catch!

config.json
devonestes: seems like a good addition to me!

parkerl:
looks good! 🎉

PR #286, opened 2017-01-16
Add parentheses to zero arity functions

techfused:
Resolves the ambiguity warnings in Elixir 1.4


devonestes:
Thanks for the contribution! 👍

PR #287, opened 2017-01-17
Fix all deprecation warnings from running `mix test` with Elixir 1.4.0

martinsvalin:
I looked at #286 and noticed there are other deprecation warnings from Elixir 1.4. Most are on zero-arity functions, some on unused module attributes, and a few remaining deprecations to use `MapSet` instead of `Set`.


devonestes:
Awesome - thanks!! 👍

PR #289, opened 2017-01-23
Add tests for exception cases to match canonical-data.json

bunnymatic:
To bring the existing test suite up to speed with the [shared canonical data version](https://github.com/exercism/x-common/blob/master/exercises/grains/canonical-data.json) I added 3 exception tests for squares -1, 0 and 65.

### DIFF
--- a/exercises/grains/grains_test.exs
+++ b/exercises/grains/grains_test.exs
@@ -16,57 +16,57 @@ defmodule GrainsTest do
 
   # @tag :pending
   test "square 1" do
-    assert Grains.square(1) === 1
+    assert Grains.square(1) === { :ok, 1 }
   end
 
   @tag :pending
   test "square 2" do
-    assert Grains.square(2) === 2
+    assert Grains.square(2) === { :ok, 2 }
   end
 
   @tag :pending
   test "square 3" do
-    assert Grains.square(3) === 4
+    assert Grains.square(3) === { :ok, 4 }
   end
 
   @tag :pending
   test "square 4" do
-    assert Grains.square(4) === 8
+    assert Grains.square(4) === { :ok, 8 }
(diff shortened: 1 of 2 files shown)


parkerl:
Hi @bunnymatic ! Thanks for the PR! We have had quite a bit of discussion in the past about the most idiomatic way to present error cases for this track. There is a discussion here https://github.com/exercism/xelixir/issues/224. Basically, we have been leaning toward the {:ok, result}, {:error, "message"} idiom. See the triangle exercise https://github.com/exercism/xelixir/blob/2a4dc4eeb14188c1e87315e690808c94d844f1ea/exercises/triangle/triangle_test.exs.

bunnymatic:
I like it.  Much more elixir-y

exercises/grains/grains_test.exs
```diff
@@ -16,57 +16,57 @@ defmodule GrainsTest do
...
   @tag :pending
   test "total grains" do
-    assert Grains.total === 18446744073709551615
+    assert Grains.total === { :ok, 18446744073709551615 }
   end
 
   @tag :pending
   test "square greater than 64 raises an exception" do
-    catch_error Grains.square(65)
+    assert Grains.square(65) === { :error, "The requested square must be between 1 and 64 (inclusive)" }
```
bunnymatic: I wondered about whether or not these tests should care what the error message is.  They probably could also be written something like
```
{ response_type, mesage } = Grains.square(65)
assert(response_type) === :error
```

Also, I'm still new to elixir so if there is a more succinct way of pattern matching that `===` with just the `:error`, just give me a pointer on how that would be done.

parkerl:
Awesome! Thanks 🎉

PR #290, opened 2017-01-24
beer-song: Update tests to match the other tracks tests

bunnymatic:
* verse(99) now produces text starting with "99 beers on the wall"
* updated example.exs


parkerl:
Awesome! Thanks so much! 🎉

PR #291, opened 2017-01-31
Diamond: Cleaned up tests expected data

Xehanort94:
[Diamond - Test]
- Replaced heredocs with String concatenation:
 * Able to remove \s while retaining trimming whitespace editor fix
 * No newline added at the end, as heredocs do
- Removed leading whitespace for the middle rows

### DIFF
--- a/exercises/diamond/diamond_test.exs
+++ b/exercises/diamond/diamond_test.exs
@@ -11,34 +11,32 @@ defmodule DiamondTest do
   # @tag :pending
   test "letter A" do
     shape = Diamond.build_shape(?A)
-    assert shape == "A\n"
+    assert shape == "A"
   end
 
   @tag :pending
   test "letter C" do
     shape = Diamond.build_shape(?C)
-    assert shape == """
-\s A \s
-\sB B\s
- C   C
-\sB B\s
-\s A \s
-    """
+    assert shape ==
+      "  A  \n" <>
+      " B B \n" <>
+      "C   C\n" <>
+      " B B \n" <>
+      "  A  "
   end
 
   @tag :pending
   test "letter E" do
     shape = Diamond.build_shape(?E)
-    assert shape == """
-\s   A   \s
-\s  B B  \s
-\s C   C \s
-\sD     D\s
- E       E
-\sD     D\s
-\s C   C \s
-\s  B B  \s
-\s   A   \s
-    """
+    assert shape ==
+      "    A    \n" <>
+      "   B B   \n" <>
+      "  C   C  \n" <>
+      " D     D \n" <>
+      "E       E\n" <>
+      " D     D \n" <>
+      "  C   C  \n" <>
+      "   B B   \n" <>
+      "    A    "
   end
 end
--- a/exercises/diamond/example.exs
+++ b/exercises/diamond/example.exs
@@ -5,22 +5,24 @@ defmodule Diamond do
   """
   @spec build_shape(char) :: String.t
   def build_shape(letter) do
-    pos = letter - 64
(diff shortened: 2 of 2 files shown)


devonestes:
@Xehanort94 Thanks for the contribution! One thing that we try and do when possible is to try and make non-breaking changes to test suites. Because the behavior in the test suite is now different (by removing the tailing newline), that means anyone who has completed this exercise in the past will no longer have a functional implementation.

In general, if we're updating a test suite, we shouldn't have to update the example implementation or else that's probably a sign that we're making a breaking change...

What exactly prompted this pull request? Did the existing test suite seem incorrect to you? I'm not sure I see a need for the change, but I might be missing something.

Thanks again!

Xehanort94:
No need to merge my PR if it doesn't conform to your updating rules.

I think the test suite is not reflecting the exercises description given in the readme:
  - The readme doesn't list a linebreak at the end of the returned string
  - There should not be a whitespace before the first character in the middle row

NobbZ:
As there is no canonical data for diamond, I took a look at the other implementations and tests and these are my observations:

* C#: Tests are completely agnostic to surrounding whitespace as I read them, also there is no trainling newline generated. But I'm not quite sure if I read the tests correctly. I will ignore this track in summary
* ecmascript: Tests the trailing newline
* elixir: Currently tests trailing newline, but has a PR which removes that
* F#: I'm totally unable to read the tests :) But example seems NOT to generate trailing newline. But I'm really not sure how F#s `List.reduce` works, since there is no “starting value” given
* go: I'm not sure about the tests, but the example does emit the trailing newline
* java: circumvents the problem by expecting a list of strings. One element per line.
* javascript: tests trailing newline
* lua: tests trailing newline
* python: tests trailing newline
* ruby: tests trailing newline

| expect trailing | expect **no** trailing | list of rows | unsure |
|:-:|:-:|:-:|:-:|
| 6 | 0 | 1 | 2 |

Xehanort94:
The F# solution doesn't produce a trailing newline. The `reduce` function works the same as Elixirs `Enum.reduce/2`, taking the first element of the enumerable as the initial accumulator value.

Also i did a bit of testing, and it turns out, that the leading whitespace for the middle row is cut, if the heredocs ending `"""` is indented by one or more spaces. Interestingly enough this would also truncate the leading whitespace for the other rows (up until the indentation level of the ending `"""`, but not truncating actual characters), but that is circumvented by using `\s` instead of a simple space for the first character.

So if you guys decide that the trailing newline is fine you can close the PR, since my second adressed point is actually fine.

Well, tbh i think that the heredoc is even more confusing now since it uses the indentation and the escaped whitespace tricks that are nowhere documented.

Xehanort94:
To illustrate my point, type the two versions into iex and observe the difference (I run Elixir 1.4 with OTP19):

# Closing triple quotes not indented
"""
\s A \s
\sB B\s
 C   C
\sB B\s
\s A \s
"""

# Closing triple quotes indented
"""
\s A \s
\sB B\s
 C   C
\sB B\s
\s A \s
    """

The first one will add the leading whitespace to the resulting string, while the second one won't. Data is taken directly from the test case.

devonestes:
@Xehanort94 So first off, thanks for pointing out the way to strip leading whitespace from a heredoc! That is indeed sadly undocumented from what I can see (and that could be a good opportunity to contribute some documentation there if you like?)

Given that there isn't yet cannonical data for this exercise, and that our implementation is mostly in line with other languages, I think we'll keep the exercise as is for now.

Thanks again, though, and if you wanted to create the cannonical data for this exercise, I'm sure that would be a very welcome contribution!

PR #293, opened 2017-02-04
Make run-length-encoding exercise expected values consistent with README

### LINKED ISSUE #292: run-length-encoding tests conflict with description of exercise in README
I noticed that the expected results in the run-length-encoding tests suggest that the solution [should translate a single letter like `H` as `1H`](https://github.com/exercism/xelixir/blob/master/exercises/run-length-encoding/rle_test.exs#L20-L28).

However, the [README description](https://github.com/exercism/x-common/blob/master/exercises/run-length-encoding/description.md) suggests that the solution should leave single letters alone, i.e. `H` remains `H`.

I'd be happy to make a PR to fix this in the test cases and example solution -- just wanted to check first to make sure that's how we want to resolve this.

daveyarwood:
Fixes #292.

### DIFF
--- a/exercises/run-length-encoding/example.exs
+++ b/exercises/run-length-encoding/example.exs
@@ -3,12 +3,19 @@ defmodule RunLengthEncoder do
   @spec encode(String.t) :: String.t
   def encode(string) do
     Regex.scan(~r/([A-Z])\1*/, string)
-    |> Enum.map_join(fn([run, c]) -> "#{String.length(run)}#{c}" end)
+    |> Enum.map_join(fn([run, c]) ->
+      times = String.length(run)
+      number = if times == 1 do "" else times end
+      "#{number}#{c}"
+    end)
   end
 
   @spec decode(String.t) :: String.t
   def decode(string) do
-    Regex.scan(~r/(\d+)(.)/, string)
-    |> Enum.map_join(fn [_,n,c] -> String.duplicate(c, String.to_integer(n)) end)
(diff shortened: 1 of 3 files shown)


devonestes:
@daveyarwood This is great! I just noticed, though, that there is a `cannonical_data.json` file for this exercise, so while we're making this breaking change, we might as well try and make this implementation line up with that set of tests. Here's the link to that file: https://github.com/exercism/x-common/blob/master/exercises/run-length-encoding/canonical-data.json

In general these tests are enough to get folks to the correct implementation so there aren't duplicate tests that will all pass at once. It would be great if you wanted to make the tests for this exercise line up with that file!

daveyarwood:
@devonestes Done!

There is one test case that doesn't work because the input string doesn't contain exclusively uppercase letters. I've filed https://github.com/exercism/x-common/issues/537 to address that in canonical-data.json.

devonestes:
@daveyarwood Thanks for the quick work on that! I checked out that issue, and from the conversation there it looks like we'll need to implement the ability to keep track of spaces, as well as uppercase and lowercase letters. However, I have a feeling that once that bit is ready then this can be merged!

daveyarwood:
@devonestes I've adjusted the example solution so that it works with lowercase letters and whitespace, including the last test case in canonical-data.json.

I also submitted a PR to x-common to adjust the description there, so things should be consistent once both PRs are merged.

devonestes:
@daveyarwood Awesome work - thanks! 👍

PR #294, opened 2017-02-09
Add exercise 'series'

DoggettCK:
Implementing the 'series' exercise: http://exercism.io/exercises/elixir/series

Chose to go with an empty list for slices larger than the input string size, rather than raise an error like the Ruby version, for example, since it seemed logical to say "there are no slices of size 7 in a string of size 6".

### DIFF
--- a/exercises/series/series.exs
+++ b/exercises/series/series.exs
@@ -4,7 +4,7 @@ defmodule StringSeries do
   of that size. If `size` is greater than the length of `s`, or less than 1,
   return an empty list.
   """
-  @spec slices(s :: String.t(), size :: integer) :: any
+  @spec slices(s :: String.t(), size :: integer) :: list(String.t())
   def slices(_s, _size) do
   end
 end
--- a/exercises/series/series.exs
+++ b/exercises/series/series.exs
@@ -1,0 +1,11 @@
+defmodule StringSeries do
+  @doc """
+  Given a string `s` and a positive integer `size`, return all substrings
+  of that size. If `size` is greater than the length of `s`, or less than 1,
+  return an empty list.
+  """
+  @spec slices(s :: String.t(), size :: integer) :: any
+  def slices(_s, _size) do
+  end
+end
+
--- a/exercises/series/series_test.exs
+++ b/exercises/series/series_test.exs
@@ -1,0 +1,57 @@
+if !System.get_env("EXERCISM_TEST_EXAMPLES") do
+  Code.load_file("series.exs", __DIR__)
+end
+
+ExUnit.start
+ExUnit.configure exclude: :pending, trace: true
+
+defmodule StringSeriesTest do
+  use ExUnit.Case
+
+  test "slices of 1" do
+    assert StringSeries.slices("01234", 1) == ["0", "1", "2", "3", "4"]
+    assert StringSeries.slices("92834", 1) == ["9", "2", "8", "3", "4"]
+  end
+
+  @tag :pending
+  test "slices of 2" do
+    assert StringSeries.slices("01234", 2) == ["01", "12", "23", "34"]
+    assert StringSeries.slices("98273463", 2) == ["98", "82", "27", "73", "34", "46", "63"]
+    assert StringSeries.slices("37103", 2) == ["37", "71", "10", "03"]
+  end
+
+  @tag :pending
+  test "slices of 3" do
+    assert StringSeries.slices("01234", 3) == ["012", "123", "234"]
+    assert StringSeries.slices("31001", 3) == ["310", "100", "001"]
+    assert StringSeries.slices("982347", 3) == ["982", "823", "234", "347"]
+  end
+
+  @tag :pending
+  test "slices of 4" do
+    assert StringSeries.slices("01234", 4) == ["0123", "1234"]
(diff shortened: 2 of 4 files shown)


devonestes:
Thanks for adding this - new exercises are always really helpful!

Since we know there's an error case to handle here, how would you feel about an API for this function that follows the `{:ok, response}` and `{:error, reason}` pattern that's so common? That would make more sense to me if we truly want that to be an exceptional case. I know there have been lots of conversations around how to handle error cases in exercises here since there are three different (and all valid) possibilities. I'd love to hear what others think as well!

DoggettCK:
@devonestes So maybe `{:ok, response}` for positive (even for greater than string size), and `{:error, reason}` for negative, or would the error case cover the too-large size as well?

I'm fine with either, but only the negative number case will actually cause an error when passed to `Enum.chunk`. Since that returns an empty list when the chunk size is too big, thought I'd follow that example.

devonestes:
Ahhh, I didn't know `Enum.chunk` returns an empty list when the chunk size is too big - that changes my original thinking!

I think then that what you have above is a good call. Let's go ahead with that for now and see what other folks chime in with (if anything).

exercises/series/series_test.exs
```diff
@@ -0,0 +1,57 @@
...
+    assert StringSeries.slices("01234", 1) == ["0", "1", "2", "3", "4"]
+    assert StringSeries.slices("92834", 1) == ["9", "2", "8", "3", "4"]
+  end
+
+  @tag :pending
+  test "slices of 2" do
+    assert StringSeries.slices("01234", 2) == ["01", "12", "23", "34"]
+    assert StringSeries.slices("98273463", 2) == ["98", "82", "27", "73", "34", "46", "63"]
+    assert StringSeries.slices("37103", 2) == ["37", "71", "10", "03"]
+  end
```
parkerl: Is there a difference between these 3 assertions? If so should they be separate tests?

exercises/series/series_test.exs
```diff
@@ -0,0 +1,57 @@
...
+
+  @tag :pending
+  test "slices same size as string" do
+    assert StringSeries.slices("01234", 5) == ["01234"]
+    assert StringSeries.slices("31001", 5) == ["31001"]
+    assert StringSeries.slices("982347", 6) == ["982347"]
+  end
+
+  @tag :pending
+  test "slices bigger than string" do
```
parkerl: Could you make the test name more descriptive? Thanks.

exercises/series/series_test.exs
```diff
@@ -0,0 +1,57 @@
...
+    assert StringSeries.slices("01234", 6) == []
+    assert StringSeries.slices("31001", 6) == []
+    assert StringSeries.slices("982347", 7) == []
+  end
+
+  @tag :pending
+  test "negative-sized slices" do
```
parkerl: Could you make this test name more descriptive as well?

exercises/series/series_test.exs
DoggettCK: Not really. They were 3 separate tests I copied from the Ruby version and condensed into one.

exercises/series/series.exs
```diff
@@ -0,0 +1,11 @@
+defmodule StringSeries do
+  @doc """
+  Given a string `s` and a positive integer `size`, return all substrings
+  of that size. If `size` is greater than the length of `s`, or less than 1,
+  return an empty list.
+  """
+  @spec slices(s :: String.t(), size :: integer) :: any
```
devonestes: For our type spec, it looks like the return type of this function could be more specific since our tests are always returning a list. Might as well be specific here, right?

DoggettCK: Good catch. Missed that when I copied this over from my earlier work on robot-simulator.

devonestes:
Looks great to me now! I'll just leave this open for another day or so to see if anyone else wants to chime in. 👍

exercises/series/series_test.exs
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  @tag :pending
+  test "slices same size as string" do
+    assert StringSeries.slices("01234", 5) == ["01234"]
+  end
+
+  @tag :pending
+  test "Unicode characters count as a single character" do
+    assert StringSeries.slices("José", 1) == ["J", "o", "s", "é"]
```
parkerl: Love the nod to José

exercises/series/series_test.exs
```diff
@@ -0,0 +1,52 @@
...
+  Code.load_file("series.exs", __DIR__)
+end
+
+ExUnit.start
+ExUnit.configure exclude: :pending, trace: true
+
+defmodule StringSeriesTest do
+  use ExUnit.Case
+
+  test "slices of size 1" do
```
parkerl: we usually have  a commented out pending tag here to demonstrate how to enable the next tests.` # @tag :pending`

parkerl:
@DoggettCK do you have a sec to take care of my one last nitpick? Great job.

DoggettCK:
@perkerl Done. Would've been sooner, but birthday weekend.

devonestes:
Awesome, thanks!

PR #303, opened 2017-03-01
New test to change exercise and solution for it.

weapp:
There is a missing case when solution involves multiples currencies.

I added a new test and I change the solution for pass new testcase.

```elixir
iex(2)> Change.generate(21, [10, 2, 5])
{:ok, %{2 => 3, 5 => 1, 10 => 1}}
```

### DIFF
--- a/exercises/change/change_test.exs
+++ b/exercises/change/change_test.exs
@@ -36,6 +36,12 @@ defmodule ChangeTest do
   end
 
   @tag :pending
+  test "generatings change using all coins" do
+    change = %{2 => 3, 5 => 1, 10 => 1}
+    assert Change.generate(21, [10, 2, 5]) == {:ok, change}
+  end
+
+  @tag :pending
   test "generates the same change given any coin order" do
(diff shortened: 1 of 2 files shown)


devonestes:
@weapp Thanks for the PR!

This exercise actually has a standardized set of tests that should be implemented in all language tracks (which you can see [here](https://github.com/exercism/x-common/blob/master/exercises/change/canonical-data.json)). Want to open a PR adding this test case there first? If it's added then we can go ahead and add it in here, but for consistency's sake we always try and just implement what's in that `cannonical_data.json` file in each language track.

weapp:
Ok, sorry @devonestes, I didn't know this file.

But, I think that the current tests don't match with this `.json`. I should change all tests?

devonestes:
@weapp Absolutely! If you see inconsistencies between our implementation and any of those `cannonical_data.json` files, pull requests updating our implementation are much appreciated!!

And I think your other test is a good one to add, by the way. It should just be discussed and added as part of that `cannonical_data.json` first 😉

PR #305, opened 2017-03-03
Updated change exercise to adjust to canonical data.

weapp:
There were some tests that differ in many ways with canonical data, like "change with Lower Elbonia Coins". That can't pass with the previous example.exs.

Furthermore, the results of canonical data are implemented with a list, not with a hash / map.

### Notes

Some tests have been lost in the migration because they aren't present in the canonical data.
This is correct? Should I add here again? Maybe in both sides?

### DIFF
--- a/exercises/change/change_test.exs
+++ b/exercises/change/change_test.exs
@@ -8,43 +8,70 @@ ExUnit.configure exclude: :pending, trace: true
 defmodule ChangeTest do
   use ExUnit.Case
 
-  test "returns :error on empty list" do
-    assert Change.generate(1, []) == :error
+  # @tag :pending
+  test "no coins make 0 change" do
+      coins = [1, 5, 10, 21, 25]
+      expected = []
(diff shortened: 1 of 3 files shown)


exercises/change/change_test.exs
```diff
@@ -8,43 +8,70 @@ ExUnit.configure exclude: :pending, trace: true
 defmodule ChangeTest do
   use ExUnit.Case
 
-  test "returns :error on empty list" do
-    assert Change.generate(1, []) == :error
+  # @tag :pending
+  test "no coins make 0 change" do
+      coins = [1, 5, 10, 21, 25]
```
parkerl: We indent 2 spaces. Thanks.

parkerl: Also, as long as we are updating to match the canonical examples we should confirm to their order as well.

parkerl:
Thanks 🎉

PR #310, opened 2017-03-07
Implement rotational-cipher exercise

DoggettCK:


exercises/rotational-cipher/rotational_cipher_test.exs
```diff
@@ -0,0 +1,80 @@
...
+  test "rotate a by 26, same output as input" do
+    plaintext = "a"
+    shift = 26
+    assert RotationalCipher.rotate(plaintext, shift) == "a"
+  end
+
+  @tag :pending
+  test "rotate a by 0, same output as input" do
+    plaintext = "a"
+    shift = 26
```
devonestes: This looks like maybe a typo? Should the `shift` here be 0?

DoggettCK: I agree. Actually, it's a typo in the canonical test data: https://github.com/exercism/x-common/blob/master/exercises/rotational-cipher/canonical-data.json#L21. I just took that and made a quick vim macro to convert each test's hash into an Elixir test.

I'll fix the shift to match the test description, since that seems clearer than understanding that 26 is the same as 0 due to modulus.

devonestes: No wonder these tests looked so good and accurate - yay for macros! Might as well open a PR to fix the canonical data as well if you like. Your call - I'd be happy to handle it if you don't want the bother.

DoggettCK:
Went ahead and added a PR to x-common to fix the canonical test data

exercises/rotational-cipher/rotational_cipher.exs
```diff
@@ -0,0 +1,13 @@
+defmodule RotationalCipher do
+  @doc """
+  Given a plaintext and amount to shift by, return a rotated string.
+
+  Example:
+  iex> RotationalCipher.rotate("Attack at dawn", 13)
+  "Nggnpx ng qnja"
+  """
+  @spec rotate(text :: String.t(), shift :: integer) :: String.t()
+  def rotate(_text, _shift) do
```
parkerl: I'm not sure how I feel about underscoring the variables in the template file. @devonestes ?

DoggettCK: @parkerl I'm not a fan of it myself, but I want to say CI was complaining about it in the past. Maybe it was just the warning in iex.

devonestes: I don't like it either, but I didn't feel strongly enough about it to warrant asking for a change. If it doesn't mess with CI, then I'd be happy to see them gone.

DoggettCK: Guess it was just warnings in iex after all. Removed them from this and poker, and the warning in iex at least tells users there's something they need to fix.

PR #312, opened 2017-03-10
Implement Diffie-Hellman key exchange exercise

DoggettCK:
I'm not sure what the policy on helper modules/functions is, but after attempting to duplicate the C#, F#, and Go tests, I realized that doing (A^B)%C with 1024-bit numbers is incredibly slow, despite Elixir's support for arbitrarily large integers. Left it running for 2 days, no signs of stopping.

Attempted to use https://en.wikipedia.org/wiki/Modular_exponentiation#Memory-efficient_method for the functionality, with a few `IO.inspect`s thrown in, and that one literally runs `B` times, so for numbers so big we don't officially have names for them, it was going to take awhile. The method right below that one, however, does some fancy math I wouldn't expect users to try to figure out, but runs almost instantaneously, so I included it, and enough documentation of how D-H works that it'd push the `Helper` module off the bottom of most editors, and hopefully they'll try `A |> :math.pow(B) |> round |> rem(C)` first.

Alternatively, I could remove the very large integer tests, but the other implementations have them with built-in functions that didn't exist in Elixir, and I think they're useful in understanding cryptography.

### DIFF
--- a/exercises/diffie-hellman/diffie_hellman.exs
+++ b/exercises/diffie-hellman/diffie_hellman.exs
@@ -26,6 +26,9 @@ defmodule DiffieHellman do
   As long as their private keys are never lost or transmitted, only they know
   their private keys, so even if Eve has P, G, and both public keys, she can't
   do anything with them.
+
+  A video example is available at:
+  https://www.khanacademy.org/computing/computer-science/cryptography/modern-crypt/v/diffie-hellman-key-exchange-part-2
   """
 
   @doc """
--- a/exercises/diffie-hellman/diffie_hellman.exs
+++ b/exercises/diffie-hellman/diffie_hellman.exs
@@ -1,0 +1,93 @@
+defmodule DiffieHellman do
+  @moduledoc """
+  Diffie-Hellman is a method of securely exchanging keys in a public-key
+  cryptosystem. Two users, Alice and Bob, want to share a secret between
+  themselves, while ensuring nobody else can read it.
+
+  Step 0: Alice and Bob agree on two prime numbers, P and G. An attacker, Eve,
+  can intercept these numbers, but without one of Alice or Bob's private keys,
+  we'll see Eve can't do anything useful with them.
+
+  Step 1: Alice and Bob each generate a private key between 1 and P-1.
+  P).
+
+  Step 2: Using the initial primes P and G, Alice and Bob calculate their
+  public keys by raising G to the power of their private key, then calculating
+  the modulus of that number by P. ((G**private_key) % P)
+
+  Step 3: Alice and Bob exchange public keys. Alice and Bob calculate a secret
+  shared key by raising the other's public key to the power of their private
+  key, then doing a modulus of the result by P. Due to the way modulus math
+  works, they should both generate the same shared key.
+
+  Alice calculates: (bob_public ** alice_private) % P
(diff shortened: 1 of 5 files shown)


DoggettCK:
Turns out the `:crypto` Erlang library already implements `mod_pow`, so I removed the `Helper` module and modified the documentation to point users in the right direction when they run into the enormous integer limitations.

exercises/diffie-hellman/diffie_hellman.exs
```diff
@@ -0,0 +1,66 @@
...
+  def generate_private_key(prime_p) do
+  end
+
+  @doc """
+  Given two prime integers as generators (`prime_p` and `prime_g`), and a private key,
+  generate a public key using the mathematical formula:
+
+  (prime_g **  private_key) % prime_p
+
+  HINT: Erlang's :crypto module has a useful function for finding the modulus of a power,
```
parkerl: Hey @DoggettCK this might belong in a `HINTS.md` file for this exercise as described here https://github.com/exercism/exercism.io/blob/master/docs/fixing-exercise-readmes.md

exercises/diffie-hellman/diffie_hellman_test.exs
```diff
@@ -0,0 +1,98 @@
...
+end
+
+ExUnit.start
+ExUnit.configure exclude: :pending, trace: true
+
+defmodule DiffieHellmanTest do
+  use ExUnit.Case
+
+  #@tag :pending
+  test "private key should be random between 1 and P-1, inclusive" do
```
parkerl: I'd like to see a test prior to this one for just one generated value.

exercises/diffie-hellman/diffie_hellman_test.exs
```diff
@@ -0,0 +1,98 @@
...
+    prime_p = 23
+
+    generated_private_keys = 1..100 |> Enum.map(fn _i -> DiffieHellman.generate_private_key(prime_p) end)
```
parkerl: Cold this be moved to a function on `DiffieHellman` and driven out with a test? If you don't agree with that, then I suggest extracting it to a helper function.

DoggettCK: This was basically just to call the function 100 times and ensure some randomness to it. Didn't need to be in two tests, so I rewrote it a bit after changing the tests to:

- Does it generate a number between 1-(P-1)? (Should handle your previous request)
- Does it handle arbitrarily large P?
- Is the number random?

Now that there's only one test using it, I cleared up what it's doing by not doing an `Enum.map` over an unused number, but calling `Stream.repeatedly` and taking 100 digits. I think it's clearer, and can move it to a helper if you still want, but don't think it belongs in the DiffieHellman class itself.

I also moved the hints to `HINTS.md`, so that should cover your concerns, I hope.

parkerl:
Nice @DoggettCK ! This cleanup looks great. @devonestes see anything else?

devonestes:
I like that this is an exercise that really requires you to dip into Erlang. Seems like more and more often that's a thing that an Elixir dev has to do.

It all looks good to me, but then again I know nothing of crypto, so don't take my word for the correctness of any of this stuff!

Huge thanks for this and all your recent contributions @DoggettCK

parkerl:
Thanks @DoggettCK 🎉

PR #314, opened 2017-03-15
2 unit tests to enfroce handling of whitespace

lex57ukr:
The README file for the run-length-encoding exercise has this line: 
> If the string contains any whitespace, it should be passed through unchanged

However, there's no unit test to enforce this rule. The `decode(encode(...)) combination` test doesn't quite cut it because cycling accepts encoded spaces as long as the resulting string gets decoded properly.

The example implementation passes these two new tests without any changes.

### DIFF
--- a/exercises/run-length-encoding/rle_test.exs
+++ b/exercises/run-length-encoding/rle_test.exs
@@ -13,8 +13,28 @@ defmodule RunLengthEncoderTest do
   end
 
   @tag :pending
-  test "encode single characters only" do
+  test "encode single characters only are encoded without count" do
     assert RunLengthEncoder.encode("XYZ") === "XYZ"
+  end
+
+  @tag :pending
+  test "encode string with no single characters" do
+    assert RunLengthEncoder.encode("AABBBCCCC") == "2A3B4C"
+  end
+
+  @tag :pending
+  test "encode single characters mixed with repeated characters" do
+    assert RunLengthEncoder.encode("WWWWWWWWWWWWBWWWWWWWWWWWWBBBWWWWWWWWWWWWWWWWWWWWWWWWB") === "12WB12W3B24WB"
+  end
+
+  @tag :pending
+  test "encode multiple whitespace mixed in string" do
+    assert RunLengthEncoder.encode("  hsqq qww  ") === "2 hs2q q2w2 "
+  end
+
+  @tag :pending
+  test "encode lowercase characters" do
+    assert RunLengthEncoder.encode("aabbbcccc") === "2a3b4c"
   end
 
   @tag :pending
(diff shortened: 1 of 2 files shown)


devonestes:
It does look like our test suite is missing some tests - thanks for pointing that out! However, for each exercise there is a file that contains the "canonical data" that each language track is supposed to implement. It appears that we're missing some of those tests in that file (found here: https://github.com/exercism/x-common/blob/master/exercises/run-length-encoding/canonical-data.json).

It would be great if you wanted to get us up to date with that file by adding those missing tests. Yours are pretty much testing the same thing, but for consistency's sake we should keep the same test data and order that's given in that file.

Thanks again!

lex57ukr:
@devonestes - check it out. The exercise is up to date with its canonical data. I had to fix the exercise.exs to pass all canonical tests.

The autobuild complains about some diamond test exiting with 1. What's up with that?

lex57ukr:
Here's another interesting piece of log for 1.3.2:
```
Location: https://raw.githubusercontent.com/danielberkompas/travis_elixir_plts/master/elixir-1.3.2_19.0.plt [following]
--2017-03-17 03:48:15--  https://raw.githubusercontent.com/danielberkompas/travis_elixir_plts/master/elixir-1.3.2_19.0.plt
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 151.101.32.133
Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|151.101.32.133|:443... connected.
HTTP request sent, awaiting response... 503 Connection refused
2017-03-17 03:48:15 ERROR 503: Connection refused.
```

The build environment is definitely sick

devonestes:
@lex57ukr Ahh, I see what's happening there. It takes _forever_ to build the PLTs for Dialyzer, so we're using some that have very nicely been pre-built for us and hosted on Github. It looks like for some reason, when your branch got pushed and CI got kicked off, that Github wasn't able to serve that file. I just restarted the CI build, and it looks like everything's good now!

exercises/run-length-encoding/rle_test.exs
```diff
@@ -28,27 +48,27 @@ defmodule RunLengthEncoderTest do
...
   @tag :pending
-  test "decode with single values" do
-    assert RunLengthEncoder.decode("12WB12W3B24WB") === "WWWWWWWWWWWWBWWWWWWWWWWWWBBBWWWWWWWWWWWWWWWWWWWWWWWWB"
+  test "decode lower case string" do
+    assert RunLengthEncoder.decode("2a3b4c") === "aabbbcccc"
   end
 
   @tag :pending
-  test "decode(encode(...)) combination" do
+  test "consistency encode followed by decode gives original string" do
```
devonestes: @parkerl So what are your thoughts on taking the exact descriptions from the `canonical_data.json` file? In general I think it's good to stay consistent, but this one just reads so weird that I think it might be worth changing.

lex57ukr:
@devonestes - I updated the comments. I like using canonical descriptions (when possible) for improved readability. Thanks for restarting the build. That was odd.

parkerl:
+1

PR #316, opened 2017-03-22
Implement scale-generator exercise

DoggettCK:
The common documentation doesn't tell you much, and there are no canonical exercises, so these are entirely based off the C# and Ruby versions, which for some reason that's not explained in the docs drop the last note, so if you're doing an `Enum.map` over the major scale pattern, you'd expect the first example to give

`["C", "D", "E", "F", "G", "A", "B", "C"]`

But the other implementations drop the final `"C"`, so to match them, I did as well.

I gave this a 5 for difficulty, but feel free to suggest higher or lower as you feel appropriate.

### DIFF
--- a/exercises/scale-generator/scale_generator.exs
+++ b/exercises/scale-generator/scale_generator.exs
@@ -1,0 +1,51 @@
+defmodule ScaleGenerator do
+  @doc """
+  The chromatic scale is a musical scale with twelve pitches, each a semitone
+  (half-tone) above or below another.
+
+  C C# D D# E F F# G G# A A# B
+
+  Notes with a sharp (#) are half a tone (a semitone) higher than the note
+  below them, where the next letter note is a full tone except in the case of B
+  and E, which have no sharps.
+
+  Sharp notes can also be considered the flat (b) note of the tone above them,
+  so the chromatic scale above can also be represented as:
+
+  C Db D Eb E F Gb G Ab A Bb B
+
+  Certain scales will require the use of the flat version, depending on the
+  `tonic` (key) that begins them, which is C in the above examples.
+
+  If you are asked to find a scale for any of the following tonics, use the
+  flat chromatic scale:
+
+  F Bb Eb Ab Db Gb d g c f bb eb
+
+  For all others, use the regular chromatic scale.
+
+  The `pattern` string will let you know how many semitones to step over to get
+  the next note in your scale:
+
+  "m": one semitone
+  "M": two semitones (full tone)
+  "A": augmented second (three semitones)
+
+  For example, a C Major scale will receive the pattern "MMmMMMm", which
+  indicates you will start with C, make a full step over C# to D, another over
+  D# to E, then a semitone, stepping from E to F (again, E has no sharp). You
+  can follow the rest of the pattern to get:
+
+  C D E F G A B C
+
+  For the purposes of these exercises, we don't repeat the tonic, so after you
+  have found your 12 notes starting with your tonic, once you run out of notes,
+  you can stop, leaving you with:
+
+  C D E F G A B
+  """
+  @spec scale(tonic :: String.t(), pattern :: String.t()) :: list(String.t())
+  def scale(tonic, pattern) do
+  end
+end
+
--- a/exercises/scale-generator/scale_generator.exs
+++ b/exercises/scale-generator/scale_generator.exs
@@ -3,33 +3,62 @@ defmodule ScaleGenerator do
   The chromatic scale is a musical scale with twelve pitches, each a semitone
   (half-tone) above or below another.
 
-  C C# D D# E F F# G G# A A# B
+  Notes with a sharp (#) are a semitone higher than the note below them, where
+  the next letter note is a full tone except in the case of B and E, which have
+  no sharps.
 
-  Notes with a sharp (#) are half a tone (a semitone) higher than the note
-  below them, where the next letter note is a full tone except in the case of B
-  and E, which have no sharps.
+  Generate these 12 notes, starting with the given `tonic` and wrapping back
+  around to the note before it. If the `tonic` is lowercase, capitalize it.
 
+  "C" should generate: ~w(C C# D D# E F F# G G# A A# B)
+  """
+  @spec chromatic_scale(tonic :: String.t()) :: list(String.t())
+  def chromatic_scale(tonic \\ "C") do
+  end
+
+  @doc """
   Sharp notes can also be considered the flat (b) note of the tone above them,
-  so the chromatic scale above can also be represented as:
+  so the 12 notes can also be represented as:
 
-  C Db D Eb E F Gb G Ab A Bb B
+  A Bb B C Db D Eb E F Gb G Ab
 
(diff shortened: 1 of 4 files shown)


exercises/scale-generator/scale_generator_test.exs
```diff
@@ -0,0 +1,94 @@
...
+  @locrian_scale_pattern         "mMMmMMM"
+  @harmonic_minor_scale_pattern  "MmMMmAm"
+  @octatonic_scale_pattern       "MmMmMmMm"
+  @hexatonic_scale_pattern       "MMMMMM"
+  @pentatonic_scale_pattern      "MMAMA"
+  @enigmatic_scale_pattern       "mAMMMmm"
+
+  #@tag :pending
+  test "C Major scale" do
+    assert ScaleGenerator.scale("C", @major_scale_pattern) == ~w(C D E F G A B)
```
parkerl: I know you copied this from other tracks. However, I feel that these tests force the user to go from 0 to 60. One of the goals of an exercism exercise is to re-create the TDD experience by carefully building up behavior in each subsequent test. Are there some intermediate tests we could add to help folks get to the solution more iteratively?

DoggettCK: @parkerl I totally agree. You're right, they're based on the C#, F#, and Ruby ones, which just dump you into it, assuming you know enumeration, rotating lists, etc... The base description in x-common just throws in a lot of terms without explaining them, so I studied some music theory to try to understand it in the function documentation.

I'll try to go through and introduce tests like stepping up a semitone, tone, or accidental second from a given note, and before that maybe generating the 12 notes in the chromatic scale starting from a given note (defaulting to C).

I still don't understand why it drops the last note on all of them, but Wikipedia appeared to agree with dropping it, which seems to defeat the purpose of the last interval. [Another site](https://www.basicmusictheory.com/c-5th-intervals#st2) appears to include it, which makes much more sense to me, and uses "W" for whole steps and "H" for half-steps, which seems more readable to me than "M" and "m".

Actually, now that I look again, Wikipedia does include the last C from the example: https://en.wikipedia.org/wiki/Scale_(music)#Scales.2C_steps.2C_and_intervals, so perhaps it's the other implementations that are incorrect.

DoggettCK:
@parkerl OK, rewrote it to start with more basic things like generating the 12 notes starting with a given note, determine whether to use the sharp or flat scale based on the tonic, and stepping over an interval to get the next note based on potential pattern characters. I stuck with m/M/A for the patterns, unless you think H/W/A is more understandable based on my last comment.

parkerl:
Great. I'm on the road for a few days and can't properly review this. I will get back as soon as I can.

devonestes:
Sweet - for once my two degrees in classical music are good for something!

So, technically an octave scale does include the repeated tonic up an octave (assuming it's an ascending scale). Scales are defined by their intervals, and all the scales in this exercise span an octave, so we need the repeat of the tonic to complete the scale. Some scales span shorter or longer distances, but no need to go down that path 😛 

The more common notation for intervals is indeed 'd', 'm', 'M', and 'A' (although since we're only dealing with 2nds we don't need 'd'), so I think keeping that is a good idea.

I think it might make sense to start the tests with that `step` function. That seems really small and self contained, and something that would be used in later examples.

I'll go through and add a few more comments. This is a sneakily tricky one to implement if you don't have some pretty serious understanding of music theory!

exercises/scale-generator/scale_generator_test.exs
```diff
@@ -0,0 +1,276 @@
...
+  @lydian_scale_pattern          "MMMmMMm"
+  @phrygian_scale_pattern        "mMMMmMM"
+  @locrian_scale_pattern         "mMMmMMM"
+  @harmonic_minor_scale_pattern  "MmMMmAm"
+  @octatonic_scale_pattern       "MmMmMmMm"
+  @hexatonic_scale_pattern       "MMMMMM"
+  @pentatonic_scale_pattern      "MMAMA"
+  @enigmatic_scale_pattern       "mAMMMmm"
+
+  describe "generate chromatic scale" do
```
devonestes: We're missing the 12th semitone at the end of these chromatic scales.

exercises/scale-generator/scale_generator_test.exs
```diff
@@ -0,0 +1,276 @@
...
+  use ExUnit.Case
+
+  @major_scale_pattern           "MMmMMMm"
+  @minor_scale_pattern           "MmMMmMM"
+  @dorian_scale_pattern          "MmMMMmM"
+  @mixolydian_scale_pattern      "MMmMMmM"
+  @lydian_scale_pattern          "MMMmMMm"
+  @phrygian_scale_pattern        "mMMMmMM"
+  @locrian_scale_pattern         "mMMmMMM"
+  @harmonic_minor_scale_pattern  "MmMMmAm"
```
devonestes: Part of me thinks that if we're going to include the harmonic minor scale, we should also include the melodic minor scale. They're frequently taught together, and both used frequently.

That pattern is "MmMMMMm", and an example based on C would be "C D Eb F G A B C".

exercises/scale-generator/scale_generator_test.exs
DoggettCK: So should the first test be:

`assert ScaleGenerator.chromatic_scale("A") == ~w(A A# B C C# D D# E F F# G G# A)`?

exercises/scale-generator/scale_generator_test.exs
DoggettCK: I'll add that one in. Took all these from the C#/Ruby versions, which were already incorrect in other ways, and they didn't implement that one, nor is it mentioned in the x-common description.

DoggettCK:
@devonestes I'll move the step tests to the beginning. I can also add 'd' to the steps, if you'll let me know how big the interval should be.

exercises/scale-generator/scale_generator_test.exs
devonestes: Yep- exactly!

devonestes:
@DoggettCK We don't need the diminished interval here since we're dealing with scales, and a diminished 2nd is just the same not. A major 2nd is two semitones, a minor 2nd is 1 semitone, and a diminished 2nd is 0 semitones.

Thanks for taking care of the other stuff, though!

exercises/scale-generator/scale_generator_test.exs
DoggettCK: Done, and moved the step function first in scale_generator.exs to match the test order.

exercises/scale-generator/scale_generator.exs
```diff
@@ -0,0 +1,83 @@
+defmodule ScaleGenerator do
+  @doc """
+  Find the note a given interval (`step`) in `scale` after the `tonic`.
```
devonestes: This reads a little wonky to me. Should it maybe be something like "Find the note for a given interval (`step`) in a `scale` after the `tonic`"?

DoggettCK: Agreed, and done.

parkerl:
Nice! 🎉

PR #320, opened 2017-03-28
Implement Secret Handshake exercise

DoggettCK:


exercises/secret-handshake/HINTS.md
```diff
@@ -0,0 +1 @@
+use Bitwise (or div/rem)
```
devonestes: Given the README, it seems to me like the purpose of this exercise for most folks is to practice binary representations of decimal numbers (and the associated binary functions in a given language), so maybe in the HINTS file we should just push towards the `Bitwise` operators? Maybe we could even recommend `&&&` specifically?

DoggettCK: Rewrote it to use Bitwise and expanded the hints.

PR #321, opened 2017-03-28
Implement Strain exercise

DoggettCK:


exercises/strain/HINTS.md
```diff
@@ -0,0 +1 @@
+`apply` will let you pass arguments to a function
```
devonestes: Great hint here!

devonestes: Or, maybe we could also include the other way that you can call an anonymous function: `fun.(head)`, which is the same in this case as `apply`, and probably a little more beginner friendly for this early in the track.

exercises/strain/strain_test.exs
```diff
@@ -0,0 +1,95 @@
...
+
+defmodule StrainTest do
+  use ExUnit.Case
+
+  defp is_odd(n), do: rem(n, 2) == 1
+  defp is_even(n), do: rem(n, 2) == 0
+
+  #@tag :pending
+  test "empty keep" do
+    assert Strain.keep([], fn e -> e < 10 end) == []
```
devonestes: What would you think about naming a `noop` function or something similar for these tests that are intended to be noops in the same way we've named `is_odd` and `is_even`?

exercises/strain/strain_test.exs
```diff
@@ -0,0 +1,95 @@
...
+  Code.load_file("strain.exs", __DIR__)
+end
+
+ExUnit.start
+ExUnit.configure trace: true, exclude: :pending
+
+defmodule StrainTest do
+  use ExUnit.Case
+
+  defp is_odd(n), do: rem(n, 2) == 1
```
devonestes: Since these are functions that return a bool, how about ending them with `?` to follow the convention in the Standard Library?

exercises/strain/strain_test.exs
DoggettCK: Done with all of your suggestions. Thanks!

PR #326, opened 2017-04-04
Implement 12 Days of Christmas

DoggettCK:


exercises/twelve-days/twelve_days.exs
```diff
@@ -0,0 +1,61 @@
...
+  @doc """
+  Given a `number`, return the word representation of it, or "a" for one.
+
+  1 => "a",
+  2 => "two",
+  3 => "three",
+  etc...
+  """
+  @spec count(number :: integer) :: String.t()
+  def count(number) do
```
devonestes: This feels to me like it's too specifically guiding folks towards an implementation. I'd prefer something a little more rough grained here.

exercises/twelve-days/twelve_days.exs
```diff
@@ -0,0 +1,61 @@
...
+  @doc """
+  Given a `number`, return the ordinal representation of it.
+
+  1 => "first",
+  2 => "second",
+  3 => "third",
+  etc...
+  """
+  @spec ordinal(number :: integer) :: String.t()
+  def ordinal(number) do
```
devonestes: Same thing here - this feels more like implementation specifics than actual usable behavior.

exercises/twelve-days/twelve_days.exs
```diff
@@ -0,0 +1,61 @@
...
+  @doc """
+  Given a `number`, return the gift for that specific day, along with its count.
+
+  1 => "a Partridge in a Pear Tree",
+  6 => "six Geese-a-Laying",
+  8 => "eight Maids-a-Milking",
+  etc...
+  """
+  @spec gift(number :: integer) :: String.t()
+  def gift(number) do
```
devonestes: And here, too. I think the only two things that really _need_ to be public functions that we need to make folks implement should be `verse/1`, `verses/2` and `sing/0`.

exercises/twelve-days/twelve_days_test.exs
```diff
@@ -0,0 +1,329 @@
...
+      assert TwelveDays.gift(11) == "eleven Pipers Piping"
+    end
+
+    @tag :pending
+    test "drummers drumming" do
+      assert TwelveDays.gift(12) == "twelve Drummers Drumming"
+    end
+  end
+
+  describe "verse" do
```
devonestes: I don't think we need tests for every verse in here. Maybe we can test the first verse, the last verse, and one more in the middle?

exercises/twelve-days/twelve_days.exs
parkerl: Yup, I think you are correct.

DoggettCK:
@devonestes @parkerl Agreed on too many tests. Removed all but `verse/1`, `verses/2`, and `sing/0`, and all corresponding tests.

config.json
```diff
@@ -74,6 +74,12 @@
       ]
     },
     {
+      "slug": "twelve-days",
+      "difficulty": 2,
```
devonestes: I think this is probably slightly more difficult than a 2. Maybe you could bump this up a bit, and also add a topic or two? I'm having a hell of a time thinking of good topics for this, though... If I think of anything, I'll come back with them!

PR #327, opened 2017-04-05
bracket-push: sync test cases with canonical x-common repo

lex57ukr:
The tests were out of sync with canonical x-common repo. I fixed that and ran all checks to make sure the corresponding example still works.

I added one more test case (I also created a PR to x-common for this test case). This case helps with broken stack-based implementations.


lex57ukr:
@devonestes -- [PR 747](https://github.com/exercism/x-common/pull/747) was merged ~11 days ago.

Thanks!

PR #329, opened 2017-04-06
Implement Tournament exercise

DoggettCK:
This one kinda goes 0-60 like one of my earlier ones, but I'm basing it off the canonical examples. 

I've chosen to implement the example with a string => 5-tuple map, but it could've been done using `defstruct` as well. I just didn't want to force the user to use a particular implementation.

I think it would be a good way to teach structs, and can change it if necessary. Just let me know if you'd prefer it that way.

### DIFF
--- a/exercises/tournament/tournament_test.exs
+++ b/exercises/tournament/tournament_test.exs
@@ -1,0 +1,96 @@
+if !System.get_env("EXERCISM_TEST_EXAMPLES") do
+  Code.load_file("tournament.exs", __DIR__)
+end
+
+ExUnit.start
+ExUnit.configure trace: true, exclude: :pending
+
+defmodule TournamentTest do
+  use ExUnit.Case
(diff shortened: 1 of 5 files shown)


exercises/tournament/tournament_test.exs
```diff
@@ -0,0 +1,96 @@
...
+    Blithering Badgers             |  3 |  0 |  1 |  2 |  1
+    Devastating Donkeys            |  3 |  0 |  1 |  2 |  1
+    """ |> String.trim
+
+    assert Tournament.tally(input) == expected
+  end
+
+  @tag :pending
+  test "mostly invalid lines" do
+    # Invalid input lines in an otherwise-valid game still results in valid
```
parkerl: Ugg, I really don't think this comment is needed but I guess I will have to take that up in x-common.

PR #330, opened 2017-04-10
Implement Matrix exercise

DoggettCK:
I wanted to use this one as an excuse to teach structs, but didn't want to tie them to using a particular internal implementation, except for when rows/columns are output, which is why the example has `matrix` and `transposed_matrix`, where the one the users see only has `matrix`, and the test just checks that it's not nil, and that `to_string` returns the same string representation as was passed in.

### DIFF
--- a/exercises/matrix/example.exs
+++ b/exercises/matrix/example.exs
@@ -1,17 +1,17 @@
-defmodule Matrix do
+defmodule MatrixStruct do
   defstruct matrix: nil, transposed_matrix: nil
 
   @doc """
   Convert an `input` string, with rows separated by newlines and values
-  separated by single spaces, into a `Matrix` struct.
+  separated by single spaces, into a `MatrixStruct` struct.
   """
-  @spec from_string(input :: String.t()) :: %Matrix{}
+  @spec from_string(input :: String.t()) :: %MatrixStruct{}
   def from_string(input) do
     rows = input
            |> String.split("\n", trim: true)
            |> Enum.map(&parse_line/1)
 
-    %Matrix{
+    %MatrixStruct{
       matrix: rows,
       transposed_matrix: rows |> List.zip |> Enum.map(&Tuple.to_list/1)
     }
@@ -27,8 +27,8 @@ defmodule Matrix do
   Write the `matrix` out as a string, with rows separated by newlines and
   values separated by single spaces.
   """
-  @spec to_string(matrix :: %Matrix{}) :: String.t()
-  def to_string(%Matrix{ matrix: rows }) do
+  @spec to_string(matrix :: %MatrixStruct{}) :: String.t()
+  def to_string(%MatrixStruct{ matrix: rows }) do
     rows
     |> Enum.map(&join_row/1)
     |> Enum.join("\n")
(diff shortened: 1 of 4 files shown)


exercises/matrix/example.exs
```diff
@@ -0,0 +1,63 @@
+defmodule MatrixStruct do
```
devonestes: What would you think about calling this module just `Matrix`? I'm cool with assuming that we'll use a struct, but the naming of the module might be a little confusing I think.

exercises/matrix/matrix_test.exs
```diff
@@ -0,0 +1,66 @@
...
+defmodule MatrixTest do
+  use ExUnit.Case
+
+  @input "1 2 3\n4 5 6\n7 8 9"
+
+  #@tag :pending
+  test "reading from and writing to string" do
+    matrix = MatrixStruct.from_string(@input)
+
+    assert matrix != nil
```
devonestes: I'm not 100% sure that these tests will be be more clarifying rather than confusing. I see where you're going there, but frankly I think we probably just need the assertion currently on line 24 for this first test.

exercises/matrix/example.exs
DoggettCK: It initially was called Matrix, but that failed CI because the saddle-points exercise has its module named Matrix for some reason. That should probably be changed, so this can be just Matrix, but if you have a better name in the meantime, I'm happy to change it.

exercises/matrix/matrix_test.exs
DoggettCK: Agreed, and honestly, it doesn't need to even test them on from_string. For this exercise, we could just pass the string to `rows` instead of a `MatrixStruct`, and let them parse it then and return the nested lists. My thought on that was that for typical uses of matrices, you're going to want them to stick around and not re-parse. That led to using a struct, and not wanting to force them to use a particular structure.

That said, I do wind up insisting that no matter the struct, they output lists, so I might as well just have `rows`/`columns` take the string and output the expected row/rows. What would you think about that?

exercises/matrix/example.exs
DoggettCK: I just added a PR to rename the `Matrix` class in saddle-points to `SaddlePoints` (https://github.com/exercism/xelixir/pull/332), so if that's accepted, I can rename this to `Matrix`.

exercises/matrix/matrix_test.exs
devonestes: Well, I actually like the reading from and writing to a string, but I think we can have just a really minimal setup for that test. We're already giving some hints that we'd like them to use a struct to represent this data in the typespecs, so I don't have a problem being really broad here. Something like this should be sufficient

```elixir
matrix = Matrix.from_string(@input)
assert Matrix.to_string(matrix) == @input
```

I like the other tests as well which sort of guide the implementation without being too specific. It should end up being a really good exercise, I think!

DoggettCK:
@devonestes Fixed your request for the test, and renamed the module to `Matrix`, but it'll rely on https://github.com/exercism/xelixir/pull/332 before it passes the CI build.

PR #331, opened 2017-04-13
Implement Simple Cipher exercise

DoggettCK:


exercises/simple-cipher/simple_cipher_test.exs
```diff
@@ -0,0 +1,117 @@
...
+
+defmodule SimpleCipherTest do
+  use ExUnit.Case
+
+  describe "random key generation" do
+    #@tag :pending
+    test "random key is all lowercase letters" do
+      key = SimpleCipher.key
+
+      assert key != nil
```
devonestes: I think this test is kind of superfluous since the regexp on the next line will only match if there is one or more letter, right? Think we still need this?

exercises/simple-cipher/simple_cipher_test.exs
```diff
@@ -0,0 +1,117 @@
...
+    end
+
+    @tag :pending
+    test "only lowercase a-z are translated, rest are passed through" do
+      assert SimpleCipher.encode("this is a test!", "d") == "wklv lv d whvw!"
+      assert SimpleCipher.decode("wklv lv d whvw!", "d") == "this is a test!"
+    end
+
+    @tag :pending
+    test "decoding an encoded text with the same key returns original text" do
```
devonestes: Do you think this test is helpful given how thoroughly the function is tested in other places? I can't think of a way a user could make the test on line 78 pass and have this one fail. Can you?

exercises/simple-cipher/simple_cipher_test.exs
DoggettCK: I didn't have this line at first, because the next line does indeed fail, but the test failure was an actual exception when trying to pass nil to the regex match, instead of a test failure, so I thought it'd be less confusing to users this way.

Actually, now that I've had a day to think about it, I don't like the key generation tests at all. They sort of presuppose a language where you're going to store the key, not pass it in to other functions. It doesn't fit in with the rest of the tests at all, because they have a known key, so they can test for known outputs. I could seed the RNG, but then what's the point? If it's supposed to be 100 random characters, how do you test that randomness?

The result is never actually used anywhere else in the tests, so unless you object, I think I'm going to get rid of those tests and the function in general.

exercises/simple-cipher/simple_cipher_test.exs
DoggettCK: You're right, I can't think of one. We've already implicitly proven in almost every previous test that `text |> encode(key) |> decode(key) == text`, so this holdover from the Ruby version is unnecessary.

PR #332, opened 2017-04-14
Rename Matrix class to SaddlePoints

DoggettCK:
Trying to implement the 'matrix' exercise with a class named Matrix
caused CI failures, as it was already implemented here.

### DIFF
--- a/exercises/saddle-points/example.exs
+++ b/exercises/saddle-points/example.exs
@@ -1,4 +1,4 @@
-defmodule Matrix do
+defmodule SaddlePoints do
   @doc """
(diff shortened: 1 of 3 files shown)


DoggettCK:
Also, this probably could've been avoided by replacing line 21 of .travis.yml with:

```
- for dir in `ls -d exercises/*/`; do elixirc --warnings-as-errors $dir/example.exs $dir/*test.exs -o ./tmp > /dev/null; done
```

That way, each directory's compiler check is independent of any others. Not sure what happens with the return code if one in the middle fails, so this may have been considered before and not used.

PR #335, opened 2017-05-06
Consider allowed range for first digit in area & exchange code

fhdhsni:
First digit in area code and exchange code should be [2-9]
Modified all the tests accordingly


parkerl:
@fhdhsni thanks for this PR! Unfortunately it looks the the current implementation in `exercises/phone-number/example.exs` does not work correctly with these changes. Could you fix up that file as well?

fhdhsni:
@parkerl It should be fine now.

PR #340, opened 2017-05-26
atbash-cipher: add missing tests from canonical data

mhinz:
This PR updates the exercise to version 1.0.0 of its canonical data.

It adds the missing decode() function and the corresponding tests.


martinsvalin:
LGTM. ⭐️ for renaming `cipher/1`, when it gets used for both encoding and decoding.

PR #346, opened 2017-06-17
introduce collatz conjecture exercise

Tuxified:
Adds Collatz Conjecture exercise as [described in x-common repo](https://github.com/exercism/x-common/blob/master/exercises/collatz-conjecture/description.md)

It wasn't clear how to use the [provided JSON as test data](https://github.com/exercism/x-common/blob/master/exercises/collatz-conjecture/canonical-data.json), so I ported them to ExUnit manually. Since this is my first (exercism) PR, let me know if I missed anything :)

### DIFF
--- a/exercises/collatz-conjecture/collatz_conjecture.exs
+++ b/exercises/collatz-conjecture/collatz_conjecture.exs
@@ -1,0 +1,13 @@
+defmodule CollatzConjecture do
+
+  @doc """
+  calc/1 takes an integer and returns the number of steps required to get the
+  number to 1 when following the rules:
+    - if number is odd, multiply with 3 and add 1
+    - if number is even, divide by 2
+  """
+  @spec calc(number :: integer) :: pos_integer
+  def calc(input) do
+    raise "ain't got no time for that"
+  end
+end
--- a/exercises/collatz-conjecture/collatz_conjecture.exs
+++ b/exercises/collatz-conjecture/collatz_conjecture.exs
@@ -6,8 +6,7 @@ defmodule CollatzConjecture do
     - if number is odd, multiply with 3 and add 1
     - if number is even, divide by 2
   """
-  @spec calc(number :: integer) :: pos_integer
+  @spec calc(number :: pos_integer) :: pos_integer
   def calc(input) do
-    raise "ain't got no time for that"
   end
 end
--- a/exercises/collatz-conjecture/collatz_conjecture_test.exs
+++ b/exercises/collatz-conjecture/collatz_conjecture_test.exs
@@ -13,7 +13,7 @@ defmodule CollatzConjectureTest do
   end
 
   test "zero is an error" do
-    assert_raise ArgumentError, "Only positive numbers are allowed", fn -> CollatzConjecture.calc(0) end
+    assert_raise FunctionClauseError, fn -> CollatzConjecture.calc(0) end
   end
 
   @tag :pending
@@ -43,6 +43,11 @@ defmodule CollatzConjectureTest do
 
   @tag :pending
   test "negative value is an error " do
-    assert_raise ArgumentError, "Only positive numbers are allowed", fn -> CollatzConjecture.calc(-15) end
+    assert_raise FunctionClauseError, fn -> CollatzConjecture.calc(-15) end
+  end
+
+  @tag :pending
+  test "string as input value is an error " do
+    assert_raise FunctionClauseError, fn -> CollatzConjecture.calc("fubar") end
   end
 end
--- a/exercises/collatz-conjecture/collatz_conjecture_test.exs
+++ b/exercises/collatz-conjecture/collatz_conjecture_test.exs
@@ -12,6 +12,7 @@ defmodule CollatzConjectureTest do
     assert CollatzConjecture.calc(1) == 0
   end
 
+  @tag :pending
   test "zero is an error" do
     assert_raise FunctionClauseError, fn -> CollatzConjecture.calc(0) end
   end
--- a/exercises/collatz-conjecture/collatz_conjecture_test.exs
+++ b/exercises/collatz-conjecture/collatz_conjecture_test.exs
@@ -1,0 +1,48 @@
+if !System.get_env("EXERCISM_TEST_EXAMPLES") do
+  Code.load_file("collatz_conjecture.exs", __DIR__)
+end
+
+ExUnit.start
+ExUnit.configure exclude: :pending, trace: true
+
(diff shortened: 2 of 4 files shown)


exercises/collatz-conjecture/collatz_conjecture_test.exs
```diff
@@ -0,0 +1,48 @@
...
+
+defmodule CollatzConjectureTest do
+  use ExUnit.Case
+
+  test "zero steps for one" do
+    assert CollatzConjecture.calc(1) == 0
+  end
+
+  test "zero is an error" do
+    assert_raise ArgumentError, "Only positive numbers are allowed", fn -> CollatzConjecture.calc(0) end
```
NobbZ: I think the error cases should be after the positiv cases. That way we can lead the student better through the process. 

Also I think an error tuple were more idiomatic.

exercises/collatz-conjecture/example.exs
```diff
@@ -0,0 +1,25 @@
...
+  @doc """
+  calc/1 takes number (> 1), and returns the number of steps required to get to 1 when
+  following the rules: if n is odd, multiply with 3 and add 1. if n is even, divide by 2
+  """
+  @spec calc(number :: integer) :: pos_integer
+  def calc(number) when number < 1 do
+    raise ArgumentError, "Only positive numbers are allowed"
+  end
+
+  def calc(number) do
```
NobbZ: Shouldn't we guard for integer here?

exercises/collatz-conjecture/collatz_conjecture_test.exs
Tuxified: I wasn't sure if I needed to follow the pattern as in x-common. I put the error case first as forgetting this might lead to an infinite loop later on.

exercises/collatz-conjecture/example.exs
Tuxified: Since the previous scenario returns an error if invalid data is supplied, that would (imo) be the best place to check for an integer

Tuxified:
So, adding it to track config means adding an entry to config.json, whoever, I don't know what to choose when it comes to difficulty & categories?

I think a difficulty lvl 2 and recursion should suffice?

exercises/collatz-conjecture/example.exs
NobbZ: `calc("foo")` will raise because you can't `div` or `rem` it inside the private function. I do not like it leaking an implementation detail in the stack trace, but this is a detail which should be decided upon by the track maintainers

NobbZ:
I'm not sure about the topic, I have decided to leave them empty for erlang in all exercises, but went with a difficulty of 2 over there as well.

exercises/collatz-conjecture/example.exs
Tuxified: If the guard for integer is added to the second function definition, then `calc("foo")` will raise a no-clause error. If we add the guard to the first definition, we'll get the error saying only positive numbers are allowed.

The first function would the look like:
```
def calc(number) when number < 1 or not is_integer(number) do
  raise ArgumentError, "Only positive numbers are allowed"
end
```

exercises/collatz-conjecture/collatz_conjecture_test.exs
devonestes: Yeah, we don't _have_ to follow the pattern exactly. Different languages have different ways of expressing error cases. This is honestly one of the hardest parts of this track, since there are totally valid arguments for each option (raising an exception vs. returning an error tuple).

I've been using this blog post as a guideline, which might be helpful to consult here: http://michal.muskala.eu/2017/02/10/error-handling-in-elixir-libraries.html#errors-vs-exceptions

devonestes:
@Tuxified Yeah, I think a `2` is a good difficulty, and probably just `recursion` as well. I would consider putting `guard clauses` in there, but you can technically solve this without using them, so maybe we can leave that out.

exercises/collatz-conjecture/collatz_conjecture.exs
```diff
@@ -0,0 +1,13 @@
...
+
+  @doc """
+  calc/1 takes an integer and returns the number of steps required to get the
+  number to 1 when following the rules:
+    - if number is odd, multiply with 3 and add 1
+    - if number is even, divide by 2
+  """
+  @spec calc(number :: integer) :: pos_integer
+  def calc(input) do
+    raise "ain't got no time for that"
```
devonestes: In general we don't start with an error when we give the function definition. I think you can just cut this line.

exercises/collatz-conjecture/example.exs
devonestes: Personally, if I were using a library and it provided a type signature for a function and I don't conform to that documentation, then I should expect a MatchError. That seems like reasonable behavior to me. Then we could also be more clear in the typespec for `calc/1` by saying `@spec calc(number :: pos_integer) :: pos_integer`, which I think would be a good thing.

I actually think it would be a _really_ great thing to show folks how to explicitly allow for a MatchError by changing the test to say `assert_raise MatchError, fn -> CollatzConjecture.calc(-15) end`. I don't think we have any other exercises about that yet, and it would be a good way to help folks explore pattern matching errors and typespecs.

exercises/collatz-conjecture/collatz_conjecture.exs
Tuxified: Agreed, done :)

Tuxified:
I've added an entry to config.json (difficulty 2, recursion). Instead of a ArgumentError, it now throws a FunctionClause Error as it wasn't clear to me how you pictured forcing a MatchError.

devonestes:
@Tuxified Yeah, sorry, I got my errors mixed up. This is totally what I was thinking of! Great job, and thanks! 👍

exercises/collatz-conjecture/collatz_conjecture_test.exs
```diff
@@ -0,0 +1,53 @@
...
+ExUnit.configure exclude: :pending, trace: true
+
+defmodule CollatzConjectureTest do
+  use ExUnit.Case
+
+  test "zero steps for one" do
+    assert CollatzConjecture.calc(1) == 0
+  end
+
+  test "zero is an error" do
```
devonestes: I think we're missing a `@tag :pending` here.

Tuxified:
Alright, this one should be good to merge
Thanks for all the remarks and improvements :)

devonestes:
Thanks for your work on this! 👍

PR #349, opened 2017-06-29
Implement dominoes exercise

Tuxified:
# Implements the [dominoes exercise](https://github.com/exercism/problem-specifications/tree/master/exercises/dominoes):

> Make a chain of dominoes.
> 
> Compute a way to order a given set of dominoes in such a way that they form a correct domino chain (the dots on one half of a stone match the dots on the neighbouring half of an adjacent stone) and that dots on the halfs of the stones which don't have a neighbour (the first and last stone) match each other.
> 
> For example given the stones 21, 23 and 13 you should compute something like 12 23 31 or 32 21 13 or 13 32 21 etc, where the first and last numbers are the same.
> 
> For stones 12, 41 and 23 the resulting chain is not valid: 41 12 23's first and last numbers are not the same. 4 != 3

## Solution/example

Creates all permutations of the given dominoes and then tries to find sets that match the specifications by simply checking if next stone can be added to last one. This example falls in the 'brute-force' category, there are definitely better solutions.

## Tests

Basically copies of the ones in problem description with some extra (that were missing imo):

- test if first and last number also match
- in all other tests the first stone can stay 'first' to make a match, added an exception ("scrambled" order)

## Topics & difficulty 

I've given it difficulty 4, since it's harder than most "starter" assignments. My example involved recursion, enumerations and sorting but someone clever might use trees and an actual algorithm to solve it.

Let me know what your thoughts are :)

### DIFF
--- a/exercises/dominoes/dominoes.exs
+++ b/exercises/dominoes/dominoes.exs
@@ -3,10 +3,8 @@ defmodule Dominoes do
   @type domino :: {1..6, 1..6}
 
   @doc """
-  chain?/1 takes a list of integer and returns boolean indicating if chain is possible
-  number to 1 when following the rules:
-    - if number is odd, multiply with 3 and add 1
-    - if number is even, divide by 2
+  chain?/1 takes a list of domino stones and returns boolean indicating if it's
+  possible to make a full chain
   """
   @spec chain?(dominoes :: [domino] | []) :: boolean
   def chain?(dominoes) do
--- a/exercises/dominoes/dominoes.exs
+++ b/exercises/dominoes/dominoes.exs
@@ -1,12 +1,14 @@
 defmodule Dominoes do
 
+  @type domino :: {1..6, 1..6}
+
   @doc """
   chain?/1 takes a list of integer and returns boolean indicating if chain is possible
   number to 1 when following the rules:
     - if number is odd, multiply with 3 and add 1
     - if number is even, divide by 2
   """
-  @spec chain?(dominoes :: list) :: boolean
-  def chain?(input) do
+  @spec chain?(dominoes :: [domino] | []) :: boolean
+  def chain?(dominoes) do
   end
 end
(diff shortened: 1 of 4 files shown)


exercises/dominoes/dominoes.exs
```diff
@@ -0,0 +1,12 @@
+defmodule Dominoes do
+
+  @doc """
+  chain?/1 takes a list of integer and returns boolean indicating if chain is possible
+  number to 1 when following the rules:
+    - if number is odd, multiply with 3 and add 1
+    - if number is even, divide by 2
+  """
+  @spec chain?(dominoes :: list) :: boolean
```
devonestes: Since we know the shape of this input pretty exactly, how about defining a type and using that in the typespec?

```elixir
@type domino :: [1..6, 1..6]

@spec chain?([domino] | []) :: boolean
```

Tuxified: `@type domino :: [1..6, 1..6]` is unfortunately not valid syntax (seems like one cannot specify the amount of members a list has. `@type domino :: {1..6, 1..6}` seems valid though. 
I'll have to change the code to use tuples instead of lists, is that OK?

devonestes: Hmmm, I just sort of assumed you could do that, but I guess you can only define that type as `[1..6]`, or we could use the tuple. Personally, I sort of like the tuple, but I'm cool with just `[1..6]` if you don't want to have to re-write this. It's not that big of a deal.

Tuxified: Switching to tuples wasn't that hard, thanks for your input :D

exercises/dominoes/dominoes.exs
```diff
@@ -0,0 +1,14 @@
+defmodule Dominoes do
+
+  @type domino :: {1..6, 1..6}
+
+  @doc """
+  chain?/1 takes a list of integer and returns boolean indicating if chain is possible
+  number to 1 when following the rules:
```
devonestes: I think maybe this line and the two below it are from a previous exercise?

Tuxified: Good catch, thanks!

PR #362, opened 2017-08-22
Add default introductory code example

kytrinyx:
Instead of having logic for a fallback in the backend, we're choosing a
default for the snippet file.

For tracks that have core exercises, we pick the first core exercise.
For tracks without these, we pick the first exercise listed in the config.

Note that we're aiming for 10 lines and a max-width of 40 columns.
This solution has 13 lines and a max-width of 55, so we may want to adjust it.

See https://github.com/exercism/meta/issues/89

### DIFF
--- a/docs/SNIPPET.txt
+++ b/docs/SNIPPET.txt
@@ -1,0 +1,13 @@
+defmodule HelloWorld do
+
+  @doc """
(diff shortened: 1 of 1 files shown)


devonestes:
I'll merge this and then update it to fit the lines and column width goals that we want to hit.

PR #365, opened 2017-08-27
Add pending test case for order of reduction

johnsyweb:
#### Context

@aliaksandr-martsinovich commented on [one of my submissions](http://exercism.io/submissions/49ec774315704a1baf7e2d67151df4e0) that `reduce` is supposed to start with the head of a list.

My implementation started at the tail and all the tests passed.

#### Change

 - Add a test that filed for that submission but passes [when we start with the head](http://exercism.io/submissions/17b3ffeca3874c048618ee4a60f415e8)

### DIFF
--- a/exercises/list-ops/list_ops_test.exs
+++ b/exercises/list-ops/list_ops_test.exs
@@ -85,6 +85,11 @@ defmodule ListOpsTest do
   end
 
   @tag :pending
+  test "order of reduction is important" do
+    assert L.reduce(~w(C B A), ". Done!", &(&1 <> &2)) == "ABC. Done!"
+  end
+
+  @tag :pending
   test "reduce of huge list" do
     assert L.reduce(Enum.to_list(1..1_000_000), 0, &(&1+&2)) ==
       Enum.reduce(1..1_000_000, 0, &(&1+&2))


devonestes:
Thanks again for your PR!! 👏 

I'm not a huge fan of saying `reduce` is supposed to have a given behavior. `reduce` is really just a more human-friendly synonym for `fold`, and it can be implemented going in either direction. This is why you often see `foldr` and `foldl` - and in fact, we do have that in Elixir!

https://hexdocs.pm/elixir/List.html#foldl/3
https://hexdocs.pm/elixir/List.html#foldr/3

I think it's fair to say that while it is _possible_ for `reduce` to be implemented starting from the head of the list, and in some contexts some folks might think of it as preferable, it's not incorrect for it to be implemented in the other way. In fact, since it's much more performant to add to the head of a list in Elixir than to add to the tail of a list, in this case `reduce` implemented as `foldr` is going to be the more performant implementation.

So, I hope you don't mind, but to avoid a breaking change here to the existing exercises, I think we'll pass on this one.

aliaksandr-martsinovich:
@devonestes thanks a lot for the explanation! The only thing I didn't quite understand is why `reduce-as-foldr` is going to be more performant than `reduce-as-foldl`. I mean, this is the solution (both mine and @johnsyweb):
```
@type acc :: any
@spec reduce(list, acc, ((any, acc) -> acc)) :: acc
def reduce([], acc, _), do: acc
def reduce([h|t], acc, f), do: reduce(t, f.(h, acc), f)
```

... and I can't think of any list insertions here. Is it something that only comes with a "proper" reduce implementation?

devonestes:
@aliaksandr-martsinovich Sorry, please forgive me, for a second I forgot the function signature for the function passed as the third argument. I was thinking about more of a `map` style function, where you would do:

```elixir
def reduce([], acc, _), do: Enum.reverse(acc)
def reduce([head|tail], acc, fun) do
  reduce(tail, [fun.(head) | acc], fun)
end
```

instead of the accumulator being passed into the given function and returned.

johnsyweb:
I don't mind either way.

To me it makes sense for `reduce` to match the behaviour of that in `Enum`:

```elixir
Interactive Elixir (1.4.1) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> Enum.reduce(~w(C B A), ". Done!", &(&1 <> &2))
"ABC. Done!"
```

Not breaking existing submissions is makes a lot of sense, also :-)

Thanks for the review and the explanation 🙇

Issue #359, opened 2017-08-13
pig_latin_test.exs could be added cases for words beginning with multiple consonant letters


alarregoity:
The test suite gives two rules:
 -   describe "first letter and ay are moved to the end of words that start with consonants" do
 -   describe "some letter clusters are treated like a single consonant" do
These two rules make us think that we must follow the "consonant rule" only for the first letter, or for the special "consonant cluster" cases given.

In contrast, wikipedia says: "For words that begin with consonant sounds, all letters before the initial vowel are placed at the end of the word sequence. Then, "ay" (some people just add "a") is added,"
Following wikipedia, any combination of consonant letters at the beginning of the word (not only the first letter) should be moved to the end of the word.
For instance, "pleasure" shoud be translated into "easureplay", not into "leasurepay".

I think that adding some test cases for the rule "all letters before the initial vowel are placed at the end ", with words beginning with multiple consonant letters, would produce more generic solutions.

devonestes:
Thanks for the recommendation! This seems reasonable to me for sure. Would you care to try out a PR?

rawkode:
Looks like this was closed by #366

devonestes:
Yep, thanks for the reminder!

PR #366, opened 2017-08-27
Pig Lating exercise: Add tests for the case of words beginning with multiple consonants

alarregoity:
To match the rule given in wikipedia: ""For words that begin with consonant sounds, all letters before the initial vowel are placed at the end of the word sequence."

Isssue #359

### DIFF
--- a/exercises/pig-latin/example.exs
+++ b/exercises/pig-latin/example.exs
@@ -17,28 +17,27 @@ defmodule PigLatin do
   def translate(phrase) do
     phrase
     |> String.split(" ")
-    |> Enum.map_join(" ", &to_pig_latin/1)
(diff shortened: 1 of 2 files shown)


devonestes:
@alarregoity Thanks so much for taking this on yourself!

If you look at the CI results, it looks like this will need an update to the `example.exs` file. That file is an example implementation that we need to ensure that we're not accidentally breaking things when we make updates (and also lets us know the tests are good). Think you could take a swing at updating the example implementation, too?

Thanks again! ❤️

alarregoity:
Thanks for the feedback.
I did not know how to modify the existing example.exs to pass the tests, so I have taken a completely different solution.

PR #367, opened 2017-09-04
Binary Search Tree Implementation

sngeth:
Hi, this is an implementation for the BST Exercise.

It looks like I may need to update the config.json update as well? Not sure exactly what goes in there though.

### DIFF
--- a/exercises/binary-search-tree/binary_search_tree.exs
+++ b/exercises/binary-search-tree/binary_search_tree.exs
@@ -1,0 +1,27 @@
+defmodule BinarySearchTree do
+  @type bst_node :: %{data: any, left: bst_node | nil, right: bst_node | nil}
+
+  @doc """
+  Create a new Binary Search Tree with root's value as the given 'data'
+  """
+  @spec new(any) :: bst_node
+  def new(data) do
+    # Your implementation here
+  end
+
+  @doc """
+  Creates and inserts a node with its value as 'data' into the tree.
+  """
+  @spec insert(bst_node, any) :: bst_node
+  def insert(tree, data) do
+    # Your implementation here
+  end
+
+  @doc """
+  Traverses the Binary Search Tree in order and returns a list of each node's data.
+  """
+  @spec in_order(bst_node) :: [any]
+  def in_order(tree) do
+    # Your implementation here
+  end
+end
--- a/exercises/binary-search-tree/binary_search_tree_test.exs
+++ b/exercises/binary-search-tree/binary_search_tree_test.exs
@@ -1,0 +1,103 @@
+if !System.get_env("EXERCISM_TEST_EXAMPLES") do
+  Code.load_file("binary_search_tree.exs", __DIR__)
+end
+
+ExUnit.start
+ExUnit.configure exclude: :pending, trace: true
+
+defmodule BinarySearchTreeTest do
+  use ExUnit.Case
+
+  test "retains data" do
+    assert BinarySearchTree.new(4).data == 4
+  end
+
+  @tag :pending
+  test "inserting lower number" do
+    root =
+      BinarySearchTree.new(4)
+      |> BinarySearchTree.insert(2)
+
+    assert root.data == 4
+    assert root.left.data == 2
+  end
+
+  @tag :pending
+  test "inserting same number" do
+    root =
+      BinarySearchTree.new(4)
+      |> BinarySearchTree.insert(4)
+
+    assert root.data == 4
+    assert root.left.data == 4
+  end
+
+  @tag :pending
+  test "inserting higher number" do
+    root =
+      BinarySearchTree.new(4)
+      |> BinarySearchTree.insert(5)
+
+    assert root.data == 4
+    assert root.right.data == 5
+  end
+
+  @tag :pending
+  test "complex tree" do
+    root =
+      BinarySearchTree.new(4)
+      |> BinarySearchTree.insert(2)
+      |> BinarySearchTree.insert(6)
+      |> BinarySearchTree.insert(1)
+      |> BinarySearchTree.insert(3)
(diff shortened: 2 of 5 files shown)


exercises/binary-search-tree/binary_search_tree_test.exs
```diff
@@ -0,0 +1,107 @@
...
+defmodule BinarySearchTreeTest do
+  use ExUnit.Case
+
+  test "retains data" do
+    assert BinarySearchTree.new(4).data == 4
+  end
+
+  @tag :pending
+  test "inserting lower number" do
+    four =
```
devonestes: Maybe instead of calling this `four` we can call it `root` or `root_node` or something like that? I think it might get a little confusing for some folks seeing `four.data == 4`.

exercises/binary-search-tree/binary_search_tree_test.exs
```diff
@@ -0,0 +1,107 @@
...
+    assert four.data == 4
+    assert four.left.data == 2
+    assert four.left.left.data == 1
+    assert four.left.right.data == 3
+    assert four.right.data == 6
+    assert four.right.left.data == 5
+    assert four.right.right.data == 7
+  end
+
+  def record_all_data(tree) do
```
devonestes: This function seems like it's not super helpful. We're essentially just renaming one of the functions provided by the module under test. Would you mind replacing `record_all_data/1` with `BinarySearchTree.in_order/1`?

exercises/binary-search-tree/binary_search_tree_test.exs
sngeth: I agree and updated to 'root'. I kinda cargo culted the naming pattern here from other impls, but we can do better for the elixir repo :-)

exercises/binary-search-tree/binary_search_tree_test.exs
sngeth: Agree, updated to call ```in_order/1``` dirctly

exercises/binary-search-tree/binary_search_tree.exs
```diff
@@ -0,0 +1,27 @@
+defmodule BinarySearchTree do
+  @type bst_node :: %{data: any, left: bst_node | nil, right: bst_node | nil}
+
+  @doc """
+  Create a new Binary Search Tree with root's value as the given 'data'
+  """
+  @spec new(any) :: %{data: any, left: bst_node, right: bst_node}
```
devonestes: I think we need to update this typespec. In your implementation it returns a `bst_node` (since it's valid for the left and right nodes to be `nil` in that type definition) but this typespec implies that we have at least a balanced tree with a depth of 1.

sngeth: Good catch, I think I fixed this. Looks like I don't take advantage of my custom type like I did for insert. It can be simply defined as `@spec new(any):: bst_node` which for left and right is a union of another bst_node or nil.

In the case of a newly created node left and right are both nil.

In the case of an insert under an existing node, one of left or right will be non-nil(another bst_node)

exercises/binary-search-tree/binary_search_tree.exs
```diff
@@ -0,0 +1,27 @@
...
+  @spec new(any) :: bst_node
+  def new(data) do
+    # Your implementation here
+  end
+
+  @doc """
+  Creates and inserts a node with its value as 'data' into the tree.
+  """
+  @spec insert(bst_node, any) :: bst_node
+  def insert(%{data: data, left: left, right: right}, data) do
```
devonestes: Sorry, just saw this, too - we have two variables named `data` here. Maybe we don't need such specific pattern matching in the first argument? Then we can use `def insert(tree, data) do`?

sngeth: This is probably a case where I may not be writing the most idiomatic code.

If we pattern match directly in the function definition we get auto destructuring to use inside the method (we need to destructure it so we can return a map that is the existing node's data with new children). The 2nd argument of data does make the signature read a bit confusing, something like new_node_value might be better)

If we refactor to your suggestion the example impl looks something like
  def insert(tree, node_value) do
    %{data: data, left: left, right: right} = tree # not neccessary with old signature

    if node_value <= data do
      %{data: data, left: insert(left, node_value), right: right}
    else
      %{data: data, left: left, right: insert(right, node_value)}
    end
  end

_or_ instead of adding a line to destructure the three, the values could be gotten directly as tree.data, tree.left, tree.right.

I'm leaning towards your refactor since it makes the function def look cleaner and less prescriptive of the implementation. Just thought i'd let you know what was going on in my mind when I wrote it. Let me know what you think and I'll make the change.

devonestes: My thinking behind not pattern matching in the function definition here was to let the person solving this problem arrive at that step on their own. I think it _is_ a really good thing to do, but since this is the template folks will be building off of, then I like to keep it as minimal as possible.

Just so you know, the implementation (`example.exs`) and the template (in this case `binary_tree.exs`) don't need to be totally the same. They should have the same public functions that satisfy the tests, but if you want to go nuts doing other things in your implementation, you're more than welcome to! So, if you want to do that pattern matching on the first argument of `insert/2` in your implementation, you're more than welcome!

sngeth: Gotcha that makes sense. Updated to new signature then and went ahead and squashed the commit.

sngeth: Gotcha makes sense. Made the update.

sngeth:
I had a little snafu with my squashing.. had to do soem force pushes

sngeth:
@devonestes cool thanks for your time to review this, definitely had fun and learned some things.

PR #371, opened 2017-10-09
Format according to 1.6 rules

devonestes:
Since the new formatter is out, I'm taking it for a spin by properly
formatting our exercises according to the new community standard. Since
1.6 hasn't been released (this was done on a build from `master`) we
can't easily hook things into CI yet, but hopefully this will get us
most of the way there. We can add a check for properly formatted code
to CI a little later.


devonestes:
Closing this since we're going to wait until 1.6 is officially released to add this in.

PR #372, opened 2017-10-10
WIP - Implement exercise transpose

chriszimmerman:


chriszimmerman:
I think I'm ready for someone to take a look at my PR. These Travis-CI builds seemed to have failed for reasons unrelated to my code changes, though. Any insights?

https://travis-ci.org/exercism/elixir/builds/287364144?utm_source=github_status&utm_medium=notification

exercises/transpose/transpose.exs
```diff
@@ -0,0 +1,4 @@
+defmodule Transpose do
+  def transpose(input) do
```
devonestes: Can you add a typespec here? Those are often really helpful for folks as they implement their solutions. At least they know what they are getting, and what they need to return.

Also, a little documentation is always helpful, even if it's duplicative of the information in the README. Many times folks don't even look at that and jump right into the exercise, or if they're viewing the submission on the website, having the documentation as a reminder right there is really helpful.

chriszimmerman:
Thanks for the constructive feedback! :)

I believe I've addressed all of your feedback, but please let me know if there's anything else I can improve.

config.json
```diff
@@ -889,6 +889,17 @@
         "Algorithms",
         "Sorting"
       ]
+    },
+    {
+      "uuid": "ce270a34-add1-422c-bb86-53b310f05e27",
+      "slug": "transpose",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 6,
```
devonestes: My gut tells me this isn't really a 6 in terms of difficulty. This number is always so hard to judge, but because of we only have one function and a lack of tricky edge cases, I have a feeling this could be considered an easier problem. Maybe a 5? I just look at some exercises that are 7's and I see stuff like `poker` and `minesweeper`, and those are all significantly more complicated.

exercises/transpose/README.md
```diff
@@ -0,0 +1,98 @@
...
+In general, all characters from the input should also be present in the transposed output.
+That means that if a column in the input text contains only spaces on its bottom-most row(s),
+the corresponding output row should contain the spaces in its right-most column(s).
+
+## Running tests
+
+Execute the tests with:
+
+```bash
+$ elixir bob_test.exs
```
devonestes: Would you mind changing this to `transpose_test.exs` and then deleting line 69?

exercises/transpose/README.md
```diff
@@ -0,0 +1,98 @@
...
+In the test suites, all but the first test have been skipped.
+
+Once you get a test passing, you can unskip the next one by
+commenting out the relevant `@tag :pending` with a `#` symbol.
+
+For example:
+
+```elixir
+# @tag :pending
+test "shouting" do
```
devonestes: Same thing here - would you mind updating this README more specifically to reflect this exercise?

config.json
chriszimmerman: I took a look at a few exercises with difficulties near the 5 level (`dominoes`, `clock`, and `difference of squares`) and 5 seems fine to me.

I agree that putting an objective measure like a number on the difficulty can be hard, especially if you're already familiar with how to solve certain sorts of problems, but not others. That can change the difficulty depending on the individual.

PR #374, opened 2017-10-12
Update travis matrix

devonestes:
We hadn't updated this in a bit. I think supporting the two most recent
versions of Elixir and OTP is reasonable.

### DIFF
--- a/bin/dialyzer_check.sh
+++ b/bin/dialyzer_check.sh
@@ -10,19 +10,8 @@ do
 done
 
 elixirc -o ./tmp/build ./tmp/src/*.exs
-
-if [ $CI ]; then
-    export PLT_FILENAME=elixir-${TRAVIS_ELIXIR_VERSION:=1.3.2}_${TRAVIS_OTP_RELEASE:=19.0}.plt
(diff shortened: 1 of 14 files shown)


bin/dialyzer_check.sh
```diff
@@ -10,19 +10,8 @@ do
...
-    wget -O $PLT_LOCATION https://github.com/danielberkompas/travis_elixir_plts/blob/master/elixir-${TRAVIS_ELIXIR_VERSION}_${TRAVIS_OTP_RELEASE}.plt?raw=true
-
-    dialyzer --no_check_plt --plt $PLT_LOCATION --no_native ./tmp/build;
-    RESULT=$?
-else
-    dialyzer ./tmp/build;
-    RESULT=$?
-fi
-
+mix dialyzer
```
devonestes: The old way we were getting the PLTs was funky, so we had to drop that and just do the slow thing and build the PLTs for each CI run. 😢

devonestes:
I'm going to go ahead and merge this and the other open PR so we can get some of that basic infrastructure stuff up to date.

PR #376, opened 2017-10-14
Changes to Luhn exercise

stfnsr:
I updated the tests for Luhn (#285),
it was also neccessary to rewrite the implementation in example.exs.
If this is unidiomatic or bad in other ways, i welcome feedback :)

### DIFF
--- a/exercises/luhn/example.exs
+++ b/exercises/luhn/example.exs
@@ -20,7 +20,7 @@ defmodule Luhn do
   @doc """
   Applies the luhn formula to a string only containing integers 
   """
-  def checksum(number) do
+  defp checksum(number) do
     0 == number
     |> String.graphemes()
     |> Enum.map(&String.to_integer/1)
@@ -33,7 +33,7 @@ defmodule Luhn do
   Reverses list and doubles every second digit,
   subtracts 9 if doubled digit is greater than 9
   """  
-  def double_even(numlist) do
+  defp double_even(numlist) do
     numlist
     |> Enum.reverse()
     |> Enum.zip(Stream.cycle([1,2]))
--- a/exercises/luhn/example.exs
+++ b/exercises/luhn/example.exs
@@ -7,7 +7,7 @@ defmodule Luhn do
   def valid?(number) do
     number_without_spaces = String.replace(number, " ", "")
     case Integer.parse(number_without_spaces) do
-      {_, ""} -> 
+      {_, ""} ->
         String.length(number_without_spaces) > 1 && checksum(number_without_spaces)
       _ ->
         false
@@ -36,6 +36,4 @@ defmodule Luhn do
 
   end
 
-  
 end
-  
--- a/exercises/luhn/example.exs
+++ b/exercises/luhn/example.exs
@@ -1,59 +1,51 @@
 defmodule Luhn do
-  @doc """
-  Calculates the total checksum of a number
-  """
-  @spec checksum(String.t()) :: integer
-  def checksum(number) do
-    number
-    |> String.reverse
-    |> String.graphemes
-    |> Enum.with_index
-    |> Enum.map(&do_checksum/1)
-    |> Enum.sum
-  end
-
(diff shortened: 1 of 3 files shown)


exercises/luhn/example.exs
```diff
@@ -1,59 +1,51 @@
...
+      true
+    else
+      _ -> false
+    end
   end
 
   @doc """
-  Creates a valid number by adding the correct
-  checksum digit to the end of the number
+  Applies the luhn formula to a string only containing integers
```
devonestes: The reason CI is failing right now is because all `@doc` tags on private functions are discarded. If you still feel like you need to document this, you can do so with normal comments, but given that this is just a sort of "proof of concept" for the exercise showing that it's possible to solve all the tests in the suite, I think you can safely delete them.

exercises/luhn/example.exs
```diff
@@ -1,59 +1,51 @@
...
+    |> Enum.sum()
     |> rem(10)
-    |> (&(10 - &1)).()
-    |> to_string
-    |> String.last
   end
 
-  defp double_if_even_index(n, index) do
-    if rem(index, 2) == 0, do: n, else: n * 2
+  @doc """
```
devonestes: Same here - this `@doc` tag will emit a warning at compile time, which fails CI for us. We try and stay warning free 😄

exercises/luhn/luhn_test.exs
```diff
@@ -8,37 +8,67 @@ ExUnit.configure exclude: :pending, trace: true
 defmodule LuhnTest do
   use ExUnit.Case
 
-  test "checksum" do
-    assert Luhn.checksum("4913") == 22
+  test "single digit strings can not be valid" do
+    assert Luhn.valid?("1") == false
```
devonestes: While we're in here, we don't actually need the `== false` in any of these assertions. Since `Luhn.valid/1` already returns a boolean, we can use `assert` and `refute` without comparing that boolean to a second boolean.

stfnsr:
Thanks for the review!
I fixed the things you mentioned, hope Travis will be happier now ^^

Also, as I haven't been using the 'with' form a lot until now, is this type of usage okay?
Asking because most examples I saw match on '{:ok, something}'

stfnsr:
Travis says:
> The command "wget https://repo.hex.pm/builds/elixir/v1.4.0-rc.1-otp-19.zip" failed and exited with 8 during .

I also can't download this file manually, works without the "rc.1-" though.

devonestes:
Yeah, that `rc` thing is an allowable failure, though. If you take a look here: https://travis-ci.org/exercism/elixir/jobs/288612628#L512 that is telling us there is some trailing whitespace somewhere in the new code. Can you take a quick peek and make sure that stuff is all cleaned up? Once it is, I bet Travis will be happy again and we can merge this puppy! 🐶

exercises/luhn/example.exs
```diff
@@ -1,59 +1,44 @@
...
-  end
 
-  @doc """
-  Creates a valid number by adding the correct
-  checksum digit to the end of the number
-  """
-  @spec create(String.t()) :: String.t()
-  def create(number) do
-    number <> do_create(number)
+    with  number       <- String.replace(number, " ", ""),
```
devonestes: So while this is technically correct, it's probably not the best usage of `with`. It's sort of a combination of the kind of railroad programming that it's designed for and some boolean logic. Plus the variable re-assignment is slightly confusing to me. On line 10, 11 and 12 you might not really know which `number` you're talking about. It's not really a blocker for deploying this, but if it were up to me, I'd rewrite this along the lines of the following:

```elixir
def valid?(number) do
  without_spaces = String.replace(number, " ", "")
  case Integer.parse(without_spaces) do
    {_, ""} ->
      String.length(without_spaces) > 1 && checksum(without_spaces)
    _ -> false
end
```

devonestes:
Thanks!! 🎉

stfnsr:
Thx - looks way better like this :tada:

PR #381, opened 2017-11-16
elixir protein translation exercise - new test case

nwshane:
Added test case for when invalid codon comes later in RNA string

### DIFF
--- a/exercises/protein-translation/protein_translation_test.exs
+++ b/exercises/protein-translation/protein_translation_test.exs
@@ -79,6 +79,11 @@ defmodule ProteinTranslationTest do
   test "invalid RNA" do
     assert ProteinTranslation.of_rna("CARROT") == { :error, "invalid RNA" }
   end
+  
+  @tag :pending
+  test "invalid codon at end of RNA" do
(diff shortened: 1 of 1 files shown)


devonestes:
@nwshane Oh man, I'm so sorry to have left this hanging out for so long!

Can you explain why you think this new case is necessary? Like, how would this guide a user in a way that the previous example doesn't?

nwshane:
@devonestes No problem! It was a while back, but as far as I can remember, this was my experience:

1. Solved the problem so that all tests passed
2. Realized (and then verified) that there was an untested kind of input that broke my program—when the invalid codon comes after the first codon in the RNA string
3. Submitted this pull request :)

Without this new test case, users can submit code that passes all tests, but which is broken for RNA strings that have invalid codons later in the string (like my old code).

Does that make sense?

devonestes:
Yeah, cool. If you've verified it, then that's a good addition to the test suite. CI is just failing at the moment because of some trailing whitespace. Once you clean that up and everything's passing on CI then we can give this a merge!

nwshane:
@devonestes I think it's fixed now? Let me know if I need to do anything else!

PR #389, opened 2018-02-09
improve "convert first base is one" test

Cohen-Carlisle:
the previous version was not properly testing that first base was validated
passing the "convert empty list" test would also pass this test


devonestes:
👍  Thanks!

PR #390, opened 2018-02-10
Pig Latin: Added tests and improved test descriptions

moxley:
Some test descriptions weren't doing a good job of communicating the rules of Pig Latin, leading to implementations that gamed the tests. For example, implementations were looking specifically for a leading 'xr' word as a vowel word (such as "xray"), but not any other combination of 'x' and a consonant (like "xbox").

Some descriptions are updated to communicate better what the implementation should provide.

A couple tests are added to break these overly-specific implementations.

The following implementation passes the new version of the tests:

http://exercism.io/submissions/2ec927894ece47baa8fe482f9f210a63

### DIFF
--- a/exercises/pig-latin/pig_latin_test.exs
+++ b/exercises/pig-latin/pig_latin_test.exs
@@ -63,7 +63,7 @@ defmodule PigLatinTest do
 
     @tag :pending
     test "word beginning with q without a following u" do
-      assert PigLatin.translate("qatar") == "atarqay"
+      assert PigLatin.translate("qat") == "atqay"
     end
 
     @tag :pending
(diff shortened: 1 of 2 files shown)


exercises/pig-latin/pig_latin_test.exs
```diff
@@ -63,7 +63,7 @@ defmodule PigLatinTest do
 
     @tag :pending
     test "word beginning with q without a following u" do
-      assert PigLatin.translate("qat") == "atqay"
+      assert PigLatin.translate("qatar") == "atarqay"
```
kbaird: "qat" is a real word (a type of bush, I think). But this also works.

moxley: Nice :+1:. You can tell I'm not much of a Scrabble player.

kbaird: Sadly, I suq at Scrabble.

moxley:
@devonestes: I adjusted the example.exs, and CI now passes.

PR #393, opened 2018-03-02
Fix largest-series-product/example.exs

parkerl:


parkerl:
I will check on that failure

parkerl:
Ah, I see, we are in a trap now because the formatter wants to make this change but 1.4 dose not support it. I was playing with adding the formatter to Travis. Is it time to drop 1.4 as a build target?

devonestes:
@parkerl I'd be cool with supporting just 1.6 and 1.5, dropping 1.4. I'm generally pretty aggressive on supporting the newest versions though. I think it makes sense especially for exercism, though, since many people starting are most likely going to be new to the language and not really working on an installed version a year old.

PR #394, opened 2018-03-10
New Exercise: Alphametics

britto:
From the official TODO list of unimplemented exercises.

I tried to optimize the reference implementation for brevity rather than performance, so the larger test case example takes about 20s to run on my machine. I can make some further optimizations if necessary, though.

### DIFF
--- a/exercises/alphametics/README.md
+++ b/exercises/alphametics/README.md
@@ -1,0 +1,73 @@
+# Alphametics
+
+Write a function to solve alphametics puzzles.
+
+[Alphametics](https://en.wikipedia.org/wiki/Alphametics) is a puzzle where
+letters in words are replaced with numbers.
+
+For example `SEND + MORE = MONEY`:
+
+```text
+  S E N D
+  M O R E +
+-----------
+M O N E Y
+```
+
+Replacing these with valid numbers gives:
+
+```text
+  9 5 6 7
+  1 0 8 5 +
+-----------
+1 0 6 5 2
+```
+
+This is correct because every letter is replaced by a different number and the
+words, translated into numbers, then make a valid sum.
+
+Each letter must represent a different digit, and the leading digit of
+a multi-digit number must not be zero.
+
+Write a function to solve alphametics puzzles.
+
+## Running tests
+
+Execute the tests with:
+
+```bash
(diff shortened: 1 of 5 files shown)


exercises/alphametics/README.md
```diff
@@ -0,0 +1,73 @@
...
+```elixir
+# ExUnit.configure exclude: :pending, trace: true
+```
+
+For more detailed information about the Elixir track, please
+see the [help page](http://exercism.io/languages/elixir).
+
+## Source
+
+Wikipedia [https://en.wikipedia.org/wiki/Alphametics](https://en.wikipedia.org/wiki/Alphametics)
```
britto: Not really sure what to put down here at this section. 🤔

exercises/alphametics/alphametics_test.exs
```diff
@@ -0,0 +1,162 @@
...
+
+defmodule AlphameticsTest do
+  use ExUnit.Case
+
+  # @tag :pending
+  test "puzzle with three letters" do
+    puzzle = "I + BB == ILL"
+
+    solution = %{
+      ?I => 1,
```
britto: We could also have used strings instead of raw chars as keys here. No strong benefits other than being more enumerable, as in `?A..?Z`.

exercises/alphametics/README.md
devonestes: This is totally fine. Many other versions don't have anything here, so I don't think we need to cite a source. This is mostly for cases where someone has invented a particular kata (like the Gilded Rose) and we want to give some attribution.

exercises/alphametics/alphametics.exs
```diff
@@ -0,0 +1,21 @@
+defmodule Alphametics do
+  @type puzzle :: binary
+  @type solution :: %{required(?A..?Z) => 0..9}
```
devonestes: Thanks for the types! These are often really helpful for folks.

exercises/alphametics/alphametics_test.exs
devonestes: I like the character literals! It's always nice to use things like this for folks that might not have seen them before.

PR #396, opened 2018-03-10
Fix confusion between proteins and amino acids

jerith:
There are several places in the Protein Translation exercise where amino acids are incorrectly referred to as proteins. This is confusing, and makes the already-difficult problem of naming things even harder.

### DIFF
--- a/exercises/protein-translation/README.md
+++ b/exercises/protein-translation/README.md
@@ -28,7 +28,7 @@ Note the stop codon terminates the translation and the final methionine is not t
 
 Below are the codons and resulting Amino Acids needed for the exercise.
 
-Codon                 | Protein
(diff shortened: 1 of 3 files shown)


devonestes:
Hey, thanks for this @jerith!

Can you do me a favor, though, and open this issue up in this repo: https://github.com/exercism/problem-specifications/tree/master/exercises/protein-translation

That's where the common metadata about these exercises is kept, and I wouldn't want to veer too far from that when it comes to terminology like this. Especially since the exercise is called `protein-translation`, this change might confuse some folks (even though it is technically correct), so I would be curious to see what the other maintainers of the other tracks think before going ahead with this.

Thanks again!

Issue #397, opened 2018-04-13
[SUGGESTION] Create docker images for participants to use


SyedAman:
We want to improve the user experience for all Exercism users. We want to make it as easy as possible to get up and running solving exercises without the frustration. Windows users usually run into some sort of installation problems too.

One way we can achieve this is by setting up environments ourselves for them to use via Docker! So in the Elixir track, we can provide participants a Docker image to use. Sometimes users also don't want to install too many packages in their machine to avoid cluttering.

Users would just need to type in something like:

```bash
$ docker run -it -v $(pwd):/root/exercism/elixir exercism-elixir
```

And get started asap!

The Dockerfile can be defined as such in an official registry on DockerHub for Exercism:

```docker
FROM elixir

# Download and install Exercism CLI
RUN wget https://github.com/exercism/cli/releases/download/v2.4.1/exercism-linux-64bit.tgz
RUN tar -xzvf exercism-linux-64bit.tgz
RUN mkdir ~/bin
RUN mv exercism ~/bin/
RUN export PATH=$HOME/bin:$PATH
RUN echo 'export PATH=$HOME/bin:$PATH' >> ~/.bashrc

# Create working directory
WORKDIR /root/exercism/elixir
VOLUME /root/exercism/elixir

ENTRYPOINT ~/bin/exercism fetch elixir && bash
```

NobbZ:
1. I'd not install `exercism` into `${HOME}/bin` but into `/usr/local/bin`.
2. The resulting image is bloated due to the nature of overlayed filesystems. Do the following instead of the single command `RUN`s you have.
  ```dockerfile
  RUN set -ex; \
      oldwd=$(pwd)
      mkdir -p /tmp/exercism-install; \
      cd /tmp/exercism-install; \
      wget https://github.com/exercism/cli/releases/download/v2.4.1/exercism-linux-64bit.tgz; \
      tar -xzvf exercism-linux-64bit.tgz; \
      mv exercism /usr/local/bin/exercism; \
      cd $oldwd; \
      rm -rfv /tmp/exercism-install
  ```
3. Provide `ARG` fields to create exercism configuration on build.
4. AFAIK you can not use `&&` in an `ENTRYPOINT`
5. I'd create a proper user inside the container
6. I'd pin the elixir version you derive the image `FROM`
[...]

PR #398, opened 2018-04-14
Create Docker image for users to use via Dockerfile

SyedAman:
We want to improve the user experience for all Exercism users. We want to make it as easy as possible to get up and running solving exercises without the frustration. Windows users usually run into some sort of installation problems too.

One way we can achieve this is by setting up environments ourselves for them to use via Docker! So in the Elixir track, we can provide participants a Docker image to use. Sometimes users also don't want to install too many packages in their machine to avoid cluttering.

Users would just need to type in something like:

```bash
$ docker run -it -v $(pwd):/root/exercism/elixir exercism-elixir
```

And get started asap!

This is in response to #397.

With the help of @NobbZ, the Dockerfile can be defined as such in an official registry on DockerHub for Exercism:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM elixir:1.6.4-alpine
+
+# Download and install Exercism CLI
+RUN set -ex; \
+    oldwd=$(pwd); \
+    mkdir -p /tmp/exercism-install; \
+    cd /tmp/exercism-install; \
+    wget https://github.com/exercism/cli/releases/download/v2.4.1/exercism-linux-64bit.tgz; \
+    tar -xzvf exercism-linux-64bit.tgz; \
+    mv exercism /usr/local/bin/exercism; \
+    cd $oldwd; \
+    rm -rfv /tmp/exercism-install; \
+    exercism fetch elixir; \
+    export PATH=/usr/local/bin:$PATH; \
+    echo 'export PATH=/usr/local/bin:$PATH' >> ~/.bashrc
+
+# Create working directory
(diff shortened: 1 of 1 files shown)


NobbZ:
I want to add, that for inexperienced users installing docker can be a bigger hurdle than installing choco + elixir packages.

This might mainly be related to the fact, that there are 2 versions of docker available for windows.

1. Docker4Windows which uses windows internal emulation stuff to create a virtual linux box and requires HyperV
2. Docker Toolbox which uses Virtual Box for the underlying VM and does not allow HyperV to be active…

Both are totally different when it comes to how you handle volumes (shared folders).

Also both do massively differ in how they handle memory allocation. The HyperV version tries to pre-allocate the memory for the complete VM, which can cause running your Host OOM very quickly. Virtual Box is much more memory friendly as it allocates memory necessary "on the go".

So in general I do not see Docker as a solution to the "how to install" problem, it just moves it elsewhere.

It seems a bit as if docker is used as vagrant here.

And to be honest, docker is not vagrant, docker is not for dev, docker is for deployment and should be used as that.

If one wants to develop in or against a virtual machine, then one should use vagrant (but again, this does not solve the "how to install" problem, it just moves it).

In my (very unimportant) opinion, one needs to learn how to install the system one wants to work with on his own.

Each one can do this through mechanisms he already knows. Maybe docker, maybe vagrant, maybe choco, maybe a self built VM, maybe installing directly or even dualboot.

devonestes:
Thanks for putting this together, but I don't think we're going to go this route.

PR #407, opened 2018-06-14
Add blurb to config for Exercism 2.0

devonestes:
I decided the best thing to do would be the pull the first two
sentences from elixir-lang.org to keep things consistent. Plus, they
probably do a better job explaining the language that I would ;)


devonestes:
Anyone have any comments on this? If not, I'm going to merge in a couple days.

PR #408, opened 2018-06-14
Format the ABOUT.md

devonestes:
The formatting guidelines stipulated that each sentence should be on
its own line, with a line in between, so I've made that change here.


devonestes:
Anyone have any comments on this? If not, I'm going to merge in a couple days.

PR #409, opened 2018-06-14
Add core exercises

devonestes:
We discussed this a bit in #353, and now I'm opening the PR to actually
make this happen. I'd be really happy to hear other people's thoughts
on this progression and set of core exercises, though!!

PR #410, opened 2018-06-19
Add auto_approve to hello-world exercise

jackhughesweb:
This adds the auto_approve property to the hello-world 
exercise as part of preparing tracks for the v2 launch.

The current code on the server will automatically approve
hello-world exercises anyway, but that is a temporary fix
and will be removed in the future.

See exercism/elixir#406


devonestes:
This is already included in #409, so I'm going to close this in favor of that PR.

PR #417, opened 2018-08-04
Add 'math' topic to relevant exercises in track config

kytrinyx:
The new site lets people filter optional exercises by topic, and this
makes it easy for people to skip past them.

This is important since the math-y exercises are not key to learning
the language, and a lot of people find them intimidating and demotivating.

This updates any 'mathematics' topics to be 'math' since it's shorter
and won't get cropped in the UI.

It also removes the math topics for any exercises that are not in the
canonical math-y list.

exercism/exercism.io#4110

### DIFF
--- a/config.json
+++ b/config.json
@@ -10,8 +10,8 @@
       "slug": "hello-world",
       "uuid": "cc96d65d-1c79-40d0-8fd2-9a6665a43b01",
       "core": true,
-      "unlocked_by": null,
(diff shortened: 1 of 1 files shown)


kytrinyx:
I ran `configlet fmt` on this after adding the topics, which reformated a bunch of things. Let me know if you want me to submit a separate PR with just the formatting changes, and rebase this onto that.

Issue #414, opened 2018-07-24
Convert exercises to mix projects


amencarini:
After a chat with the active mentors on the Slack channel, a consensus emerged that the best way forward would be to have the exercises structured as mix projects.

The reasons given are:
1) It's how Elixir projects work in the real world, and students must be made aware that's the case when using exercism as a learning tool
2) `mix` already comes with the language so we aren't adding dependencies
3) It will execute any doctests you write
4) Dialyzer support can also be added to check typespecs
5) It lets you use `mix test.watch`
6) It lets you add dependencies easily (e.g, for debugging)
7) The current way of running tests makes using `IEx.pry` really complicated

As long as it doesn’t create incompatibility issues with how exercism works we'd like to start converting the exercises.

Could you please advise?

devonestes:
So this has been brought up at least 5 times in the last two years, which tells me that folks are clearly pretty interested in this. In the past it has always been rejected in favor of making these exercises as simple as possible for beginners to get going with - no need to understand a build system or anything. However, since folks keep asking, I'm open to exploring this again. A few notes on your reasons given

1) This has always been the strongest argument, and I'm happy with it
2) Yeah, it's good that we don't need any dependencies
3) This isn't actually true. You still need to include `doctest Module.Name` in the test file to run doctests.
4) Dialyzer is extremely complicated to set up, and extremely slow to run for the first time. I would be generally cautious about recommending dialyzer for these exercises.
5) This to me isn't a huge benefit. They're small exercises with only a few tests.
6) Can you tell me some of the dependencies you're thinking of here? I'm unaware of any dependencies that are purely for debugging.
7) I've always found `IEx.pry` kind of complicated since it can't be used in a test, only in an IEx session. Using Mix won't change that.

So, there are definite pros and cons. Our Elixir community doesn't have nearly as many "true" beginners (almost everyone I've ever met or heard from came from another language), so maybe the focus on keeping things easy for those people might not have much benefit. However, by making this change, we'd potentially be making it harder for any "true" beginner to get started and have a great first experience with the language.

I'll leave this open for a little longer to see what other folks chime in with.

Cohen-Carlisle:
I'd like to use mix as well. I think the extra files are worth it for the tools mix brings to the table, not to mention that it teaches people how to use elixir's built-in tooling.

NobbZ:
I'm also strictly in favor of using `mix`, as it gives much more flexibility to the student and makes him able to easily play around with things as he likes.

[...]

PR #420, opened 2018-08-19
"Mixify" hello-world and nucleotide-count

amencarini:
As discussed in https://github.com/exercism/elixir/issues/414, this shows how exercises refactored to use `mix` would look like.

This can be further simplified if necessary. `.formatter.exs`, `.gitignore`, and the `config` directory can be removed and everything would still work.


NobbZ:
You need to update the testrunner as well, also I'm not a fan of porting the exercises piece wise, it should happen all at once or not at all.

amencarini:
@NobbZ agreed that we should either port everything or nothing, this is to provide a sample as @devonestes asked in the issue.

re: the test runner, it needs no change, the problem with the build you saw was due to `hello_world_test.exs` having a `doctest` call that wouldn't work with the way tests are run across the exercises. That has been fixed.

devonestes:
Thanks for the PR!

I think if we're going for full compatibility with a standard mix project, we'll need to keep all the files that are normally generated when you do `mix new`. It's a lot of files, and many of them are unneeded, but that's part of the cost of going that route - bloat. Personally, after seeing the changes, I'm a little less enthusiastic about it.

The reason for that is that most of these exercises aren't really applications, but more like single functions. It would make sense to me if we could somehow get all of the exercises into a single mix application, but individual mix apps for each exercise feels a little weird to me. Am I the only one who's thinking that?

amencarini:
I think that despite the bloat, the "this is what real world apps look like" approach is worth the cost, and am still leaning towards 1 mix project per exercise. I know I might be biased because I am very familiar with that structure.

It would be interesting to take a couple of possible students, an experienced programmer that doesn't know Elixir and someone very new to programming, and ask them to complete a couple of exercises with the mix project structure (and amended documentation). Depending on the level of the struggle they encounter we could make a more informed decision on this.

neenjaw:
Out of curiosity, what is the status of this? Is the issue man-power and mixifying the course? or have the winds of opinion changed on this?

As @NobbZ mentioned in #478, `Code.load_file` has been soft deprecated and should be evaluated for migration to the newer implentations, so this could be an opportunity for that.

Cohen-Carlisle:
I think the issue is just getting the time to do this. I'm still in favor of it as it shows elixir as you'll encounter it in real-world projects.

neenjaw:
Ok, I have boiled it down to about 10 steps, working on a shell script that will just do it all programmatically and can probably submit a pull request with all of the exercise updates in a day or two.

neenjaw:
I think it's almost ready for a pull request, will submit tomorrow or the next day.  If you want to see a preview, check out my forked branch here: https://github.com/neenjaw/elixir/tree/mixify-exercises

neenjaw:
This PR has now been accomplished with PR #480.  Close this issue?

PR #425, opened 2018-08-30
difference-of-squares: renamed occurances of sums to sum

Scientifica96:
Renamed occurrences of sums to sum, like in `square_of_sums`.

### DIFF
--- a/exercises/difference-of-squares/difference_of_squares.exs
+++ b/exercises/difference-of-squares/difference_of_squares.exs
@@ -1,6 +1,6 @@
 defmodule Squares do
   @moduledoc """
-  Calculate sum of squares, square of sum, difference between two sum from 1 to a given end number.
+  Calculate sum of squares, square of sum, difference between two sums from 1 to a given end number.
   """
 
   @doc """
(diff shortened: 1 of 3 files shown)


exercises/difference-of-squares/difference_of_squares.exs
```diff
@@ -1,6 +1,6 @@
 defmodule Squares do
   @moduledoc """
-  Calculate sum of squares, square of sums, difference between two sums from 1 to a given end number.
+  Calculate sum of squares, square of sum, difference between two sum from 1 to a given end number.
```
devonestes: This should still be `difference between two sums from...`

exercises/difference-of-squares/example.exs
```diff
@@ -1,6 +1,6 @@
 defmodule Squares do
   @moduledoc """
-  Calculate sum of squares, square of sums, difference between two sums from 1 to a given end number.
+  Calculate sum of squares, square of sum, difference between two sum from 1 to a given end number.
```
devonestes: This should still be `difference between two sums from...`

exercises/difference-of-squares/difference_of_squares.exs
Scientifica96: Thank you for your remark. I'll change it now!

exercises/difference-of-squares/example.exs
Scientifica96: Thank you for your remark. I'll change it now.

PR #433, opened 2018-09-21
List ops clarify allowed functions

### LINKED ISSUE #430: Student confused by instructions
https://exercism.io/mentor/solutions/fedd5619ee694309a30539ab35976a7b

A student was confused by the instruction not to use Kernel. Should be clearer.

Cohen-Carlisle:
resolves #430

### DIFF
--- a/exercises/list-ops/example.exs
+++ b/exercises/list-ops/example.exs
@@ -1,10 +1,9 @@
 defmodule ListOps do
-  # Please don't use any external modules (especially List) in your
-  # implementation. The point of this exercise is to create these basic functions
-  # yourself.
-  #
-  # Note that `++` is a function from an external module (Kernel, which is
-  # automatically important`) and so shouldn't be used either.
+  # Please don't use any external modules (especially List or Enum) in your
+  # implementation. The point of this exercise is to create these basic
+  # functions yourself. You may use basic Kernel functions (like `Kernel.+/2`
+  # for adding numbers), but please do not use Kernel functions for Lists like
+  # `++`, `--`, `hd`, `tl`, `in`, and `length`.
 
   @spec count(list) :: non_neg_integer
   def count(l), do: do_count(l, 0)
--- a/exercises/alphametics/example.exs
+++ b/exercises/alphametics/example.exs
@@ -51,7 +51,7 @@ defmodule Alphametics do
 
   defp to_number(letters, numbers) do
     letters
-    |> Enum.map(& numbers[&1])
+    |> Enum.map(&numbers[&1])
     |> Integer.undigits()
   end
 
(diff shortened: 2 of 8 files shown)


Cohen-Carlisle:
Bump.

exercises/list-ops/example.exs
```diff
@@ -1,10 +1,9 @@
...
-  # implementation. The point of this exercise is to create these basic functions
-  # yourself.
-  #
-  # Note that `++` is a function from an external module (Kernel, which is
-  # automatically important`) and so shouldn't be used either.
+  # Please don't use any external modules (especially List or Enum) in your
+  # implementation. The point of this exercise is to create these basic
+  # functions yourself. You may use basic Kernel functions (like `Kernel.+/2`
+  # for adding numbers), but please do not use Kernel functions for Lists like
+  # `++`, `--`, `hd`, `tl`, `in`, and `length`.
```
devonestes: I think telling folks not to use `++` and `--` is good, but I sort of think we should allow `hd`, `tl`, `in` and `length`. Otherwise, we're really forcing them into a certain implementation (using solely pattern matching).

Maybe we can say there are two sorts of "levels" of solutions? Like, saying not to use `++` & `--` always, but if you want a challenge you can try to solve this without the other functions as well?

Cohen-Carlisle: I just did a quick scan for list related functions in `Kernel`. I think we must tell them to avoid `++` and `length` because they solve `ListOps.append` and `ListOps.count` respectively. I actually don't see how `--` or `in` would be useful. `hd` and `tl` are probably not useful either as they raise an `ArgumentError` for empty lists For getting the head and tail, you need to pattern match instead. I'd rather keep it simple and just forbid all List related functions, or alternatively, just `++` and `length` for the aforementioned reasons.

Cohen-Carlisle: @devonestes thoughts?

devonestes: @Cohen-Carlisle Sorry I took so long - I was very very busy following a conference, and then very very sick with a 103 fever for a week.

The more I think about it, the more I'm actually OK with allowing _all_ list related functions in `Kernel`, and then noting that if people want a challenge they can solve this with _none_ of the list related functions in `Kernel`. My thinking:

1) I agree that it's best to keep it simple instead of picking and choosing from `Kernel`. It's much easier to say "Please don't use any functions from `List` or `Enum`".

2) I personally see the real value of this exercise in implementing `map`, `filter` and `reduce`, since those are essentially the three elemental types of functions in all of FP. If we give them `append` and `count`, I'm not so worried about that. What they would learn in implementing those functions without `Kernel` functions is a duplication of the other three since `count` and `append` are both `reduce` functions.

Basically, I want to make this exercise as easy as possible without sacrificing implementation of those three functions, and we can do that by letting them use all the `Kernel` functions.

Cohen-Carlisle: This kinda dropped off my radar, but I think the most conservative/least change is as-is because it addresses that you are allowed to use, for example, `Kernel.+/2`. What do you think of merging this here and revisiting the two tiered (with and without forbidding certain `Kernel` functions) as a follow up?

PR #435, opened 2018-11-10
Check hit count for binary search to assert that search is binary

Virviil:
While mentoring the task, I found out that some student don't understand what's the difference between "search" and "binary search"

So I decided to modify the task and the test, in order to check "algorithmic complexity" of this search.

While this is a first approach - I hope that this can work.

I probably need some suggestions for the Api.

And also need help with task description update, in order to point out, that now it's not the **tuple with ints**, but **a tuple with closures, that returns ints**.

### DIFF
--- a/exercises/binary-search/binary_search_test.exs
+++ b/exercises/binary-search/binary_search_test.exs
@@ -8,44 +8,81 @@ ExUnit.configure(exclude: :pending, trace: true)
 defmodule BinarySearchTest do
   use ExUnit.Case
 
+  def box_tuple(tuple) do
+    tuple
+    |> Tuple.to_list()
+    |> Enum.reduce({}, fn (elem, acc) ->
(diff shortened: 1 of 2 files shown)


Virviil:
From my opinion, tests should cover the task - more or less. That is their purpose. Currently, test covers only **finding place** if an element, but not the *binarity* of the search.

So, to let current tests pass - I can pick any search algorithm. Of course, mentor will see the problem, but following this logic we can through away all the tests.

While this can be a bit confusing to use, I think that it's very useful for people to see tests that cover **not half of the task**, in the same time to understand, how and why binary search is working ( Aha, for this tuple we can find solution in 4 steps, because...)

NobbZ:
Even though the tests use tuples nowadays, I'm still strictly in favour of deprecating this exercise. Tuples should be seen as anonymous structs, not as a fixed width list.

PR #436, opened 2018-11-12
Added a small fix to bob.exs and bob_test.exs

chriseyre2000:
Added a comment to the main bob.exs source to remind the user how to submit a new solution.
Several students have failed to do this for the second solution.

Also added a test that detects functions that should be defp.
Adding this test should greatly reduce the number of times students need to be told this.

### DIFF
--- a/exercises/bob/bob.exs
+++ b/exercises/bob/bob.exs
@@ -1,3 +1,7 @@
+
+# You can submit a new solution using: 
+# exercism boc.exs
+
 defmodule Bob do
   def hey(input) do
     cond do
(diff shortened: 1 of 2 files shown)


chriseyre2000:
How did I break the build? I added a comment and a test!

devonestes:
@chriseyre2000 Thanks for taking the time to submit a PR!

CI is failing because we have a check to make sure we don't introduce trailing whitespace, which frequently creates noise in git diffs and history.

We don't prescribe specific implementations with our tests, so I don't want to force learners to make sure that all other functions in that module are private.

Also, if folks are having issues figuring out how to submit solutions, the proper place for that instruction is not in the exercises themselves but somewhere else in the onboarding processes.

Thanks again for taking the time to put together the PR, but I don't think this is something we want to merge.

PR #438, opened 2018-11-14
Remove invalid spec from queen-attack

angelikatyborska:
When I wanted to solve this exercise, I noticed that the provided empty solution file doesn't even compile:
```
** (CompileError) queen_attack.exs:8: spec for undefined function new/0
```

This PR fixes that problem, and additionally removes one typo.

### DIFF
--- a/exercises/queen-attack/queen_attack.exs
+++ b/exercises/queen-attack/queen_attack.exs
@@ -5,13 +5,12 @@ defmodule Queens do
   @doc """
   Creates a new set of Queens
   """
-  @spec new() :: Queens.t()
   @spec new({integer, integer}, {integer, integer}) :: Queens.t()
   def new(white, black) do
   end
 
   @doc """
-  Gives a string reprentation of the board with
+  Gives a string representation of the board with
   white and black queen locations shown
   """
   @spec to_string(Queens.t()) :: String.t()
--- a/exercises/queen-attack/queen_attack.exs
+++ b/exercises/queen-attack/queen_attack.exs
@@ -5,8 +5,9 @@ defmodule Queens do
   @doc """
   Creates a new set of Queens
   """
+  @spec new() :: Queens.t()
   @spec new({integer, integer}, {integer, integer}) :: Queens.t()
-  def new(white, black) do
+  def new(white \\ nil, black \\ nil) do
   end
 
   @doc """


angelikatyborska:
After 5 more minutes of thought, this actually looks like it was intended like that because `Queens.new` is supposed to accept either 0 or 2 arguments. Now I feel a bit stupid, but if I got confused, than maybe students will get confused about this too...

I'm not sure what is the standard thing to do with specs for functions that have default arguments. Should we always provide a spec for each possible combination:
```elixir
@spec new() :: Queens.t()
@spec new({integer, integer}) :: Queens.t()
@spec new({integer, integer}, {integer, integer}) :: Queens.t()
```
Or just the most specific one? I'm asking both generally for Elixir code and also specifically for this repository.

If this:
```elixir
@spec new() :: Queens.t()
@spec new({integer, integer}, {integer, integer}) :: Queens.t()
```
is indeed the desired version by the maintainers and we are ok with letting the compilation error be, then I would at least open a different PR with the typo 😅.

britto:
I am not sure if there is a documented standard/styleguide about this particular issue either, but I used to get confused about it quite often too. What I could observe from Elixir's own codebase is that when there are default arguments, the longest-arity version of the function is the one that gets a spec. Take [`Map.pop`](https://github.com/elixir-lang/elixir/blob/v1.7.3/lib/elixir/lib/map.ex#L616) and [`String.replace`](https://github.com/elixir-lang/elixir/blob/v1.7.3/lib/elixir/lib/string.ex#L1344) for example. So, I would say that it is safe to follow the same conventions here.

devonestes:
@angelikatyborska @britto 

Yeah, I can see how this is confusing because we give a spec for `new/0` and `new/2`, but we don't show that the arguments in `new/2` can/should have default arguments.

I think maybe instead of getting rid of that `new/0` spec maybe we can give another empty function head for `new/0` and then later on folks can figure out that they can use default arguments instead?

That seems kind of weird, too, though, since it assumes that folks know `new/2` with defaults actually compiles to `new/0` and would satisfy that spec...

I kind of want to keep that style of spec writing so folks get exposure to more than one way of writing specs - especially since that's the way you have to write specs for that function if those two arguments have defaults. Maybe we can give them `def new(white \\ nil, black \\ nil) do` so it at least compiles and folks figure out how that style of spec writing works?

angelikatyborska:
I think I would prefer the solution with invalid default values `def new(white \\ nil, black \\ nil) do` and keeping the spec for `new/0` because it compiles, and it teaches right away about default values. If we provide an empty function for both `new/0` and `new/2`, people might assume it's meant to be this way and will not try to use default values for `new/2`.

britto:
It is also worth noting that `def new(white \\ default, black \\ default)` will generate `new/1` too, which may not be too useful or even semantically desirable.

devonestes:
@angelikatyborska Ok, so it looks like giving them defaults is the way to go. Want to update your PR with that?

angelikatyborska:
Done.

Issue #426, opened 2018-08-31
Find better exercise than Robot Simulator for the "OTP" tag


umamaistempo:
The exercise _Robot Simulator_ has the OTP tag but doesn't specify that the solution must be done using processes.

Right now I am recommending the students not to use processes as this exercise doesn't require it for anything nor does it specify this need (and I used a similar exercise for code interviews and usually preferred when the candidate didn't use processes as usually when they did, it was solely to show that they know how to use an Agent and only made the code harder).

What I see people doing is starting an Agent for each robot and I don't find it adequate for several reasons:
- The robots are FSM-like machines, so what you want is a process to represent them and their actions (a GenServer), not just to hold _shared_ state*
- IF what you want is to hold state, since no one else is going to access it, you could keep that on your current scope by just assigning to a binding
- The robots don't interact in any way nor is it requested to have them all in any kind of collection, so there is no advantage in keeping them in process (actualy I think it's bad because this incentive the student to think that every different _object_ on the business model should be a different process, one of the most common conceptual mistakes from new erlang developers**)
- This makes code harder to read without any need

\* According to the own [Agent module manual](https://hexdocs.pm/elixir/Agent.html#content),
> Agents are a simple abstraction around state.
> Often in Elixir there is a need to share or store state that must be accessed from different processes or by the same process at different points in time.

\*\* TODO: find article about the missuse of processes to make separation of business entities

## TL;DR
We should probably remove the OTP tag from the Robot Simulator exercise because it is inducting students to follow bad practices

devonestes:
While this may not be the ideal use case for a GenServer, I think this is about as good a case as we're going to find available to us in Exercism. The platform is just not designed for larger problems, the likes of which would necessitate larger, concurrent solutions like those provided by GenServer.

What this is, though, is a perfect conceptual idea (and test design) for a GenServer. When implementing this, we specifically wanted it to be solvable with either some sort of in-process state (like a struct or map), or by spinning up a process, so folks have the flexibility to learn either (or both?) ways of solving this problem, and then seeing the pros and cons of both.

This is where you mentors come in! If you see someone who just shoves state in an Agent, you can show them how to properly create an API for a GenServer, and then how to use `call` and/or `cast` in that API to implement the GenServer behaviour.

[...]

PR #439, opened 2018-11-22
Removed OTP from Robot

chriseyre2000:
OTP is not useful for the Robot Simulator.

### DIFF
--- a/config.json
+++ b/config.json
@@ -681,7 +681,6 @@
       "unlocked_by": null,
       "difficulty": 6,
       "topics": [
-        "otp",
         "pattern_matching",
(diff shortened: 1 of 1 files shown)


devonestes:
Hi @chriseyre2000!

We already had a pretty long discussion about this both in Slack and in #426. Basically, once we find a better exercise to use to teach some OTP concepts, then we'll remove this tag, but we won't remove it until we find something better. So I'm going to close this PR for now, and if you can think of a better exercise to use to teach OTP, totally add that to the discussion in #426 and we'll see if we can make some improvements.

PR #440, opened 2018-11-24
Made the instructions more explicit.

chriseyre2000:
Added some more details to the instructions to make the problem more explicit.


Cohen-Carlisle:
Thanks for this, sorry that it sat for so long.

PR #450, opened 2019-02-20
fix doctest in exercise markdown

Sayrus:
Match doctest with given code

### DIFF
--- a/exercises/markdown/markdown.exs
+++ b/exercises/markdown/markdown.exs
@@ -7,8 +7,8 @@ defmodule Markdown do
     iex> Markdown.parse("This is a paragraph")
     "<p>This is a paragraph</p>"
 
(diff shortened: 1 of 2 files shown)


Sayrus:
Hello,

Thanks for the recommandation.
I removed the iex prompt and added a few useful tests that should help building a clean markdown parser.

> Testing text with left-flanked #:
> #This is not a header
> 
> Testing text with only left-flanking:
> __ This is not bold
> 
> Testing text with only right-flanking:
> Neither is this __
> 
> Testing text with enclosed with non-flanking:
> __ This is not bold __
> 
> And finally, testing if the parser handles lists correctly.

I cannot provide a decent teaching implementation at the moment. Most I can do is provide a working one.

Sayrus:
PR is not active anymore, closing.

PR #452, opened 2019-03-02
Fix zipper typespecs

Cohen-Carlisle:
I noticed the typespecs were a bit off for this exercise. Also added parentheses to function calls in the tests, which required one additional change for line length.

### DIFF
--- a/exercises/zipper/zipper.exs
+++ b/exercises/zipper/zipper.exs
@@ -9,7 +9,7 @@ defmodule BinTree do
   `right` is the right subtree (nil if no subtree).
   """
   @type t :: %BinTree{value: any, left: BinTree.t() | nil, right: BinTree.t() | nil}
-  defstruct value: nil, left: nil, right: nil
+  defstruct [:value, :left, :right]
 
   # A custom inspect instance purely for the tests, this makes error messages
   # much more readable.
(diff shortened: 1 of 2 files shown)


exercises/zipper/zipper.exs
```diff
@@ -29,6 +29,9 @@ defmodule BinTree do
 end
 
 defmodule Zipper do
+  alias BinTree, as: BT
+  alias Zipper, as: Z
```
devonestes: A couple things:

1) If we're going to use an alias here we probably shouldn't use single letter alias names. It's not seen very often in Elixir libraries or applications.

2) Since we're in the `Zipper` module we can use `t()` in type specs in this module. Or, alternatively, we could use `__MODULE__.t()`, but the plain `t()` is probably more common in the community.

Cohen-Carlisle: I aliased these since `BT` and `Z` are already used in the typespecs. But I think it probably makes more sense to just use the full `BinTree` and `Zipper`.

Cohen-Carlisle:
@devonestes I got rid of the aliases and other overly terse names. I also noticed that `BinTree.inspect/2` seems to not actually work? Should it just be removed?

PR #456, opened 2019-03-17
fix bowling typespecs

Cohen-Carlisle:
I'm thinking we should have the non-error returns in an ok tuple as well. It's nicer to pattern match `{:ok, game}`  and `{:ok, score}` than `game when is_game(game)` or `score when is_integer(score)`. But that could be done later, if desirable.

### DIFF
--- a/exercises/bowling/bowling.exs
+++ b/exercises/bowling/bowling.exs
@@ -1,29 +1,30 @@
 defmodule Bowling do
+  @type t :: any()
+
   @doc """
     Creates a new game of bowling that can be used to store the results of
     the game
   """
-
-  @spec start() :: any
+  @spec start() :: t()
   def start do
   end
 
   @doc """
     Records the number of pins knocked down on a single roll. Returns `any`
     unless there is something wrong with the given number of pins, in which
-    case it returns a helpful message.
+    case it returns a helpful error tuple.
   """
 
-  @spec roll(any, integer) :: any | String.t()
+  @spec roll(t(), integer()) :: t() | {:error, String.t()}
   def roll(game, roll) do
   end
 
   @doc """
     Returns the score of a given game of bowling if the game is complete.
-    If the game isn't complete, it returns a helpful message.
+    If the game isn't complete, it returns a helpful error tuple.
   """
 
-  @spec score(any) :: integer | String.t()
+  @spec score(t()) :: integer() | {:error, String.t()}
   def score(game) do
   end
 end


exercises/bowling/bowling.exs
```diff
@@ -1,29 +1,30 @@
 defmodule Bowling do
+  @type t :: any()
```
devonestes: I don't see what this type definition adds here. What was your intention for adding this?

exercises/bowling/bowling.exs
```diff
@@ -1,29 +1,30 @@
...
 
   @doc """
     Records the number of pins knocked down on a single roll. Returns `any`
     unless there is something wrong with the given number of pins, in which
-    case it returns a helpful message.
+    case it returns a helpful error tuple.
   """
 
-  @spec roll(any, integer) :: any | String.t()
+  @spec roll(t(), integer()) :: t() | {:error, String.t()}
```
devonestes: Typically if there's an `{:error, _}` response, the other possible response to a function is either `:ok` or `{:ok, _}`. If we want to recommend that these functions returning tagged tuples that's fine, but we should be consistent in that case.

devonestes:
Traditionally we've avoided making breaking changes unless they were fixing some sort of bug, since that would effectively invalidate all previous solutions to this exercise.

This change would require changing the test suite, so that would be a breaking change.

Personally I don't think this is worth that cost, but I'd like to hear your thoughts on it.

NobbZ:
> Traditionally we've avoided making breaking changes unless they were fixing some sort of bug, since that would effectively invalidate all previous solutions to this exercise.

This is only true in the the old exercism. The new exercism will associate the solution with the correct test suite.

----
 
A problem that I see here, without having look at the exercise itself… Since `t` is `any`, how would one differentiate between a valid return of `{:error, …}` and one that indicates actual error…

Any type that finally gets expanded to `any | …` is pretty useless as a return value.

exercises/bowling/bowling.exs
Cohen-Carlisle: To make clear that the previous `any`s were all the same data structure. Perhaps `@opaque` would be more appropriate? The type would still need to be specified as `any` since we leave that up the the exercist.

Cohen-Carlisle:
So the tests are already testing for `{:error, "string"}`, the typespecs were just behind. Perhaps now is the time to update the tests to use `{:ok, _}` as well?

exercises/bowling/bowling.exs
devonestes: I don't think we necessarily need to make it clear that they're the same data structure since they technically don't have to be to make the tests pass.

Also, `t()` is (in most cases other than a few special ones in Elixir itself like `String.t()` or `Keyword.t()`) reserved for the definition of a struct that is defined within a module. Other uses for defining type `t()` are exceedingly rare in real world Elixir.

Cohen-Carlisle: Perhaps `bowling_game()` then? I don't think `t()` is bad, though, since it suggests but does not require a struct.
The data structures _do_ need to be the same data structure, though, something made clear by reading the tests but not by the typespecs. The first arguments to `roll` and `score` must always be the structure returned by `start`.

devonestes: Technically they don't need to be the same. For example, there could be:
```elixir
def roll(game, roll) when is_list(game) do
  game
  |> Enum.into(%{})
  |> roll(roll)
end

def roll(game, roll) when is_map(game) do
  # implementation here that returns a keyword list
end
```

angelikatyborska:
I'm currently doing many breaking changes to practice exercises to bring them up to date with canonical data in time for the v3 release. Syncing this repo to v2 has been turned off for a month now. My thinking is that the shock of v3 in general is going to make it easier to accept the specific changes to specific exercises. And breaking changes weren't a problem in v2 either, like already was said.

Let's fix those typespecs at last :). I have recreated this PR as https://github.com/exercism/elixir/pull/652 because this one is very much out of date with `main`.

PR #465, opened 2019-04-11
Fixed the "-i" flag in the last test case.

neenjaw:
In the original, it is `"i"`, now changed to `"-i"` for consistency.


Cohen-Carlisle:
Sorry this sat so long. Thanks for this!

PR #467, opened 2019-04-14
Split encode/decode tests

nathanchere:
Encode and decode tests should be kept separate.

Aside from the principle that tests should be granular and only be testing one thing at a time, it can be particularly frustrating to be debugging 'failures' on a test when for example only wanting to work on encode first and decode later or vica versa.


nathanchere:
Note that this change will not break any previously working solutions.

neenjaw:
nathan, not sure if you noticed but your build is failing because of trailing whitespace.  Also, the repo is a few commits ahead of your PR now.

PR #477, opened 2019-05-26
Rename roman.exs to roman_numerals.exs

cetinajero:
Just a little filename rename for consistency


cetinajero:
Hi @devonestes,

> Which do you think would be better?

Based on other exercises it feels right to give it the complete name of the folder and exercise name (i.e. `RomanNumerals` and `RomanNumeralsTest`) but that implies that the asserts will read as `RomanNumerals.numerals(9) == "IX"`...

It seems to me that the function should also be renamed to `numeral` so it reads as `RomanNumerals.numeral(9) == "IX"` but... maybe on a separate PR?

devonestes:
@cetinajero This looks good to go, but it just needs to be updated to the new exercise structure.

cetinajero:
Sure @devonestes, done 👍

Issue #478, opened 2019-06-04
Code.load_file/2 soft deprecated


NobbZ:
`Code.load_file/2` as its used currently in the tests has been soft deprecated a year ago (elixir-lang/elixir#7201). It will be hard deprecated in elixir 1.9 and spit out warnings.

Instead one of `Code.require_file/2`, `Code.compile_file/2`, or `Code.eval_file/2` should be used.

neenjaw:
What about doing something like this?

```elixir
# test.exs
ExUnit.start(autorun: false)
ExUnit.configure(exclude: :pending, trace: true)

Code.require_file("two_fer.exs", __DIR__)
Code.require_file("two_fer_test.exs", __DIR__)

ExUnit.Server.modules_loaded()

ExUnit.run()
```

```elixir
# two_fer_test.exs
defmodule TwoFerTest do
  use ExUnit.Case

  test "no name given" do
    assert TwoFer.two_fer() == "One for you, one for me"
  end

  @tag :pending
  test "a name given" do
    assert TwoFer.two_fer("Gilberto Barros") == "One for Gilberto Barros, one for me"
  end

  # and so on...
end
```

Advantages:

- uses the new functions
- decouples the test cases from the implementation of the test framework

Disadvantages:

- adds a new file to the problem checkout

---

I have been doing a lot of reading about the auto-analyzer project that exercism is undertaking, and it might be useful to decouple these two aspects so that they can be used separately in an implementation.

Thoughts?

Cohen-Carlisle:
I'm not sure which one is the most appropriate, honestly. It seems like any of them will work and `eval_file` is the simplest? I'm honestly more in favor of moving over to using mix projects for each exercise and this would be a good excuse to do so.

neenjaw:
I think this could be a first step though, we eliminate the problem when 1.9 drops, and then we can just transplant the test case to the proper location if/when/once we move to mix

neenjaw:
Now resolved with PR #480

PR #480, opened 2019-06-16
Mixify All Exercises

neenjaw:
### Abstract

Taking #478 and #420 as inspiration, here is a pull request which puts every exercise as a separate mix project.  The mix project is created using the main class as the project name and entry point.  There are some things which I would like to have your input on:

if you were to run `mix new ocr-numbers` the main module would be named `OcrNumbers` instead of `OCRNumbers`.  Which would be preferrable?

Because of this, while I would like you to look at the whole commit, but particularly these ones required some manual adjustment to make consistent within a mix project.

* atbash-cipher
* bank-account
* difference-of-squares
* dot-dsl
* isbn-verifier
* kindergarten-garden
* largest-series-product
* leap
* nth-prime
* ocr-numbers
* palindrome-products
* phone-number
* pythagorean-triplet
* queen-attack
* rna-transcription
* series
* word-count
* zipper

```
acronym
├── lib
│   └── acronym.ex
├── test
│   ├── acronym_test.exs
│   └── test_helper.exs
├── example.exs
├── mix.exs
└── README.md
```

### Repo File Structure

```
.
├── bin/
│   ├── dialyzer_check.sh
│   └── fetch-configlet
├── config/
│   ├── exercise_readme.go.tmpl
│   └── maintainers.json
├── docs/
│   ├── ABOUT.md
│   ├── INSTALLATION.md
│   ├── LEARNING.md
│   ├── RESOURCES.md
│   ├── SNIPPET.txt
│   └── TESTS.md
├── exercises/
│   ├── accumulate/
│   ├── acronym/
│   ├── ...
│   ├── wordy/
│   ├── zipper/
│   └── test_helper.exs
├── config.json
├── LICENSE
├── mix.exs
├── mix.lock
└── README.md
```

### Test Suite

The test suite for each exercise has been updated, and can be run with `mix test` inside of the exercise folder.

The CI test suite has also been updated with regard to the new mixified structure and can be run from the repo base folder with `mix test`.  See [`exercises\test_helper.exs`](https://github.com/neenjaw/elixir/blob/mixify-exercises/exercises/test_helper.exs) for the update.

### Docs

[`TESTS.md`](https://github.com/neenjaw/elixir/blob/mixify-exercises/docs/TESTS.md) has also been updated to reflect the changes.


neenjaw:
Oh, this will fail the in-place Travis CI as the current requires `elixirc` to successfully complete

```bash
# will fail at present due to the changes made.
EXERCISM_TEST_EXAMPLES=true elixirc --warnings-as-errors **/example.exs **/*_test.exs -o ./tmp > /dev/null
```

This could be mitigated and be able to complete as it is if the following was prepended to each test suite in the `test` folder of each exercise:

```elixir
if System.get_env("EXERCISM_TEST_EXAMPLES") do
  ExUnit.start()
  ExUnit.configure(exclude: :pending, trace: true)
end
```

neenjaw:
Now TravisCI fails on the command:

```bash
$ ! git grep ' $' -- \*.exs | grep -v -e 'exercises/diamond/diamond_test.exs'
```
What is this testing?

NobbZ:
> To pass the TravisCI command:
>
> ```
> EXERCISM_TEST_EXAMPLES=true elixirc --warnings-as-errors **/example.exs **/*_test.exs -o ./tmp > /dev/null
> ```

Instead I'd prefer if travis would be changed to use `mix test` instead…

devonestes:
A couple points

1) None of these exercises should have any configuration, so that directory should be removed.

2) I think we need to remove the example from each Mix project. It has the same module name as the actual file that learners are supposed to fill in, and we don’t want them to be able to pull it down when they fetch exercises with the API.

NobbZ:
The example has to stay in the repo, it is to prove whether or not the exercise is solvable in the language. But as long as it's filename matches the example pattern given in the track config, it shouldn't get delivered to students.

NobbZ:
> What is this testing?

Looks like it were testing for some specific character sequence in any other file than the diamond tests, but maybe I'm misreading it... Perhaps you can use `git blame` to track changes down to a commit or even PR where that is discussed?

neenjaw:
Yes, that is why I left `example.exs` in the same location as the previous iteration. 

@devonestes -- alright, I will submit an iteration of this PR removing the `config` folder

@NobbZ -- Good idea, I will try to track down that command addition. Depending on what I find, maybe I'll submit an iteration of the PR with an updated `.travis.yml`

neenjaw:
The travisCI line is from issue #159, which resulted in commit `c09705a` ~3 years ago. Will do some checking, maybe I've left some trailing white space unnecessarily and might pass once I've cleaned it up.

neenjaw:
1. Config directories removed
2. Trailing white space issues fixed
3. `diamond_test.exs` exemption from trailing whitespace check removed from TravisCI as the test has been redefined to not require the exception.

.gitignore
```diff
@@ -6,3 +6,4 @@ bin/configlet
 bin/configlet*
 deps
 _build
+.elixir_ls/
```
neenjaw: so to not include ElixirLS build files from VSCode

.travis.yml
```diff
@@ -20,7 +20,7 @@ script:
 
   - travis_wait 30 bin/dialyzer_check.sh
 
-  - "! git grep ' $' -- \\*.exs | grep -v -e 'exercises/diamond/diamond_test.exs'"
+  - "! git grep ' $' -- \\*.exs"
```
neenjaw: `diamond`'s test file doesn't require trailing white space anymore, so exemption is removed

.gitignore
NobbZ: In my opinion this should **not** be part of a per-project `.gitignore` but the users global one instead, see also https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

neenjaw: Sure, can do that

neenjaw:
Are there issues remaining with this pull request?  I have time this week to work on them if there are.

Cohen-Carlisle:
I don't see any big outstanding issues. I think it makes sense to get this merged and if there are any other problems or fixes to roll forward with them.
The CI tests look good and seem to be running the tests for all exercises.
I'll merge this in a few days unless anyone has any objections.

Cohen-Carlisle:
I did have a few more thoughts that I wanted to document somewhere that could be follow up tasks or not done at all. In order of relative importance:
1. Should the elixir requirement in mix.exs be relaxed a bit? Currently we require 1.8 which is pretty new. This should probably be relaxed, but I'm not sure how the compatibility would function.
1. Is there a way to remove the `if System.get_env("EXERCISM_TEST_EXAMPLES") do ... end` code? It looks like this is used by CI... perhaps CI could insert the code into the files instead of making students read weird CI code?
1. Are we sure we want to remove `config/config.exs`? It generates with the projects by default so it makes some degree of sense to leave it in. This would allow interested students to also read some of the comment documentation in the config to learn a bit about configuring elixir projects and might spur them to read more about it later.

neenjaw:
Here are my thoughts in response to your 3 points, numbering maintained for organizational clarity.

1. I'm not super familiar enough with how each stable release has changed the library, but I think it's been relatively stable for the most part since 1.7 for everything as far as these exercises go.  `1.7` brought in the the changes to Enum interface, e.g., `Enum.chunk/2/3/4` in favour of `Enum.chunk_every/2/3/4`.  
I guess this comes down to a more philosophical question, should we promote people to use the newest stable release of elixir or not?
2. I believe the CI script to be adjusted to:
    i. concat all of the test files to a temp file
    ii. then prepend that snippit to it
    iii. then compile the examples with the merged testfile (leaving the test files to their original state, and less weird)
    iv. clean up the temp file
3. As of Elixir 1.9, `mix new` does not generate `config/config.exs` so I don't think there is too much of a loss going forward to not include it.

NobbZ:
> Should the elixir requirement in mix.exs be relaxed a bit? Currently we require 1.8 which is pretty new. This should probably be relaxed, but I'm not sure how the compatibility would function.

In erlang I do test against the last 3 OTP versions in their latest patch release. Perhaps adopting this for elixir is a good idea as well, resulting in 1.7, 1.8, 1.9.

I wouldn't test all of those on all supported OTP versions, but only oldest and newest per elixir version.

> Is there a way to remove the if System.get_env("EXERCISM_TEST_EXAMPLES") do ... end code? It looks like this is used by CI... perhaps CI could insert the code into the files instead of making students read weird CI code?

I think the easiest where to delete the stub and just doing `mix test` in each exercises folder, rather than some global command on a joined test file.

> Are we sure we want to remove config/config.exs?

Yes, since as @neenjaw has already said, `config/` won't be generated anymore with elixir 1.9.

neenjaw:
The value in the global test is that it tests the example for each exercise against the test suite for each problem.

I'm not sure how to achieve that on a per-exercise scale.

NobbZ:
As the example lives in each project, I'm not sure if I understand you…

The elixir compiler does not care if in an `*.ex` file the module `Foo` or `Bar` gets defined. So we can define any module in `example.ex` we want.

So just dropping the `example.ex` in `src`, while deleting the stub file to avoid module redefinition.

For local testing though, a "runner" should be available that runs against a copy in a temporary folder, such that we do not delete the stubs ;)

Running tests per exercise would also fail when example files end up in the wrong project for some reason.

The current approach might accept when one accidentally swaps out example files between projects.

neenjaw:
@NobbZ Ok, I'm picking up what you're putting down now. Let me see over the next few days of I can get that running.

neenjaw:
Ok, I have a plan for this:
1. TravisCI file to include change to test against elixir `1.7.2`, `1.8.2`, `1.9.0` and OTP `20.3` and `22.0`
2. Change the mix.exs:
    - require `~> 1.7`
    - alias the compile flag `--warnings-as-errors`
3. Bash script to do as @NobbZ described:
    -Create temp exercise copy
    -for each exercise, remove stub, copy example to lib, mix test for each
    -if error, fail
4.TravisCI call that script, like it does for dialyzer

Will work on this... will commit a new update when have it done.

NobbZ:
As you are wrapping in a shell script anyway, perhaps do not fail on the first failed exercise, but let all run and print a final list of failed tests? I do that in erlang.

neenjaw:
That sounds fair

neenjaw:
Summary of changes:
- exercises tested individually against example solution, results reported
- each exercise `mix.exs` has had the `elixir: "~> 1.8"` commented out to allow people to run older, but valid elixir run times, so as to promote language fluency over programming pedantics
- in each test, old test loader code used now commented out, to be removed at a later time
- travisCI build updated for newer versions of elixir (1.9.0) and OTP (22)
- travisCI build script updated, no longer uses elixirc directly, but is tested in above mentioned script

.travis.yml
```diff
@@ -1,25 +1,30 @@
...
-  - mkdir ./tmp
-  - EXERCISM_TEST_EXAMPLES=true elixirc --warnings-as-errors **/example.exs **/*_test.exs -o ./tmp > /dev/null
-  - rm -rf ./tmp
+  # OLD - fail if there are any compiler warnings
+  # - mkdir ./tmp
+  # - EXERCISM_TEST_EXAMPLES=true elixirc --warnings-as-errors **/example.exs **/*_test.exs -o ./tmp > /dev/null
+  # - rm -rf ./tmp
+
+  # NEW - fail if there are any compiler and test problems for individual tests.
+  - bin/test_exercises.sh
```
Cohen-Carlisle: Since we have this in git history anyway, I would prefer to simply delete the commented out code and not label line 22 with `NEW - `

.travis.yml
```diff
@@ -1,25 +1,30 @@
 language: elixir
 elixir:
-  - 1.6.6
```
Cohen-Carlisle: I think it's fine to keep 1.6.6 here, especially since it's still receiving security fixes.

exercises/accumulate/test/accumulate_test.exs
```diff
@@ -1,7 +1,7 @@
-if System.get_env("EXERCISM_TEST_EXAMPLES") do
-  ExUnit.start()
-  ExUnit.configure(exclude: :pending, trace: true)
-end
+# if System.get_env("EXERCISM_TEST_EXAMPLES") do
+#   ExUnit.start()
+#   ExUnit.configure(exclude: :pending, trace: true)
+# end
```
Cohen-Carlisle: Since these are in git history, I don't see any reason not to delete them, especially as their presence could confuse students.

exercises/accumulate/mix.exs
```diff
@@ -0,0 +1,35 @@
...
+      deps: deps(),
+      aliases: aliases()
+    ]
+  end
+
+  def aliases do
+    [
+      compile: ["compile --warnings-as-errors"]
+    ]
+  end
```
Cohen-Carlisle: Is the `--warnings-as-errors` flag intended for student consumption, or merely for the tests? As written, student code will fail on their machines when they run `mix test`. That might actually be okay, but it is a change from what's happening right now. If this is only intended for the tests, the test script could be extended to do a `MIX_ENV=test mix compile --warnings-as-errors` and check the exit code before the `mix test` for each exercise. If we do end up keeping the aliases, the aliases function should be made private with `defp` and moved after into the private function section (I'd say after `defp deps`). I would lean towards not having this alias present in the config to minimize changes to the student experience in this PR.

bin/test_exercises.sh
```diff
@@ -0,0 +1,92 @@
...
+    mv example.exs lib/example.ex
+
+    # do testing then based on exit code print the result
+    mix test --no-elixir-version-check --include pending > /dev/null
+
+    if [ "$?" -eq 0 ]; then
+        printf " -- pass ✅\n"
+        pass_count=$((pass_count+1))
+    else
+        printf " -- fail ❌\n"
```
Cohen-Carlisle: In the failing case, it would be nice to know what actually failed. The simplest thing would be to change
```
mix test --no-elixir-version-check --include pending > /dev/null
```
to
```
mix test --color --include pending > output
```
cat output
```
after this `printf`.

File: .travis.yml (hunk 4 above)
Comment (neenjaw) [reply .travis.yml]: I'm alright with that, I think was a decision based on @NobbZ comment to test against last three major versions.

File: .travis.yml (hunk 3 above)
Comment (neenjaw) [reply .travis.yml]: For sure, I had left it commented to convert the change as I have in my personal practice, but I forget how rich GitHub is for looking at these things. I will amend.

Comment (neenjaw) [reply .travis.yml]: Re line 15, good point. I think we can get rid of it. Had left it because it worked and was fast, but if it were to fail Im not sure it would be helpful output.

File: exercises/accumulate/test/accumulate_test.exs (hunk 5 above)
Comment (neenjaw) [reply exercises/accumulate/test/accumulate_test.exs]: See above, agreed

File: bin/test_exercises.sh (hunk 7 above)
Comment (neenjaw) [reply bin/test_exercises.sh]: True, would save you a step, I like it, will add it in.

File: exercises/accumulate/mix.exs (hunk 6 above)
Comment (neenjaw) [reply exercises/accumulate/mix.exs]: I had put it in initially just to facilitate the individual testing for CI, but one thing I do like about this is that I find students often ignore warnings by the compiler, eg:

```
iex> import Date, only: [utc_today: 0]
Date
iex> utc_today
warning: variable "utc_today" does not exist and is being expanded to "utc_today()", please use parentheses to remove the ambiguity or change the variable name
```

I don't mind having it in, because I think it will help steer students toward better coding practices, but I can see that this is a departure from what is currently being done.

Am willing to remove it if it is a breaking change.

File: bin/test_exercises.sh (hunk 7 above)
Comment (neenjaw) [reply bin/test_exercises.sh]: New version will output the following on fail:

```
➜ bin/test_exercises.sh
Testing: accumulate -- fail ❌
  * test accumulate upcased strings (5.7ms)
  * test accumulate square numbers (4.2ms)
  * test accumulate reversed strings (186.6ms)
  * test nested accumulate (0.00ms)
  * test accumulate empty list (0.00ms)
Testing: acronym -- pass ✅
Testing: allergies -- pass ✅
Testing: all-your-base -- pass ✅
Testing: alphametics -- pass ✅
Testing: anagram -- pass ✅
```

**Comment by NobbZ** [comment]
there are currently builds failing because 1.6 is not advertised as compatible with erlang 22. Therefore no prebuilt binaries do exist.

This has to be removed via build matrix.

```yaml
matrix:
  exclude:
  - elixir: 1.6.6
    otp_release: 22
```

neenjaw:
> there are currently builds failing because 1.6 is not advertised as compatible with erlang 22. Therefore no prebuilt binaries do exist.
> 
> This has to be removed via build matrix.
> 
> ```yaml
> matrix:
>   exclude:
>   - elixir: 1.6.6
>     otp_release: 22
> ```

Thanks @NobbZ! I saw that the build had failed, and was going to research how to do just that! Great timing!

bin/test_exercises.sh
neenjaw: This is added in now.  I think in the future it might be handy to have a custom result formatter for this purpose as the output (while readable to humans) isn't as easily grepped for the results and so it feels a bit janky matching on ANSI color codes to determine the individually failing tests, but it works for now.

exercises/accumulate/mix.exs
Cohen-Carlisle: The more I think about this one the more I think we should not do it without a wider discussion since a) it does introduce new blockers into the workflow and b) it is not the default config for mix projects.

bin/test_exercises.sh
```diff
@@ -0,0 +1,110 @@
...
+      printf " -- \\033[31mFail\\033[0m\n"
+
+
+      test_results=$(echo "$test_results" \
+                      | grep "* test" \
+                      | sed 's:\r:\n:' \
+                      | awk 'NR % 2 == 1' \
+                      | sed -r "s:\x1B\[31m:$(printf " -- \\033[31mFail\\033[0m"):g" \
+                      | sed -r "s:\x1B\[32m:$(printf " -- \\033[32mPass\\033[0m"):g")
+      printf "${test_results}\n"
```
Cohen-Carlisle: I'm not sure this does what we want in a portable way. Running this on mac on elixir 1.8.2/erlang 22 `grep "* test" \` comes up empty because failing tests are printed like this: `1) test`. My version of `sed` also does not have an `-r` option. And at least for the way my `mix test` output if formatted, this would lose the stacktrace in the `grep`.

neenjaw: Interesting... 

I suppose then for now can have it just print the whole output from the failing test.  The most robust way I think to have it do something more than that is to write a custom `ExUnit.CLIFormatter` module.

exercises/accumulate/mix.exs
neenjaw: sounds good, will revert

exercises/accumulate/mix.exs
```diff
@@ -0,0 +1,35 @@
...
+  defp deps do
+    [
+      # {:dep_from_hexpm, "~> 0.3.0"},
+      # {:dep_from_git, git: "https://github.com/elixir-lang/my_dep.git", tag: "0.1.0"}
+    ]
+  end
+
+  defp aliases do
+    [
+      compile: ["compile --warnings-as-errors"]
```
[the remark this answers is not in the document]
neenjaw: same removed, see above review conversation

neenjaw:
## Summary of changes
- new exercise directory format to follow mix application structure
- exercise testing now done with `bin/test_exercises.sh` testing both individual compilation and unit tests
- docs updated to reflect new structure for new learners to the elixir track

## Work to be done
- none?

## Issues Pending
- awaiting feedback from website-copy in order to understand how merge will propagate to exercism

Cohen-Carlisle:
Well, here goes nothing! :crossed_fingers:

PR #482, opened 2019-06-28
Adds test for very long input

### LINKED ISSUE #481: rna-transcription should test a long list
I've noticed some students folding their charlist using an append function, instead of a prepend function.  Since all the tests use a very small charlist, they don't see a problem with this.  I think it would be a good idea to add a test with a charlist that's long enough to showcase the benefit of prepending.

miread:
This test builds a charlist of 500,000 codepoints to be tested.  This should not take much time to process, unless the student's algorithm reduces the list via appending/concatenation, in which case it will probably not finish.  Hopefully this test will showcase the poor performance of appending vs prepending.

resolves #481

### DIFF
--- a/exercises/rna-transcription/rna_transcription_test.exs
+++ b/exercises/rna-transcription/rna_transcription_test.exs
@@ -32,7 +32,7 @@ defmodule RNATranscriptionTest do
   test "transcribes all dna nucleotides to rna equivalents" do
     assert RNATranscription.to_rna('ACGTGGTCTTAA') == 'UGCACCAGAAUU'
   end
-  
+
   @tag :pending
   test "transcribes very long dna" do
     dna_list = Enum.reduce(1..500000, [],
@@ -41,7 +41,6 @@ defmodule RNATranscriptionTest do
     rna_list = Enum.reduce(1..500000, [],
                          fn x, acc -> if x <= 250000, do: [?U | acc], else: [?G | acc]
                          end)
-    
     assert RNATranscription.to_rna(dna_list) == rna_list
   end
 end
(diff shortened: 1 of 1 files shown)


miread:
I'm not sure what it means that the Travis CI build failed.  I did run the updated test suite on my computer and it worked ok.  When Travis was doing the check, it seemed to take a while..  Maybe it's trying the test the "wrong way" and timing out?

NobbZ:
The the trailing whitespace test that fails. Please check your added lines for trailing spaces and tabs.

NobbZ:
There are still 2 empty lines with trailing whitespace.

miread:
Thanks, @NobbZ!

neenjaw:
I think this is is a good test-case to add in, as I also see this come up lots.  I don't think this will prevent students from making that mistake, but at least gives the mentor more space to comment on the poor performance of using `++` and how lists work in general.

neenjaw:
Oh, Have you thought about using a stream for this?

```elixir
{dna, rna} = 
  [{?A, ?U}, {?C, ?G}, {?T, ?A}, {?G, ?C}] 
  |> Stream.cycle() 
  |> Stream.take(500_000) 
  |> Enum.unzip()
```

miread:
Thanks for the suggestion, @neenjaw!  Stream.cycle looks like a nice way of doing it, but it's much less efficient (each test takes about twice as long to setup).

devonestes:
I'm not a huge fan of making sure the implementation is performant in the tests, so I don't think we should do this one. Performance optimizations are something that learners will get after they're much further along their learning journey.

PR #484, opened 2019-07-29
Fix instructions for each exercise

neenjaw:
The exercise readme.md was forgotten to be updated.  Now fixed, see diff.


neenjaw:
something messed up

PR #492, opened 2019-08-30
Enable tests by comment `@tag :pending`

### LINKED ISSUE #462: markdown as a refactoring exercise should have all tests enabled from the beginning
Since markdown is a refactoring exercise, all tests should be enabled from the start.

Since refactoring is the art to change the code in a way that still everything works.

neenjaw:
Since this is a refactoring exercise, it should still pass all tests from the start, so enable them to be able to ensure that all refeactoings still pass the test suite.

Close #462

### DIFF
--- a/exercises/markdown/test/markdown_test.exs
+++ b/exercises/markdown/test/markdown_test.exs
@@ -8,56 +8,56 @@ defmodule MarkdownTest do
     assert Markdown.parse(input) == expected
   end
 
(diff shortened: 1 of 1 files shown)


neenjaw:
Any objection if I merge this? Will do so in 24hrs or so unless there is an objection.

neenjaw:
:+1: Thanks @devonestes!

PR #498, opened 2019-10-01
Port dnd-character

filalex77:


neenjaw:
Your build is failing for two reasons:

1) config.json must be updated in order for TravisCI to pass, this required the exercise to be registered for this track using the configlet (I'm not sure that we have elixir-specific documentation for this -- @Cohen-Carlisle can you help?)
2) The track has, thus far, not used external dependencies for testing.   I'm not sure what the stance is, but I can see a lean towards not including external dependencies for the sake of complexity with the build, and, eventually, the `exercism/elixir-test-runner` project. 
I can see why you have based on the `ability/0` function's use of random.  Is there a way you might be able to do this using built-ins? Eg, generating a list of `x` length with the `ability/0` function and then testing the results?  I think the limitation is that this may only test repeatability rather than randomness of the function. (@devonestes -- has there been previous decisions to limit outside libraries, @NobbZ -- will this affect the test-runner's performance, particularly if it depends on getting and compiling a dependency)

filalex77:
@neenjaw I added an entry to `config.json` using `configlet`, and replaced `excheck` with `Enum.each/2`.

neenjaw:
Oh, you also need to mark which exercise you would like to unlock the exercise.  

>The exercise 'dnd-character' is being unlocked by a non-core exercise. Non-core exercises can only be unlocked by core exercises.

Maybe the Robot-simulator exercise can unlock it for more practice with structs?

Could then also tag it with 'structs' to hint at the solution we would like to see

---

You have a compiler error

```
Compiling 1 file (.ex)
== Compilation error in file lib/example.ex ==
** (CompileError) lib/example.ex:12: undefined function floor/1
    (stdlib) lists.erl:1338: :lists.foreach/2
    (stdlib) erl_eval.erl:670: :erl_eval.do_apply/6
```

Need `:math.` to prefix

---

Also there is some implicit type casting going on as `:math.floor/1` returns a float and your test suite does a comparison `==` to `6`, I think we should decide with a typespec what we want these to return in the stub.

exercises/dnd-character/lib/dnd_character.ex
```diff
@@ -0,0 +1,7 @@
+defmodule DndCharacter do
+  def modifier(_score)
```
neenjaw: might be better to stub it something like:

```elixir
@spec modifier(int()) :: int() # or whatever the types should be
def modifier(score) do
  # replace with your solution code
end
```

exercises/dnd-character/example.exs
```diff
@@ -0,0 +1,29 @@
...
+    :strength,
+    :dexterity,
+    :constitution,
+    :intelligence,
+    :wisdom,
+    :charisma,
+    :hitpoints
+  ]
+
+  def modifier(score), do: floor((score - 10) / 2)
```
neenjaw: needs to be `:math.floor`

exercises/dnd-character/test/dnd_character_test.exs
```diff
@@ -0,0 +1,110 @@
...
+  use ExUnit.Case
+
+  import DndCharacter, only: [modifier: 1, ability: 0, character: 0]
+
+  # canonical data version: 1.1.0
+
+  describe "ability modifier" do
+    # @tag :pending
+    test "for score 3 is -4" do
+      assert modifier(3) == -4
```
neenjaw: I think to test for type, it should be `===` vs `==`

exercises/dnd-character/example.exs
filalex77: Not sure on this. I intended to use [`Kernel.floor/1`](https://hexdocs.pm/elixir/Kernel.html#floor/1), and that should be available.

exercises/dnd-character/test/dnd_character_test.exs
filalex77: Sure, strict equality seems like a good fit here.

exercises/dnd-character/example.exs
neenjaw: Kernel.floor is only available in >= 1.8

filalex77: Okay. Well, I'll go figure out a 1.6.6 solution.

neenjaw:
Looks great, I can't merge this yet (don't have write access yet) but I will when I can (likely the next few days pursuant to #499).

I would like @Cohen-Carlisle to review once just to make sure I'm not missing something.

exercises/dnd-character/mix.exs
```diff
@@ -0,0 +1,25 @@
+defmodule DndCharacter.MixProject do
+  use Mix.Project
+
+  def project do
+    [
+      app: :dnd_character,
+      version: "1.1.0",
```
Cohen-Carlisle: This seems to match up with the canonical data version, but I think it would be better to keep the generated `"0.1.0"`.

exercises/dnd-character/lib/dnd_character.ex
```diff
@@ -0,0 +1,15 @@
...
+  def modifier(score) do
+    # code here
+  end
+
+  @spec ability() :: pos_integer
+  def ability() do
+    # code here
+  end
+
+  def character() do
```
Cohen-Carlisle: A typespec here would be nice. There should be some examples of declaring a type (e.g., defining `.t()`) for the module and then referencing it in functions in this repo.

exercises/dnd-character/lib/dnd_character.ex
```diff
@@ -0,0 +1,15 @@
+defmodule DndCharacter do
+  @spec modifier(pos_integer) :: integer
+  def modifier(score) do
+    # code here
+  end
+
+  @spec ability() :: pos_integer
+  def ability() do
```
Cohen-Carlisle: I don't think this codebase uses empty parentheses for function definitions with zero arguments. I'd like to stay consistent.

exercises/dnd-character/lib/dnd_character.ex
```diff
@@ -0,0 +1,15 @@
+defmodule DndCharacter do
+  @spec modifier(pos_integer) :: integer
+  def modifier(score) do
+    # code here
```
Cohen-Carlisle: I don't think we usually put anything inside the the stub functions and just keep it blank.

exercises/dnd-character/mix.exs
```diff
@@ -0,0 +1,25 @@
...
+  # Run "mix help compile.app" to learn about applications.
+  def application do
+    [
+      extra_applications: [:logger]
+    ]
+  end
+
+  # Run "mix help deps" to learn about dependencies.
+  defp deps do
+    []
```
Cohen-Carlisle: I'd like to keep this file as "stock" as possible. Could we replace this line with:
```
    [
      # {:dep_from_hexpm, "~> 0.3.0"},
      # {:dep_from_git, git: "https://github.com/elixir-lang/my_dep.git", tag: "0.1.0"}
    ]
```

exercises/dnd-character/test/dnd_character_test.exs
```diff
@@ -0,0 +1,110 @@
...
+    end
+
+    @tag :pending
+    test "for score 18 is +4" do
+      assert modifier(18) === 4
+    end
+  end
+
+  @tag :pending
+  test "random ability is within range" do
```
Cohen-Carlisle: It's a bit odd that this is the only un`describe`d `test.

exercises/dnd-character/lib/dnd_character.ex
filalex77: Does defining a `DndCharacter.t()` as a map with required fields make sense?

```elixir
@type t :: %{                     
        strength: pos_integer,    
        dexterity: pos_integer,   
        constitution: pos_integer,
        intelligence: pos_integer,
        wisdom: pos_integer,      
        charisma: pos_integer,    
        hitpoint: pos_integer     
      }                           
```

filalex77:
@Cohen-Carlisle Thanks a lot for the detailed review! I've pushed a commit to incorporate those changes.

exercises/dnd-character/lib/dnd_character.ex
Cohen-Carlisle: It would be slightly better to use:
```
@type t :: %__MODULE__{                     
        strength: pos_integer,    
        dexterity: pos_integer,   
        constitution: pos_integer,
        intelligence: pos_integer,
        wisdom: pos_integer,      
        charisma: pos_integer,    
        hitpoint: pos_integer     
      }
```

As this communicates it will return the struct of the current module.

exercises/dnd-character/lib/dnd_character.ex
```diff
@@ -0,0 +1,23 @@
...
+
+  @spec modifier(pos_integer) :: integer
+  def modifier(score) do
+  end
+
+  @spec ability :: pos_integer
+  def ability do
+  end
+
+  @spec character :: DndCharacter.t()
```
Cohen-Carlisle: You could leave this as simply:
```elixir
   @spec character :: t()
```
as the current module is implied.

exercises/dnd-character/example.exs
```diff
@@ -0,0 +1,29 @@
...
+      strength: ability(),
+      dexterity: ability(),
+      constitution: constitution,
+      intelligence: ability(),
+      wisdom: ability(),
+      charisma: ability(),
+      hitpoints: 10 + modifier(constitution)
+    }
+  end
+end
```
Cohen-Carlisle: It's not strictly necessary but it would be nice to have the example file be consistent with stub file (for example, having the same typespecs). In other words if you use the stub as the starting point for this file, then fill in the implementation, it gives some nice symmetry.

neenjaw:
Cleared it with @filalex77, last changes implemented.

1) dnd_character.ex is now the starting point for example.ex

neenjaw:
Agh, they changed the stuct interface checks between 1.7.2 and 1.8.2....

PR #505, opened 2019-10-07
Dnd character

Cohen-Carlisle:


exercises/dnd-character/example.exs
```diff
@@ -0,0 +1,42 @@
...
+    %__MODULE__{
+      strength: ability(),
+      dexterity: ability(),
+      constitution: constitution,
+      intelligence: ability(),
+      wisdom: ability(),
+      charisma: ability(),
+      hitpoints: 10 + modifier(constitution)
+    }
+  end
```
neenjaw: I think this is the better way.  Let's do this.

exercises/dnd-character/example.exs
```diff
@@ -0,0 +1,42 @@
+defmodule DndCharacter do
+  @type t :: %__MODULE__{
+    strength: pos_integer(),
+    dexterity: pos_integer(),
+    constitution: pos_integer(),
+    intelligence: pos_integer(),
+    wisdom: pos_integer(),
+    charisma: pos_integer(),
+    hitpoints: pos_integer()
+  }
```
neenjaw: I think putting the __MODULE__ is important on my 3rd thought through it to have as an example for others.  This is better.

exercises/dnd-character/example.exs
```diff
@@ -0,0 +1,42 @@
...
+    Integer.floor_div(score - 10, 2)
+  end
+
+  @spec ability :: pos_integer()
+  def ability() do
+    1..4
+    |> Enum.map(fn _ -> Enum.random(1..6) end)
+    |> Enum.sort()
+    |> Enum.drop(1)
+    |> Enum.sum()
```
neenjaw: I like it, very clean.

exercises/dnd-character/lib/dnd_character.ex
```diff
@@ -0,0 +1,25 @@
...
+    strength: pos_integer(),
+    dexterity: pos_integer(),
+    constitution: pos_integer(),
+    intelligence: pos_integer(),
+    wisdom: pos_integer(),
+    charisma: pos_integer(),
+    hitpoints: pos_integer()
+  }
+
+  defstruct ~w[strength dexterity constitution intelligence wisdom charisma hitpoints]a
```
neenjaw: I like this

config.json
```diff
@@ -1020,6 +1020,17 @@
         "algorithms",
         "math"
       ]
+    },
+    {
+      "slug": "dnd-character",
+      "uuid": "bd15a0e0-c19e-4a52-be6c-1a61d14c54cd",
+      "core": false,
+      "unlocked_by": "robot-simulator",
```
neenjaw: Maybe we should put this earlier so that people are exposed to structs before robot in a smaller scale exercise?

config.json
```diff
@@ -1020,6 +1020,17 @@
...
+      "difficulty": 3,
+      "topics": [
+        "random",
+        "structs"
```
neenjaw: maybe also typespec?

Cohen-Carlisle: I definitely think it's worth discussing having students do typespecs, but I also think we should look at evaluating that for the track as a whole.

config.json
Cohen-Carlisle: This exercise definitely seems easier than robot-simulator, so I like that idea.

config.json
neenjaw: Then let's leave as is for now, and eval as a whole later

neenjaw: Then let's leave as is for now, and eval as a whole later

Issue #506, opened 2019-10-09
Track maintenance issues


neenjaw:
With the release of @SleeplessByte's track maintenance tool, the elixir track can't make use of it just yet.

The track needs to:
- [x] Decide where problem version (against the canonical version) is recorded
- [x] Add this path to `tracks-maintenance-dashboard/src/data/tracks.json`
- [x] Ensure the regex used to parse the version in the dashboard is accurate for elixir

Are there thoughts, re: the first point?

We could record it in the mix.exs app version, and for now set the version to less than the canonical one, the as we are able to evaluate it against the canonical, change the version to match. 

We could also use a separate identifier in mix.exs or in another file, thoughts?

devonestes:
I would argue that this version is internal to exercism and if we put in in mix.exs we’d be leaning that to learners. I would favor a separate file that isn’t included when a learner clones the exercise to their computer.

neenjaw:
I think that's a good point.  

How do we specify what files get ignored when an exercise gets pulled? I know that the example.exs file doesn't get pulled, but where is that configured?

For the external file, is there a preferred format? JSON, YAML, etc?

neenjaw:
Is the .meta folder ignored by the download API call?

NobbZ:
I can't find any documentation for that, but I know that `bash` track is using it (and does not even have a label, but only the version in the `.meta/version`-file) successfully and I don't get the `.meta` downloaded when I do `bash` exercises.

neenjaw:
It could be just like how `example.exs` isn't downloaded for this track, is that configured in this repo somewhere (try as I have, I can't find where if it is)

NobbZ:
`config.json` does know some patterns, but I currently can't find the documentation…

Though non of the patterns does semantically fit.

We have a pattern for examples (defaults to `/[Ee]xample/` AFAIR), but the version is not part of the example, and we have a pattern for tests (which defaults to `/test/` AFAIR), but the version is not a test.

So if you do not want to put the version into student visible places, than `.meta` is the only location that makes sense in my eyes.

neenjaw:
So then a `.meta` folder with a `version` file, where the contents is simply: `vsn x.x.x`?

NobbZ:
Yeah, that would be possible, or as the `bash` track does it, they have a plain `.meta/version` which contains the plain version number, eg:

https://github.com/exercism/bash/blob/master/exercises/acronym/.meta/version

neenjaw:
[...]

PR #507, opened 2019-10-09
add version 0.0.0 to each exercise

neenjaw:
- creates `exercises/{slug}/.meta/version` file with the inital version set to `0.0.0`


Cohen-Carlisle:
Can you explain what these changes do? Is the version related to the canonical data from problem specifications?

neenjaw:
Forgot to link #506, yes, it is in regard to the problem specification and the track maintenance tool at https://tracks.exercism.io

devonestes:
LGTM :+1:

neenjaw:
@cohen-carlisle, in order to make use of the track maintenance tool, we need to add in what version our tests are built from. 

Since this repo doesn't currently track it, and adding them all with their current actual version is a lot of work. My thought is that by adding this with version `0.0.0` should at least make it so that the tool works for us so we can then start to pick away at the more time consuming labeling work.

neenjaw:
Ok, I've been looking over some of the other tracks, it is not uncommon for tracks to list exercises at version `0.0.0`, I think we should just merge this to get the tool running...

SleeplessByte:
Once [this](https://github.com/exercism/tracks-maintenance-dashboard/commit/ce69f0c886e9c4f1f11d102662b57b13fe012740/checks?check_suite_id=258741262) is green, it will be live [here](https://tracks.exercism.io/elixir/master/versions).

PR #509, opened 2019-10-09
nucleotide count fixup

Cohen-Carlisle:
Found an old branch with some nucleotide count fixes, so here some are!

### DIFF
--- a/exercises/nucleotide-count/example.exs
+++ b/exercises/nucleotide-count/example.exs
@@ -2,7 +2,7 @@ defmodule NucleotideCount do
   @nucleotides [?A, ?C, ?G, ?T]
 
   @doc """
-  Counts individual nucleotides in a NucleotideCount strand.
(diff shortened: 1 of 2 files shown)


neenjaw:
The only thing I see is that the typespec uses the bracket form of the type (E.g. `map()` vs `map`) and I think we had decided with dnd-character that the form without brackets was our standard.

Cohen-Carlisle:
If any parens were left off typespecs, that was a miss. I try to put them on since elixir style seems to [have them on](https://hexdocs.pm/elixir/typespecs.html). To be honest, I'm not sure why, though. I vaguely have a sense that they are functions and so end with parentheses.

neenjaw:
For sure! I don't mind either way, just remembered your comment and wanted to be consistent.

We should add it to the list to (at somepoint) do a sweep of the repo and ensure all of the typespecs are uniform.  I'll withdraw my suggestions

Issue #510, opened 2019-10-09
Verify exercise test versions against canonical test versions


neenjaw:
Now that we have implemented exercise version tracking against the canonical test versions at `exercism/problem-specifications` it is visible at [tracks.exercism.io](https://tracks.exercism.io/elixir/master/versions)

Now is a bit of a longer-term process of ensuring that our tests at least cover the test cases of the canonical versions.  In discussion with @sleeplessbyte and others, if we have test cases that are unaccounted for in the canonical that is okay -- we should try to just comment as such.

So I propose that over time (JavaScript and Typescript only do a few per week) we bring up our test cases to cover the canonical cases, then mark the exercise as the current canonical version in the `exercise/{slug}/.meta/version` file to match.

I have created a project to keep track of this progress, and to be able to call out which ones you might be working on, if unsure if it is complete, tracks.exercism.io will list unverified exercises as version `0.0.0`.

---

If there is no canonical test data version, please leave the exercise version at `0.0.0` in the case the test data is added at a later date, this will give us an easy way to identify if our test data moves out of sync with exercism/problem-specifications

neenjaw:
#511 The first 5 exercises reviewed.

filalex77:
Did 5 from `anagram` to `beer-song`.

neenjaw:
Going to close this with the impending change to the problem specifications repo and the introduction of the `tests.toml` file.

PR #512, opened 2019-10-10
Verify 'anagram' - 'beer-song' exercise versions

filalex77:
- [x] anagram
- [x] armstrong-numbers
- [x] atbash-cipher
- [x] bank-account: doesn't have `canonical-data.json`; put at `1.0.0`
- [x] beer-song

Related to #510. 

r? @neenjaw 

---
This change is [Reviewable](https://reviewable.io/reviews/exercism/elixir/512)


exercises/beer-song/test/beer_song_test.exs
```diff
@@ -1,356 +1,381 @@
 defmodule BeerSongTest do
   use ExUnit.Case
 
-  test "getting the first verse (99 bottles)" do
-    assert BeerSong.verse(99) == """
-           99 bottles of beer on the wall, 99 bottles of beer.
-           Take one down and pass it around, 98 bottles of beer on the wall.
-           """
+  describe "verse" do
+    describe "single verse" do
```
devonestes: Nested describes are not supported in Elixir, so you'll need to find another way to set this up.

filalex77: Thanks for pointing out! I think just dropping the second level is okay.

exercises/bank-account/.meta/version
```diff
@@ -1 +1 @@
-0.0.0
+1.0.0
```
neenjaw:
```suggestion
0.0.0
```

For tests that don't have data, we are going to leave it at `0.0.0` like the JavaScript and TypeScript group. I'll clarify in the issue.

filalex77: Sure, thanks for pointing out.

PR #138, opened 2016-03-01
Add queen attack exercise

dalexj:
I took all the tests from the [xruby version](https://github.com/exercism/xruby/blob/master/exercises/queen-attack/queen_attack_test.rb) except left one out for being overly redundant

PR #515, opened 2019-11-07
Added a test to Queen Attack

mexicat:
This new test checks that a board is correctly created when both queens are on the same row.

This is needed because someone — including me, on my first attempt — might do something like this:

```elixir
def to_string(%Queens{black: {bx, by}, white: {wx, wy}}) do
  Enum.map_join(0..7, "\n", fn
    ^bx ->
      empty_row() |> List.replace_at(by, "B") |> Enum.join(" ")

    ^wx ->
      empty_row() |> List.replace_at(wy, "W") |> Enum.join(" ")

    _ ->
      empty_row() |> Enum.join(" ")
  end)
end
```

which would pass the tests but, in case the two queens are on the same row, would only print the first one to match.

So with this new test the function would need an extra case:

```elixir
def to_string(%Queens{black: {bx, by}, white: {wx, wy}}) do
  Enum.map_join(0..7, "\n", fn
    ^bx when bx == wx ->
      empty_row() |> List.replace_at(by, "B") |> List.replace_at(wy, "W") |> Enum.join(" ")

    ^bx ->
      empty_row() |> List.replace_at(by, "B") |> Enum.join(" ")

    ^wx ->
      empty_row() |> List.replace_at(wy, "W") |> Enum.join(" ")

    _ ->
      empty_row() |> Enum.join(" ")
  end)
end
```

or probably a refactor into something more elegant :)


neenjaw:
Does this test case exist in the canonical exercism/problem-specifications repo?

World you be willing to so far as verify the current problem test suite against the canonical -- following the process in #510?

We are allowed to have extra test cases, so if we can arrange the test suite first in the order of the tests in the canonical repository, then add our additional ones after, it would be great.

mexicat:
Looking at the [problem spec](https://raw.githubusercontent.com/exercism/problem-specifications/master/exercises/queen-attack/canonical-data.json), the tests there seem to be quite different from the Elixir version.

In the Elixir version, there are basically three categories of tests:
- tests for the creation of a new `Queens` struct, which always has both `black` and `white` properties (with defaults)
- tests for the string representation of the board
- tests for the `can_attack?` function

The problem spec has only two kinds of tests:
- tests for the creation of queens, i.e. verification that they fits inside the board's bounds. These only test one queen at a time and they're completely missing from the Elixir test suite
- tests for the attack part, pretty much the same as the Elixir version

So there's no tests for the string representation of the board there (which is the category my new test falls in). And the first category of tests in the spec is very different from what we have in Elixir: I think adding those would require a complete refactor of the test suite/exercise because at the moment there's no way to express the creation of a single queen, they're always created in pairs with the `Queens` struct.

neenjaw:
Maybe @Cohen-Carlisle or @NobbZ can comment more precisely, but I think the eventual trajectory exercism is heading for is to have each language track's problem list driven from the problem specifcation list.

I think if our problem diverges too much from the mainstream, at some point ours will break and we will have to fix it then.  If you look back at #138 in the issues/PRs, when this problem was added it was added from the `xruby` track (which is now defunct and has been refactored in the new `ruby` track).

I'm suggesting that maybe we take the time now, to adjust our example solution and problem test suite to better matched to the canonical.  I think keeping the to_string tests are good, and your test case if a good catch for that!

@devonestes, do you think worthwhile to update the problem?

---

@mexicat - Sorry if this opens up a lot of issues about this problem that you weren't anticipating, I think PRs are a good opportunity for conversation about how things are and how things should be.  I really appreciate the time and effort you have put into creating this test case!

What do you think? Should we update it?

mexicat:
@neenjaw I agree with you, it would be better to fix it now rather than eventually. And it's not a problem, I'm glad that a simple PR sparked a larger discussion on what to do. If you give me the go-ahead, I can actually try refactoring the test suite and example exercise myself, adding the test cases from the spec and reordering everything so it's clear which tests are from the spec and which ones are custom.

neenjaw:
Personally I think we should do it, it shouldn't be too big of a patch.  What I see needing to be done:

- make the `new/2` function a function that accepts an optional keyword list -- [example](http://elixir-recipes.github.io/functions/option-parameters-with-keyword-lists/)
- remove the default values from the struct
- remove the default test case
- add the "new" test cases from the canonical
- add some "new" additional graphical test cases?

Cohen-Carlisle:
Let us know how things go @mexicat! If you need help or even to hand it off, we are here! :+1:

mexicat:
@neenjaw @Cohen-Carlisle something like this? (Rough draft)

```elixir
def new(opts \\ [])

def new(white: same, black: same),
  do: raise(ArgumentError)

def new(opts) do
  white = opts |> Keyword.get(:white) |> check_range()
  black = opts |> Keyword.get(:black) |> check_range()

  %Queens{white: white, black: black}
end

defp check_range({x, y})
     when x not in @board_range
     when y not in @board_range,
     do: raise(ArgumentError)

defp check_range(queen), do: queen
```

So it can be called like `Queens.new(white: {0, 1}, black: {1, 2})` and both parameters are optional. (I added the `check_range/1` function because the new spec tests verify that all positions are between `0..7`).

neenjaw:
Push a commit with the changes, it's easier to review your code meaningfully.

I like some things about your solution, there are things I would change but I want to be able to comment on the file. :)

mexicat:
I added the changes to the example file, let me know what you think.

exercises/queen-attack/example.exs
```diff
@@ -1,20 +1,24 @@
...
+  defstruct [:white, :black]
+  @board_range 0..7
 
   @doc """
   Creates a new set of Queens
   """
-  @spec new() :: Queens.t()
-  @spec new({integer, integer}, {integer, integer}) :: Queens.t()
-  def new(same, same), do: raise(ArgumentError)
+  @spec new(white: {integer, integer} | nil, black: {integer, integer} | nil) :: Queens.t()
```
neenjaw:
```suggestion
  @spec new(Keyword.t()) :: Queens.t()
```

exercises/queen-attack/example.exs
```diff
@@ -1,20 +1,24 @@
...
+  def new(opts \\ [])
+
+  def new(white: same, black: same),
+    do: raise(ArgumentError)
```
neenjaw: I don't think this will ever match, as the keywords will be collected into a single keyword list.

exercises/queen-attack/example.exs
```diff
@@ -1,20 +1,24 @@
...
+
+  def new(opts) do
+    white = opts |> Keyword.get(:white) |> check_range()
```
neenjaw: Maybe can implement `check_range()` as a guard ~~> `is_valid_position()`, similar to the RobotSimulator exercise

exercises/queen-attack/example.exs
```diff
@@ -1,20 +1,24 @@
...
+    black = opts |> Keyword.get(:black) |> check_range()
```
neenjaw: Then you can have more of the structure (in pseudocode):

1. Collect queen positions from keyword list
2. Make sure they are valid
3. Create blank board
3. Add each queen to the board, when added, they check for the other queen already at that position 
    - this might be an addition private or public function

exercises/queen-attack/example.exs
```diff
@@ -38,6 +42,13 @@ defmodule Queens do
     white_x == black_x || white_y == black_y || diagonal?(white, black)
   end
 
+  defp check_range({x, y})
+       when x not in @board_range
+       when y not in @board_range,
+       do: raise(ArgumentError)
+
+  defp check_range(queen), do: queen
+
```
neenjaw: See above, I think this could be implemented as a guard more concisely

exercises/queen-attack/example.exs
mexicat: Yeah it doesn't work, I will implement this check another way.

exercises/queen-attack/example.exs
mexicat: Not sure how to implement a guard against a function's keyword options since I can't pattern match against them (as both keys would need to be present) to extract the parameters for the guard. Or do you mean to first `Keyword.get` the options and then use the guard in a `for`/`with`?

neenjaw: I mean a function which can type check that the coordinate position is valid and return a strict boolean response, it doesn't have to be placed in a `when` guard clause.

exercises/queen-attack/example.exs
mexicat: Could this be a good fit for a `with` pipeline or would it be too messy?

neenjaw: I find with constructs to be really useful if you have lots of potential branches (errors) to catch.  I'm not sure this is complex enough to really benefit, but it's worth the experiment if you want to try it!

mexicat:
In the end I decided a `with` wasn't worth it. This new refactor seems a bit more clean to me, all new functions are short and simple.

exercises/queen-attack/example.exs
```diff
@@ -38,6 +39,15 @@ defmodule Queens do
     white_x == black_x || white_y == black_y || diagonal?(white, black)
   end
 
+  defp is_valid_position?({x, y}) do
+    x in @board_range && y in @board_range
+  end
```
neenjaw: This should either be made into a guard function:

```elixir
defguardp is_coodinate(c)
  when is_tuple(c)
  and tuple_size(c) == 2
  and c |> elem(0) |> is_integer()
  and c |> elem(1) |> is_integer()

defguardp is_board_position(p)
  when is_coordinate(p)
  and p |> elem(0) |> Kernel.in(@board_range)
  and p |> elem(1) |> Kernel.in(@board_range)
```

or be renamed without the `is_` prefix.  I am inclined towards the former just because I think guards are under expressed in the exercises.

exercises/queen-attack/example.exs
```diff
@@ -1,19 +1,20 @@
...
-  @spec new({integer, integer}, {integer, integer}) :: Queens.t()
-  def new(same, same), do: raise(ArgumentError)
+  @spec new(Keyword.t()) :: Queens.t()
+  def new(opts \\ []), do: do_new(opts, %Queens{})
 
-  def new(white, black) do
-    %Queens{white: white, black: black}
+  defp do_new([{queen, pos} | rest], queens) do
+    if not is_valid_position?(pos), do: raise(ArgumentError)
+    do_new(rest, add_to_queens(queens, {queen, pos}))
```
neenjaw: Could this be handled in an `Enum.reduce` expression?

exercises/queen-attack/example.exs
```diff
@@ -1,19 +1,20 @@
...
+  @board_range 0..7
 
   @doc """
   Creates a new set of Queens
   """
-  @spec new() :: Queens.t()
-  @spec new({integer, integer}, {integer, integer}) :: Queens.t()
-  def new(same, same), do: raise(ArgumentError)
+  @spec new(Keyword.t()) :: Queens.t()
+  def new(opts \\ []), do: do_new(opts, %Queens{})
```
neenjaw:
```suggestion
  def new(opts \\ []), do: do_new(%Queens{}, opts)
```
The reversed arguments would be more normal to see

exercises/queen-attack/example.exs
```diff
@@ -38,6 +39,15 @@ defmodule Queens do
...
   end
 
+  defp is_valid_position?({x, y}) do
+    x in @board_range && y in @board_range
+  end
+
+  defp add_to_queens(queens, {queen, pos}) do
+    if pos in Map.values(queens), do: raise(ArgumentError)
+    %{queens | queen => pos}
+  end
```
neenjaw: I am unsure how I feel about any duplicate raising an error.  I think what I'd rather see is is a check for each queen

```elixir
case queen do
  :black -> if queens.white == pos, do: raise ArgumentError, "white queen already placed here"
  :white -> if queens.black == pos, do: raise ArugmentError, "black queen already placed here"
  _ -> raise ArguementError, "invalid queen color"
end
```

I'm not sure if it matters if they try to redefine themselves.

exercises/queen-attack/example.exs
neenjaw: and just because it is defined as a guard doesn't mean we can't use it as a regular function, it just does additional compiler checks and ensures that it doesn't have any side effects.

exercises/queen-attack/test/queen_attack_test.exs
```diff
@@ -1,180 +1,212 @@
...
   end
 
   @tag :pending
-  test "can attack on other diagonal" do
-    queens = Queens.new({0, 6}, {1, 7})
-    assert Queens.can_attack?(queens)
+  test "cannot occupy same space" do
+    assert_raise ArgumentError, fn ->
+      Queens.new(white: {2, 4}, black: {2, 4})
+    end
```
neenjaw: Could you add a test case to test these in reverse? e.g Queens.new(black: pos, white: pos)

exercises/queen-attack/test/queen_attack_test.exs
```diff
@@ -1,180 +1,212 @@
...
   end
```
neenjaw: And maybe a test case for each color as a string representation when there isn't the other color declared.

mexicat:
I updated the tests by dividing them into `describe` blocks, sorting them, renaming them and changing the expected values according to the spec. I removed the default test cases and added the string representation ones at the bottom. Might it be worth it to add some other tests, like ensuring that colors other than `:black` and `:white` raise `ArgumentError`?

exercises/queen-attack/test/queen_attack_test.exs
mexicat: I can do that but I would have to change the implementation because right now `Queens.to_string/1` will fail with only one queen declared.

neenjaw:
Ya, I left two comments on the files for suggesting a couple more test cases.

What happens right now if `can_attack?` is called when only one queen is defined?  From the looks of it, you would receive a pattern match error because one queen in the struct is `nil` could we adjust it to return false?  Or maybe some sort of {:ok, result} {:error, reason} tuple?

exercises/queen-attack/test/queen_attack_test.exs
neenjaw: If you insert at line 70:

```elixir
defp insert_queen(board, nil, _colour), do: board
```

then it won't error

mexicat:
Something like:
```elixir
def can_attack?(%Queens{white: white, black: black})
    when is_nil(white) or is_nil(black),
    do: false
```
?

neenjaw:
That would work

mexicat:
I added some new tests and inserted them into the relevant sections.

Apparently the CI build is failing only on Elixir ≤ 1.7.2 because of `mix format --check-formatted` on my `example.exs`, but there's no additional info... some syntax change between versions, maybe?

neenjaw:
Just run `mix format` on the `example.exs`, that command checks whether or not you have done that already.

mexicat:
I have `mix format` configured on save in VSCode (and running it again via CLI doesn't change anything); if you look at the CI build it's only failing on old versions of Elixir, maybe due to different formatting rules in the past?

Screen Shot 2019-11-12 at 15 16 54

neenjaw:
interesting...

neenjaw:
alright, here's the problem:
If we start with:
```elixir
  defguardp is_coordinate(c)
            when is_tuple(c)
            and tuple_size(c) == 2
            and c |> elem(0) |> is_integer()
            and c |> elem(1) |> is_integer()

  defguardp is_board_position(p)
            when is_coordinate(p)
            and p |> elem(0) |> Kernel.in(@board_range)
            and p |> elem(1) |> Kernel.in(@board_range)
```
mix format in 1.6.6 (and I assume 1.7.2) formats the guards like this:
```elixir
  defguardp is_coordinate(c)
            when is_tuple(c) and tuple_size(c) == 2 and c |> elem(0) |> is_integer() and
                   c |> elem(1) |> is_integer()

  defguardp is_board_position(p)
            when is_coordinate(p) and p |> elem(0) |> Kernel.in(@board_range) and
                   p |> elem(1) |> Kernel.in(@board_range)
```
and in 1.8.2+
```elixir
  defguardp is_coordinate(c)
            when is_tuple(c) and
                   tuple_size(c) == 2 and
                   c |> elem(0) |> is_integer() and
                   c |> elem(1) |> is_integer()

  defguardp is_board_position(p)
            when is_coordinate(p) and
                   p |> elem(0) |> Kernel.in(@board_range) and
                   p |> elem(1) |> Kernel.in(@board_range)
```

> Interestingly, if the file is formatted in the 1.6.6 version, 1.8.2 doesn't attempt to reformat

Which isn't a breaking change other than breaking our CI...

For now, @mexicat is it okay if I push the 1.6.6 version to your branch to see if it builds? (which it should if my tests are correct on my machine).  I will open another issue to look at this problem more.

mexicat:
Sure, go ahead :)

neenjaw:
Hmm, git is acting funny because I think you built this on your master branch

maybe if you change the guards on your file manually to: 
```elixir
  defguardp is_coordinate(c)
            when is_tuple(c) and tuple_size(c) == 2 and c |> elem(0) |> is_integer() and
                   c |> elem(1) |> is_integer()

  defguardp is_board_position(p)
            when is_coordinate(p) and p |> elem(0) |> Kernel.in(@board_range) and
                   p |> elem(1) |> Kernel.in(@board_range)
```

after, if you run `mix format --check-formatted; echo $?` if the exit code is 0, then this should build with your next push

neenjaw:
@Cohen-Carlisle Given that this is a breaking change for old solutions, mind taking a peek

Cohen-Carlisle:
Sorry for the delay, I'll take a look at this tonight.

Cohen-Carlisle:
Hey, thanks for these changes! I only have two thoughts about things to change:
1. It would be nice to either have `new()` take two positional arguments or go fully in on taking an enumerable. What I mean by the latter is to take any number of queen coordinates, like so: `[{1,1},{2,3},{4,6}]`
2. I think since the canonical data did away with `to_string()`, we should do the same.

Now, I also think think what we have here is still very valuable, so I'm not against merging this as is. But if we do, I'll probably follow up with the two changes above.

neenjaw:
As followup, what would an enumerable of >2 coordinate pairs mean if not attributed to a specific color, unless it should mean color shouldn't be counted at all?


### if color matters:

Maybe it could work for there to be a new/0 to return a blank board, an add/2 to add a queen using a color, coordinate , and then a new/1 which then accepts an enumerable compatible with add/2 arguments

### If color doesn't matter

Then a new/1 accepting an enumerable of coordinates makes a lot of sense and simplifies a lot of the complexity

---

I'm hesitant to ask him to do any more revision of we don't provide a clear-cut path here, but I think what you're saying has merit

Cohen-Carlisle:
I think color should not matter. Removing `to_string()` helps facilitate this. But, adding arbitrary amounts of queens makes `can_attack?()` potentially much more complicated. I'm in favor of merging this as is and then think about next steps afterwards.

PR #516, opened 2019-11-08
Rename test file to mirror implementation filename

kronn:
This helps with editor expectations to run the tests quickly, from the implementation-file.


neenjaw:
I'm not sure I understand the context of this PR, is there some difficulty that you were encountering?

From shell: `$ mix test --include pending:true` is the quickest way to run tests in a mix project root directory

NobbZ:
I think I do understand the problem.

Some editors augment some buttons into the source code, to run the tests for the current module. Those though do not use the module name to find the test, but the filename instead. And if the current implementation is `foo.ex` they will run `foo_test.exs`.

Though I do think, the proper way to repair this were to rename the implementation to `word_count.ex` and the module to `WordCount` as well as the test module to `WordCountTest` to create a deterministic relationship between filenames and exercise names.

neenjaw:
Ah, that makes some sense of this to me.

As you say, I think if we need to make a change, it should be to change the modules to `WordCount` and `WordCountTest` and rename the files to `word_count.ex` and `word_count_test.exs`

--

@kronn - would you be able to push another commit with those changes?

neenjaw:
Out of curiosity, what editor might do that? I don't think ElixirLS for vscode does that. An emacs setup?

kronn:
It's vim, not emacs :smile: . And I run tests by pressing `,tf`, which always works from a test-file and also from the implementation if the names match. I should have clarified that in the original PR.

I will update the PR tomorrow. Renaming one file is nice in the web-ui, renaming everything to the name of the exercise - while making more sense - is not something I want to do that way.  :wink:

kronn:
I renamed everything to match the exercise. I intentionally put things into multiple commits to ease review or reverts, feel free to squash them if you prefer fewer commits.

It was an interesting exercise to actually look into all files, normally, I just open the exercise and run the tests from the editor or with `mix test`. At some point I will have to learn all that plumbing as well. For now it was interesting and I am happy that you provide those files in the correct way.

neenjaw:
Nice work @kronn! Thanks for this contribution!

PR #518, opened 2019-12-02
A test to ensure solution can handle multiple accounts.

nilvon9wo:
(Credit to neenjaw who authored this test when I pointed out that this requirement was missing, which made the inclusion of the "account" variable superfluous.)


neenjaw:
Thanks @nilvon9wo for submitting the PR!

PR #519, opened 2019-12-08
Update tests - binary, binary-search, binary-search-tree, bob, bowling

neenjaw:
Summary of changes (see each commit for detailed changes):
- tmp-exercises added to .gitignore, testing locally can sometimes make this junk directory and if the tests are aborted, it persists.  This is added to prevent accidental commits.
- binary tests updated to canonical, combined with current
- binary-search tests updated to canonical, combined with current
- binary-search-tree tests updated to canonical, combined with current
- bob tests updated to canonical, combined with current
- bowling tests update to canonical, example solution now compliant with test suite

---

The big thing in this commit is that the example bowling solution failed the canonical tests examples.  I have detailed it in the commit summary, but in short:
- The example uses a two-roll-per-frame design, where a strike fills the frame and advances the counter
- The last frame, if there are 3 strikes, takes up 3 frames in this.
- In some cases (the last frame had strikes, or a spare then any combination), it would not signal the end of the game appropriately (therefore test cases that failed)
- I added the missing test cases to preserve the design but now the solution passes the tests

I tested it rigorously using a jig, and it appears to work more extensively than tested by the test suite.  You can see the function/testing output at [my gist: here](https://gist.github.com/neenjaw/2f917a802edb98da450e0638fc9da98d)


exercises/bob/test/bob_test.exs
```diff
@@ -60,16 +85,56 @@ defmodule BobTest do
...
+  end
+
+  @tag :pending
+  test "starting with whitespace" do
+    assert Bob.hey("         hmmmmmmm...") == "Whatever."
+  end
+
+  @tag :pending
+  test "ending with whitespace" do
+    assert Bob.hey("Okay if like my  spacebar  quite a bit?   ") == "Whatever."
```
Cohen-Carlisle: I was doing some quick spot-checking and it looks like this [should be returning](https://github.com/exercism/problem-specifications/blob/master/exercises/bob/canonical-data.json#L185-L187) `"Sure."` instead of `"Whatever."`.

Cohen-Carlisle: It would be nice to have some automated generators that consume `canonical-data.json`s so that this becomes a non-problem. That's probably beyond the scope of this PR.

neenjaw: Hmm... I can fix that, it's likely a copy/paste error. It is surprising that it passed the tests.

It would be great to have a generator, I also think it's beyond the scope of this PR. :)

neenjaw: This is now fixed. Thanks for catching it.

neenjaw:
Awesome, thanks for looking over it.

PR #522, opened 2019-12-24
Fix student name order

nathanielknight:
The order of student names in one of the Kindergarten Garden tests is reversed, which results in a spurious test failure.

### DIFF
--- a/exercises/kindergarten-garden/test/garden_test.exs
+++ b/exercises/kindergarten-garden/test/garden_test.exs
@@ -43,7 +43,7 @@ defmodule GardenTest do
 
   @tag :pending
   test "accepts custom child names" do
-    garden_info = Garden.info("VC\nRC", [:nate, :maggie])
+    garden_info = Garden.info("VC\nRC", [:maggie, :nate])
     assert garden_info.maggie == {:violets, :clover, :radishes, :clover}
     assert garden_info.nate == {}
   end
--- a/exercises/kindergarten-garden/test/garden_test.exs
+++ b/exercises/kindergarten-garden/test/garden_test.exs
@@ -53,7 +53,6 @@ defmodule GardenTest do
     names = [
       :maggie,
       :nate,
-      :xander,
       :ophelia,
       :pete,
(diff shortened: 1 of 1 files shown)


neenjaw:
I don't think this should be changed. The problem supposes that the garden is ordered alphabetically, so when providing custom names, this test case makes sure that solutions are ordering the names before assigning the plants to the students.

Have you encountered this error with students you are mentoring or have you discussed this with a mentor reviewing your solution?

devonestes:
Thanks for opening this PR!

Yeah, this is intentional. If the names are always alphabetical in the input then folks don't even need to consider the alphabetization of the garden and can just slice it.

Maybe, if folks think it might be a bug that the names aren't alphabetized, maybe they should be even _more_ out of order so it's abundantly clear that the learner needs to do the alphabetization themselves?

neenjaw:
Would an additional test case featuring 3 names out of order make this clearer? Or does this require more explicitness in the problem statement?

nathanielknight:
Ah! I understand now. Yes, I misunderstood the problem, and assumed that the given order of names was significant. A small problem that calls that out in the test suite might help ward off folks who misread the problem and then show up thinking they're being helpful 😕 

Something like:

```elixir
  test "accepts custom child names" do
    garden_info = Garden.info("VC\nRC", [:maggie, :nate])
    assert garden_info.maggie == {:violets, :clover, :radishes, :clover}
    assert garden_info.nate == {}
  end

  test "assigns plants in alphabetical order" do
    garden_info = Garden.info("VC\nRC", [:samson, :ahmed])
    assert garden_info.ahmed == {:violets, :clover, :radishes, :clover}
    assert gaden_info.samson == {}
  end
```

neenjaw:
@nathanielknight - if you'd be willing to submit a PR with that new test case referencing this PR discussion, it would be a good addition. Or reopen this one with a PR to amend. :)

nathanielknight:
New PR submitted. Thanks for the help and feedback folks :slightly_smiling_face:

PR #523, opened 2019-12-24
Split out Kindergarten Garden's order requirement

nathanielknight:
This commit adds a test to the Kindergarten Garden exercise that
explicitly calls out that arguments should be processed in alphabetical
order instead of in the order they're given.

Before this commit, this requirement was checked in the first test where
a non-default value for the argument in question was provided. For
learners who haven't read the specification carefully, it's easy to get
the wrong idea about why that test is failing. An explicit reference to
that requirement in the test suite gives those folks a second chance to
understand it properly.

This commit is a follow-up to #522


neenjaw:
Thanks!!

PR #528, opened 2020-01-17
Fix dialyzer check and cache PLTs

devonestes:
This will cache the Erlang & Elixir PLTs for each combination, and also
fixes the dialyzer run for us to make sure we're not referencing any
invalid types or anything else like that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ bin/configlet
 bin/configlet*
 deps
 _build
+/priv/plts/*.plt
+/priv/plts/*.plt.hash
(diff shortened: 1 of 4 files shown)


filalex77:
Related: has it been considered to use something like https://dependabot.com/elixir/ for this repo? Thought about it after noticing the dialyxir version bump.

devonestes:
@filalex77 I don't think we really need dependabot. We only have one real dependency, and it isn't even user facing. I'm not opposed to it, but I don't think we'd get a lot of value out of it.

filalex77:
@devonestes Sure, reasonable.

neenjaw:
@devonestes Thanks for taking this on, I really appreciate it as my experience with dialyzer is limited.

PR #532, opened 2020-04-12
ci update

Cohen-Carlisle:
ci update to fix some issues 

* fix trying to build nonexistent elixir v1.6.6-otp-22
  * instead of a matrix, specify elixir and otp versions together
    * now running elixir 1.{6..10}.0 with minumum otp versions
      * fix the following so elixir 1.6.0 / otp 19.0 pass in ci:
        * only use ascii chars in test names
        * fix a formatting issue to appease 1.6.0 formatter
      * skip dialyzer for elixir 1.7 as it times out after 30 mins
* update fetch-configlet to be case insensitive on http headers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,18 @@
 language: elixir
-elixir:
-  - 1.6.6
-  - 1.7.2
-  - 1.8.2
-  - 1.9.0
 
-otp_release:
-  - 21.3
-  - 22.0
-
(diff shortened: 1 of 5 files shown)


neenjaw:
Thanks for looking at this issue, @Cohen-Carlisle!

neenjaw:
LGTM!!

PR #533, opened 2020-04-14
Fixed triangle inequality explications

ryanzidago:
The previously linked [Wikipedia page](https://en.wikipedia.org/wiki/Triangle_inequality) that was cited by Exercism actually contained a false explication of what a triangle is: 
> In mathematics, the triangle inequality states that for any triangle, the sum of the lengths of any two sides must be greater than **_or equal to the length_** of the remaining side.

If you actually take a look at the [first source](https://mathworld.wolfram.com/TriangleInequality.html) cited by the Wikipedia article itself, it has a different definition; which is the proper definition:
>  Geometrically, the right-hand part of the triangle inequality states that the sum of the lengths of any two sides of a triangle is greater than the length of the remaining side. 

One can quickly check on [this website](https://www.mathwarehouse.com/triangle-calculator/online.php), if one can build a triangle with the following values: (a = 2, b = 4, c = 2); which is indeed, not possible since a + c is not greater than b.

Hence, 
- I updated the explication of what a triangle is
- replaced the cited Wikipedia article by the correct source

### DIFF
--- a/exercises/ocr-numbers/lib/ocr_numbers.ex
+++ b/exercises/ocr-numbers/lib/ocr_numbers.ex
@@ -3,7 +3,7 @@ defmodule OcrNumbers do
   Given a 3 x 4 grid of pipes, underscores, and spaces, determine which number is represented, or
   whether it is garbled.
   """
-  @spec convert([String.t()]) :: String.t()
+  @spec convert([String.t()]) :: {:ok, String.t()} | {:error, charlist()}
   def convert(input) do
   end
 end


angelikatyborska:
Hi Ryan! Thank you for your PR. I am incredibly sorry that it took us so long to get back to you. In the chaos caused by the global pandemic and most of the Exercism's community focused on preparing V3, we must have forgotten about your PR 😞.

I understand your doubts about the inclusion of equality in the triangle's definition. It seems like it's a disputed issue in the mathematical world. At the time of writing my comment, the second sentence in that Wikipedia article explains:

> This statement permits the inclusion of degenerate triangles, but some authors, especially those writing about elementary geometry, will exclude this possibility, thus leaving out the possibility of equality.[3]

It's important to note that it is not Elixir track's own decision to define the triangle this way. All exercises are standardized between all tracks, in a repository called [problem-specifications](https://github.com/exercism/problem-specifications). That's where the description of this exercise comes from: [exercises/triangle/description.md](https://github.com/exercism/problem-specifications/blob/master/exercises/triangle/description.md).

It looks to me like this issue was already discussed on the organization level because I found this comment:
https://github.com/exercism/problem-specifications/blob/1a7290b76b2603fb7257b0faf831604b5be16328/exercises/triangle/canonical-data.json#L3-L18

The Elixir track shouldn't change the exercise description without the `problem-specifications` repository changing it first, but based on that comment, it looks like we should just delete this test case and let students solve the exercise with or without allowing for degenerate triangles, whichever they prefer.

https://github.com/exercism/elixir/blob/51dd9b7d95b064d227416154e3baadc9cc742b62/exercises/triangle/test/triangle_test.exs#L69-L72

ryanzidago:
@angelikatyborska thanks for coming back to me regarding this PR :pray: 

I would also argue that it make sense to delete the test that you mentioned. It will be less confusing :wink:.

angelikatyborska:
> I would also argue that it make sense to delete the test that you mentioned. It will be less confusing 😉.

@ryanzidago Great! Would you like to open a pull request with that change or should I do it?

ryanzidago:
@angelikatyborska I closed this PR and opened up a new one here: #550

PR #529, opened 2020-01-17
Adds fireproofsocks as maintainer

fireproofsocks:

PR #536, opened 2020-05-18
Add angelikatyborska as maintainer

angelikatyborska:


config/maintainers.json
```diff
@@ -80,6 +80,16 @@
       "link_url": null,
       "avatar_url": "https://s.gravatar.com/avatar/8fd910b4f7c7baabc22976dc6b29b744",
       "bio": "Elixir developer based in Los Angeles, former instructor at USC"
+    },
+    {
+      "github_username": "angelikatyborska",
+      "alumnus": false,
+      "show_on_website": false,
```
angelikatyborska, line 87: I actually like putting my name on nice things, but I saw that on https://github.com/exercism/elixir/pull/529 fireproofsocks was asked to change theirs to `false`, so I took that as a hint that it's not appropriate to show up on the v2 track if what I am doing is working on v3.

neenjaw: I think that was done with many of the new maintainers because they were brought in on the wave following that initial call for maintainers from the mentors because many of them were relatively new and unknown. 

I think there is a good case for yours to be set to true because you have been on exercism as a mentor for a while, have previously contributed to this track. 

I would support it if there is agreement to it. @ihid / @ erikshierboom?

neenjaw:
I'll leave it to you, can either wait for a response for the above comment or merge now and update it later. :)

angelikatyborska:
If the wait for a response is too long, it can be merged as it is and I will set myself a reminder to change `show_on_website` to `true` in a few months, to enjoy the glory after v3 starts :)

iHiD:
Set it to true for sure!

PR #537, opened 2020-07-04
Add missing nil return value

ChristianTovar:
Hey everyone, do you think adding that `nil` on the `@spec` as a return value is accurate? After all, the exercise description clearly states that returns a sequence of numbers (list) or a `nil`.


neenjaw:
Hey @ChristianTovar! I'm not familiar with this wording exactly, let me take a look at it and probably approve it!  Thanks for submitting this PR!

neenjaw:
Looks good to me, thanks!

ChristianTovar:
Thanks for checking @neenjaw

PR #538, opened 2020-09-25
fix the sorting of map keys

herminiotorres:
the map always be alphabetically sorted.

ex:
```
iex> %{"ability" => "a", "aardvark" => "a", "ballast" => "b", "beauty" =>"b" }
%{"aardvark" => "a", "ability" => "a", "ballast" => "b", "beauty" => "b"}
```

### DIFF
--- a/exercises/etl/lib/etl.ex
+++ b/exercises/etl/lib/etl.ex
@@ -5,7 +5,7 @@ defmodule ETL do
   ## Examples
 
   iex> ETL.transform(%{"a" => ["ABILITY", "AARDVARK"], "b" => ["BALLAST", "BEAUTY"]})
-  %{"ability" => "a", "aardvark" => "a", "ballast" => "b", "beauty" =>"b"}
+  %{"aardvark" => "a", "ability" => "a", "ballast" => "b", "beauty" => "b"}
   """
   @spec transform(map) :: map
   def transform(input) do


neenjaw:
Hello @herminiotorres,

Thanks for submitting this PR.  Can you give me a bit more context around the motivation for this PR?

I ask because maps make no guarantee about key order in Elixir.  Under the hood of the BEAM VM, maps are implemented as `sorted keyword lists` until they hit ~32 entries, then it is flipped to a `hash array mapped trie` type of structure. [See here](https://stackoverflow.com/a/40408469/8583288).

So yes, they often appear ordered, but that is coincidence due to the underlying implementation rather than a design guarantee.

Cohen-Carlisle:
I do feel like although elixir maps are unordered, it makes sense to show in the examples what the elixir repl will show (currently). It doesn't seem like something likely to change and may reduce confusion. Also, [José made a similar change to elixir docs](https://github.com/elixir-lang/elixir/commit/72ebd9bfc7dd322bf025b1263c26f06793d8b5ff) when [I was a bit confused](https://github.com/elixir-lang/elixir/issues/6759) over map ordering.

neenjaw:
:+1: Sounds good. I was also interested in the motivation behind it as it is a common misconception that maps have ordered keys.  I find this often happens when discussing the roman numerals problem with students, but it would be nice to find new ways of communicating this, but perhaps that is what v3 is for.

herminiotorres:
hello @neenjaw 

My main motivation its when I pick this map and put in my `iex` I see organize different then represent in the documentation.

So based on the Elixir doc:
> Key-value pairs in a map do not follow any order (that's why the printed map in the example above has a different order than the map that was created).

When you use a mixed data type for your keys in your map, some data type has more weight than other to sorted your map.
But when you choose one kind of data type like string, the map will be keeping sort your keys.

```elixir
iex> %{1 => "one", "b" => "two", 2 => "three", "a" => "four",  d: 1, e: 2} 
%{1 => "one", 2 => "three", :d => 1, :e => 2, "a" => "four", "b" => "two"}
```

neenjaw (approved):

neenjaw:
@herminiotorres, even with different data types used for each key, it is still sorted, but just sorted according to term order (https://hexdocs.pm/elixir/operators.html#term-ordering). Bitstrings (which is a superset of binaries [strings]) just happens to be considered greater than the others.

Sounds good, let's make the change.

:tada: Thanks for the PR!

PR #539, opened 2020-09-25
fix doc the IsbnVerifier

herminiotorres:


neenjaw:
Just looking at this, I wonder if this module should be changed to ISBNVerifier rather than changing the docs.

Cohen-Carlisle:
I think this is the correct change. Initialisms and acronyms usually just get their first letter capitalized in the rest lower case. This is nice since a programmatic case changer can't tell the difference between them and real words.

herminiotorres:
good point @Cohen-Carlisle

neenjaw:
Sounds good, I think it was the other way before we set up each exercise as a mix project. 

I don't know why TravisCI failed with the configlet. Is there a way to re-trigger it?

Cohen-Carlisle:
Hmm, I'm not sure, to be honest. if it can't be rerun I think it's fine to merge this as is since it's documentation only and wouldn't have caused the build to fail.

herminiotorres:
I could try to close and reopen the pull-request and hope the CI trigger again

Cohen-Carlisle:
I'll just merge! :wink:

PR #545, opened 2020-10-30
Github Actions Support

jwarwick:
First pass migration from Travis to Github Actions. 

Almost certainly will need fixes, since it's difficult to test until the basic action has been added to the main branch. At that point, it can be triggered manually (with `workflow_dispatch`).

I'm not running `mix format` and `dialyzer` on all Elixir/OTP combinations (as I believe Travis was). I think I have my head around this now, so changes should be easier going forward.

Thanks to SleeplessByte for guidance.

Addresses #543


neenjaw:
Do you still not have write privileges? I'll make a branch for this, then let's merge it to there to be able to test it

neenjaw:
I know @iHiD has been slammed this week, so i think that has just been missed

neenjaw:
there is a branch now `gha-migration` which you can target, then let's merge it and start testing it out.  If it works then we'l just PR to master

.github/workflows/ci.yml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Find trailing whitespace
+        uses: harupy/find-trailing-whitespace@master
+
+  ci:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        elixir: ['1.6.0', '1.7.0']
+        otp: ['19.0']
```
neenjaw, line 46: These generate a matrix right? then 47-55 are the specific additions since the earlier versions don't support those versions of OTP, correct?

bin/dialyzer_check.sh
```diff
@@ -1,9 +1,13 @@
...
 # currently dialyzer is timing out after 30 mins only on elixir 1.7 / otp 19
 if elixir --version | grep -q 'Elixir 1.7'; then
   echo "skipping dialyzer for Elixir 1.7"
   exit 0
 fi
 
+mkdir -p ./priv/plts
+
 mkdir -p ./tmp/src
 mkdir ./tmp/build
```
neenjaw, line 13: This may require some sort of caching step in our action, due to the slowness of dialyzer

neenjaw (approved):
This is really good @jwarwick!!! Just two comments, but nothing wrong with this so I will merge

neenjaw:
I merged #546

.github/workflows/ci.yml
jwarwick: Correct. Looks like Travis specified things a bit different, so I just used the same combinations. Might be good to decide what our baseline is.

bin/dialyzer_check.sh
jwarwick: Agreed. GHA has some caching actions would should be applicable. Looks like a 5 minute runtime at the moment (all of it in generating the plt).

neenjaw: There has been discussion prior whether or not we should even do a dialyzer check.  Perhaps now is the time to ditch it? @NobbZ @Cohen-Carlisle @devonestes

neenjaw:
@saschamann

neenjaw:
@exercism/maintainers-admin this is ready once @saschamann gives it a once over

bin/dialyzer_check.sh
neenjaw: PLT caching makes this insignificant,  nvm!

.github/workflows/pr.ci.yml
```diff
@@ -0,0 +1,114 @@
...
+        run: |
+          mix local.rebar --force
+          mix local.hex --force
+          mix deps.get
+
+      - name: Build Project
+        run: mix
+
+      - name: Find Trailing Whitespace
+        uses: harupy/find-trailing-whitespace@master
```
SaschaMann, line 43: This is a third party action, so you should pin it to a commit hash rather than `@master` or a tag for security reasons: https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions

It's somewhat tedious to keep commit hashes up-to-date, so I'd personalyl remove this third party action and find another solution to deal with trailing whitespace or not bother with it at all.

TL;DR: If that actions is compromised, a malicious actor could gain write access to the repo through the GITHUB_TOKEN.

SaschaMann (commented):
Didn't check the bash scripts.

Other than the security issue, it looks good to me.

.github/workflows/pr.ci.yml
jwarwick: Thanks. I moved the whitespace checking into the bash file. Using the same mechanism we previously had in Travis.

neenjaw (approved):

SaschaMann (approved):

iHiD:
> I know @iHiD has been slammed this week, so i think that has just been missed

The webhook that should do this automatically is broken.

iHiD (approved):

neenjaw:
Great work @jwarwick!!!! :tada: :dancers: :tada:

Issue #554, opened 2021-01-18
poker: tests are hard to pass with strict order


MeerKatDev:
I just did the Poker exercise, and one main point of my strategy was to sort the cards from 1 to A just after the input ([my solution](https://exercism.io/tracks/elixir/exercises/poker/solutions/acc983b5aed14ff9a5521c3c0e2b4445)). When I was trying to pass the tests, it hit me that I had a lot of tests not passing because they are checking on direct equality between two lists, i.e.:
```
Poker.best_hand(hand) == ~(4H 3S 5S 6S AH)
```
which checks that the returned hand is exactly in the order in which the cards were dealt. 

I think this is really confusing, since in Poker the order with which you get the cards don't matter (to my knowledge), and moreover it's written nowhere in the exercise that the cards need to be in the same order as they were given in the input, which I think would also add a unnecessary complication to the exercise.

I saw that the Allergies exercise has a utility in its tests that checks for equality of a set, not of a list, which would suit this case well already, so I'd propose to change the tests in Poker to follow that example.

neenjaw:
I'd like to hear from other maintainers what their experience has been of this problem, but I have been thinking about this since you brought it up yesterday in slack. 

If you have time to create a PR, adding a wait to compare the hands so that the order doesn't matter (like the allergies exercise), I am open to it and would probably have time to review it. 

If you can include and estimate of how much work it is and how much it would take to maintain, that would be nice.

MeerKatDev:
I think 30 minutes top, it's only about changing the `==` operators, I suppose.
It would be good to create a note somewhere that there is such a comparison method in some exercises (maybe Allergies is not the only one?), so that we would be able to maintain them together in case of changes.

I'm not sure what you mean by 

> how much it would take to maintain

 , but I suppose that changes would happen when something in the Elixir API will change. I think it's a remote possibility, in this case, so not much work.

neenjaw:
I think I mean that everything has a maintenance cost, but I think you're right in this case once the comparator works, there is probably little maintenance required.

PR #555, opened 2021-01-20
change to poker tests in ref to issue #554

MeerKatDev:
@neenjaw could you have a look?

### DIFF
--- a/exercises/poker/test/poker_test.exs
+++ b/exercises/poker/test/poker_test.exs
@@ -1,10 +1,25 @@
 defmodule PokerTest do
   use ExUnit.Case
 
+  defp assert_is_a_set_containing(lists, to_contain) do
+    sets = Enum.map(lists, &Enum.into(&1, MapSet.new()))
+
+    same_contents = Enum.map(to_contain, &Enum.into(&1, MapSet.new()))
+
+    are_they_eq =
+      Enum.zip(sets, same_contents)
+      |> Enum.all?(fn {l1, l2} -> MapSet.equal?(l1, l2) end)
+
+    assert are_they_eq,
+           "Expected a set with: #{inspect(to_contain)} got #{
+             inspect(Enum.map(sets, &MapSet.to_list(&1)))
+           }"
+  end
+
   # @tag :pending
   test "single hand always wins" do
     high_of_jack = ~w(4S 5S 7H 8D JC)
-    assert Poker.best_hand([high_of_jack]) == [high_of_jack]
+    assert Poker.best_hand([high_of_jack]) |> assert_is_a_set_containing([high_of_jack])
   end
 
   @tag :pending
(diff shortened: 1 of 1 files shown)


neenjaw:
I'll take a look after work today. 👀

neenjaw:
❤️ Thanks for opening this PR, what if we did something simpler so that we can leverage ExUnit's diff on failure?

```elixir
  defp assert_poker(dealt_hands, expected_winning_hands) do
    winnings_hands = dealt_hands |> Poker.best_hand() |> Enum.map(&Enum.sort/1) |> Enum.sort()
    expected_winning_hands = expected_winning_hands |> Enum.map(&Enum.sort/1) |> Enum.sort()

    assert winnings_hands == expected_winning_hands
  end

  # @tag :pending
  test "single hand always wins" do
    high_of_jack = ~w(4S 5S 7H 8D JC)
    assert_poker([high_of_jack], [high_of_jack])
  end
```

Then if this fails it gives us:

```
 1) test single hand always wins (PokerTest)
     test/poker_test.exs:12
     Assertion with == failed
     code:  assert winnings_hands == expected_winning_hands
     left:  [["4S", "5S", "7H", "8D", "JC"]]
     right: [["4S", "5S", "7H", "8D", "KC"]]
     stacktrace:
       test/poker_test.exs:14: (test)
```

Thoughts?

MeerKatDev:
That doesn't sort the the letter-cards (A, J, Q, K) and the diff works anyway in the code I posted (manually, of course).
To use `Enum.sort` you have to implement some more complicated custom sorting for the above reason - I did implement it while doing the exercise and it's much more prone to error than this MapSet.equal strategy.
Moreover, this function I put in the PR is already used in another exercise.. so nothing new under the sun

MeerKatDev:
On a second thought, that might work as well, because whatever order you use, it's used for both lists.
I still prefer the first version because it's more adequate to the case at hand (the order shouldn't matter), but as you prefer.

neenjaw:
Like your original issue said, the order of the cards in your hands isn't specified by the problem spec, so like your second thought here, I am also in agreement that the card order doesn't matter.  What I think is important is the color-coded diff which isn't available in your function (or the function implemented by the `allergies` problem).

Your assertion function also requires the order of the expected hands to be in a specific order for the comparison to work, for ex: If you reverse the expected hands on line 45, the test fails.

https://github.com/exercism/elixir/blob/b5d5597793707e2f0fcb993de33867237f6daf6e/exercises/poker/test/poker_test.exs#L45

So If we are going to implement a function that compares the equality of the returned vs the expected agnostic to the order of the cards in the hand, it should also be able to compare the list of hands agnostic to the order of the list returned.

neenjaw:
I think it's also important to point out that while many of the examples are examples of good code, some aren't perfect and there is room for improvement / iteration in many exercises.  Which underscores is why I think we should advance this exercise's test suite to the next level, like you suggested in the issue, to create a better experience for students.

In the case of this exercise, it keeps the focus on the problem of determining poker hand winners, not on ordering a list which has no external meaning.

MeerKatDev:
Feel free to change this PR with the code you proposed, then :)

neenjaw:
Closed due to merge conflicts, PR recreated with new repo structure at #585

PR #562, opened 2021-01-29
Crypto square: Keep trailing spaces of the chunks

ggpasqualino:
According to the [specification](https://github.com/exercism/problem-specifications/blob/c76d358b75c32468b8a6541484c50f8144e91beb/exercises/crypto-square/canonical-data.json#L59-L65), the trailing spaces of each chunk must be kept in the output.

I was helping a student in my company and he was having the tests failing because of whitespaces. We realized that the tests were not following the specification of the problem.

### DIFF
--- a/exercises/crypto-square/example.exs
+++ b/exercises/crypto-square/example.exs
@@ -15,7 +15,7 @@ defmodule CryptoSquare do
 
     normalized
     |> String.graphemes()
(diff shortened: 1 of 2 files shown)


angelikatyborska:
You are totally right! Thank you for the fix 💜 

PS: One of the CI checks is failing because of the branch renaming that has been happening exercism-wide yesterday. I'll fix it later today.

angelikatyborska (approved):

PR #575, opened 2021-01-29
[v3] Post big-bang fixes

angelikatyborska:
Big bang PR: https://github.com/exercism/elixir/pull/563

- Adds Elixir 1.11 with OTP 23.2 as the main version pair to run the checks on
- Renames example solutions for practice exrcises from `/example.exs` to `/.meta/example.ex`.
- Fixes the `test_exercises.sh` script to test _all_ exercises
- Fixes the `dialyzer_check.sh` script to check _all_ exercises
- Adds a note for the `secrets` exercise that in Elixir <= 1.9, the Bitwise module must be required before it can be used.


bin/test_exercises.sh
```diff
@@ -66,7 +74,7 @@ do
     if [ "$compile_exit_code" -eq 0 ]
     then
       # perform unit tests
-      test_results=$(mix test --no-compile --color --no-elixir-version-check --include pending 2> /dev/null)
```
angelikatyborska, line 69: Running the test with `--no-compile` makes tests fail for bread-and-potions which requires that protocols are consolidated.

angelikatyborska:
@neenjaw do you have any idea about this error from the `elixir / pr / precheck (pull_request)` check?
```
:dialyzer.run error: Old PLT file /home/runner/work/elixir/elixir/priv/plts/eventstore.plt
```

exercises/concept/secrets/.docs/introduction.md
```diff
@@ -65,3 +65,5 @@ Bitwise.<<<(1, 2)
 ```
 
 All bitwise functions only work on integers.
+
+If you are running Elixir version 1.9 or lower, you will need to call `require Bitwise` at the beginning of the module definition to be able to use the _Bitwise_ module.
```
angelikatyborska, line 69: I want people solving this locally on an older Elixir version to still be able to solve it, but I don't want to explain at this point what's `require`, what's `use` and the difference between functions and macros.

bin/dialyzer_check.sh
```diff
@@ -11,14 +11,21 @@ mkdir -p ./priv/plts
 mkdir -p ./tmp/src
 mkdir ./tmp/build
 
-for example_file in exercises/**/example.exs
+for example_file in exercises/concept/**/mix.exs
 do
   exercise_dir=$(dirname $example_file)
   exercise=$(basename $exercise_dir)
-  cp "$exercise_dir/example.exs" "./tmp/src/$exercise.exs"
+  cp "$exercise_dir/.meta/exemplar.ex" "./tmp/src/$exercise.ex"
```
neenjaw, line 18:
```suggestion
  cp "${exercise_dir}/.meta/exemplar.ex" "./tmp/src/${exercise}.ex"
```

bin/dialyzer_check.sh
```diff
@@ -11,14 +11,21 @@ mkdir -p ./priv/plts
...
 done
 
-elixirc -o ./_build ./tmp/src/*.exs
+for example_file in exercises/practice/**/mix.exs
+do
+  exercise_dir=$(dirname $example_file)
+  exercise=$(basename $exercise_dir)
+  cp "$exercise_dir/.meta/example.ex" "./tmp/src/$exercise.ex"
```
neenjaw, line 25:
```suggestion
  exercise_dir=$(dirname ${example_file})
  exercise=$(basename ${exercise_dir})
  cp "${exercise_dir}/.meta/example.ex" "./tmp/src/${exercise}.ex"
```

neenjaw (approved):
🎉 🎉 This is so good, great work, left a couple comments about bash variables when I saw them (there are likely other examples where we should perhaps add some of the braces for bash safety, but can wait for another PR)

PR #585, opened 2021-01-31
Add poker assertion helper

### LINKED ISSUE #554: poker: tests are hard to pass with strict order

neenjaw:
@MeerKatDev: I recreated the PR because structural changes have occured to the repo with the explosion of the ``v3` repo.  Thanks again for championing the issue and bringing it attention.

---

Summary: The poker exercise holds some unstated assumptions about the output of the function.  Namely:

- The cards of the hand passed in is then returned in the same order.
- If multiple hands win, the hands are returned in the order that they are passed in.

This complicates the problem unnecessarily and there is some precedent of including helpers to simplify the comparison (eg. the `allergies` problem's test suite).

## Approach:

On completion, the hands (winning and expected winning) are sorted arbitrarily using `Enum.sort/1`, then compared using `assert x == y` to make use of ExUnit's diff on error.

close #554

### DIFF
--- a/exercises/practice/poker/test/poker_test.exs
+++ b/exercises/practice/poker/test/poker_test.exs
@@ -3,15 +3,18 @@ defmodule PokerTest do
 
   defp assert_poker(winning_hands, expected_winning_hands) do
     winning_hands =
-      winning_hands
-      |> Enum.map(&Enum.sort/1)
-      |> Enum.sort()
+      if winning_hands do
+        winning_hands
+        |> Enum.map(&Enum.sort/1)
+        |> Enum.sort()
+      end
 
     expected_winning_hands =
-      expected_winning_hands
-      |> Enum.map(&Enum.sort/1)
-      |> Enum.sort()
-
+      if expected_winning_hands do
+        expected_winning_hands
+        |> Enum.map(&Enum.sort/1)
+        |> Enum.sort()
+      end
     assert winning_hands == expected_winning_hands
   end
 
--- a/exercises/practice/poker/test/poker_test.exs
+++ b/exercises/practice/poker/test/poker_test.exs
@@ -15,6 +15,7 @@ defmodule PokerTest do
         |> Enum.map(&Enum.sort/1)
         |> Enum.sort()
       end
+
     assert winning_hands == expected_winning_hands
   end
 
--- a/exercises/practice/poker/test/poker_test.exs
+++ b/exercises/practice/poker/test/poker_test.exs
@@ -1,10 +1,25 @@
 defmodule PokerTest do
   use ExUnit.Case
 
+  defp assert_poker(winning_hands, expected_winning_hands) do
+    winning_hands =
+      winning_hands
+      |> Enum.map(&Enum.sort/1)
+      |> Enum.sort()
+
+    expected_winning_hands =
+      expected_winning_hands
+      |> Enum.map(&Enum.sort/1)
+      |> Enum.sort()
+
+    assert winning_hands == expected_winning_hands
+  end
+
   # @tag :pending
   test "single hand always wins" do
     high_of_jack = ~w(4S 5S 7H 8D JC)
-    assert Poker.best_hand([high_of_jack]) == [high_of_jack]
+    winning_hands = Poker.best_hand([high_of_jack])
+    assert_poker(winning_hands, [high_of_jack])
   end
 
   @tag :pending
@@ -12,7 +27,8 @@ defmodule PokerTest do
     high_of_8 = ~w(4D 5S 6S 8D 3C)
     high_of_10 = ~w(2S 4C 7S 9H 10H)
     high_of_jack = ~w(3S 4S 5D 6H JH)
-    assert Poker.best_hand([high_of_8, high_of_10, high_of_jack]) == [high_of_jack]
+    winning_hands = Poker.best_hand([high_of_8, high_of_10, high_of_jack])
+    assert_poker(winning_hands, [high_of_jack])
   end
 
   @tag :pending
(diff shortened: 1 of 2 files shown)


exercises/practice/poker/test/poker_test.exs
```diff
@@ -1,18 +1,34 @@
...
+    winning_hands =
+      winning_hands
+      |> Enum.map(&Enum.sort/1)
+      |> Enum.sort()
+
+    expected_winning_hands =
+      expected_winning_hands
+      |> Enum.map(&Enum.sort/1)
+      |> Enum.sort()
+
```
angelikatyborska, line 14: The first error you get before having implemented anything is pretty ugly right now:
```
  1) test single hand always wins (PokerTest)
     test/poker_test.exs:19
     ** (Protocol.UndefinedError) protocol Enumerable not implemented for nil of type Atom. This protocol is implemented for the following type(s): Range, HashSet, HashDict, GenEvent.Stream, IO.Stream, File.Stream, Function, List, Stream, Map, Date.Range, MapSet
     code: assert_poker(winning_hands, [high_of_jack])
     stacktrace:
       (elixir 1.11.3) lib/enum.ex:1: Enumerable.impl_for!/1
       (elixir 1.11.3) lib/enum.ex:141: Enumerable.reduce/3
       (elixir 1.11.3) lib/enum.ex:3473: Enum.map/2
       test/poker_test.exs:7: PokerTest.assert_poker/2
       test/poker_test.exs:22: (test)
```

Let's make it nicer, like this:
```
 1) test single hand always wins (PokerTest)
     test/poker_test.exs:24
     Assertion with == failed
     code:  assert winning_hands == expected_winning_hands
     left:  nil
     right: [["4S", "5S", "7H", "8D", "JC"]]
     stacktrace:
       test/poker_test.exs:27: (test)
```

By doing this simple code change:

```suggestion
    winning_hands =
      if winning_hands do
        winning_hands
        |> Enum.map(&Enum.sort/1)
        |> Enum.sort()
      end

    expected_winning_hands =
      if expected_winning_hands do
        expected_winning_hands
        |> Enum.map(&Enum.sort/1)
        |> Enum.sort()
      end
```

angelikatyborska (approved):
@MeerKatDev thank you for your contribution 💜 

I personally think the assumption that the solution should do stable sorting (keeping the original order of elements that are considered equal) is reasonable. However, I haven't solved this exercise yet on my own. I have to confess that I gave up because it annoyed me by being so complex 😅. And because of that, I can understand the need to drop that stable sorting assumption and allow any order. If @neenjaw and @MeerKatDev think that's a good idea, I'm on board.

MeerKatDev:
> @MeerKatDev thank you for your contribution 💜
> 
> I personally think the assumption that the solution should do stable sorting (keeping the original order of elements that are considered equal) is reasonable. However, I haven't solved this exercise yet on my own. I have to confess that I gave up because it annoyed me by being so complex 😅. And because of that, I can understand the need to drop that stable sorting assumption and allow any order. If @neenjaw and @MeerKatDev think that's a good idea, I'm on board.

If the need is to specifically avoid the `nil`, I would rather write `unless is_nil(expected_winning_hands) do`, because it better describes the case at hand, but the change in general seems an improvement to me :)

MeerKatDev:
@neenjaw I do understand that there are many more important things to do at the moment, but I think you didn't take into consideration my last comment very much. That syntax that was just implemented just doesn't belong to elixir, it's more of a hack that people use in javascript because of the ambiguity between `NaN`, `undefined`, and `null`, in other languages an `if` construct should always have a boolean, as it's the reason the `if` construct exists. I remind that the code serves also as an example to people coming to learn on the platform.

SleeplessByte:
Hi there!

PHP, Python, Groovy, Javascript, and Ruby, among others, all support thruthy- and falsiness, and it's often idiomatic to use it. Are you arguing that the language designers of Elixir made a mistake? I'm not asking to be annoying, but to better understand _what exactly_ you're arguing. As a non-elixir developer, it feels like a stylistic choice? 

For example, I often write:

```javascript
if (something === null) {

}
```

But I will almost _never_ write:
```ruby
if (something.nil?) 

end
```

But I have people seen do the exact opposite. Feels stylistic, when the falsy values passed in is only ever `nil`.

neenjaw:
Hey @MeerKatDev, just writing a quick note to say that I've seen your reply, I don't have time right now to respond fully, but will do so tonight after work.

MeerKatDev:
> Hi there!
> 
> PHP, Python, Groovy, Javascript, and Ruby, among others, all support thruthy- and falsiness, and it's often idiomatic to use it. Are you arguing that the language designers of Elixir made a mistake? I'm not asking to be annoying, but to better understand _what exactly_ you're arguing. As a non-elixir developer, it feels like a stylistic choice?
> 
> For example, I often write:
> 
> ```js
> if (something === null) {
> 
> }
> ```
> 
> But I will almost _never_ write:
> 
> ```ruby
> if something.nil?
> 
> end
> ```
> 
> But I have people seen do the exact opposite. Feels stylistic, when the falsy values passed in is only ever `nil`.

@SleeplessByte That it is a very ambiguous way of checking things. Some languages enforce strictness, some don't. If I write:

`if is_nil(something) do`

it's unambiguous what I'm checking for. I cannot go wrong. I'm checking if `something` is `nil`. I know that `something` gives an object of some type or a nil. It's a nullable type. If I instead write

`if something do`

and that `something` is a function (in general), you don't know if the return type is a boolean or an object. What am I checking for? There are sketchy developers that like unreadable functions, so sometimes you may meet functions returning `true`, `false`, or `nil`. Also, there are a lot of newbies in elixir, which might be used to the fact that their previous language interprets falsyness in some other ways. So, that's why I'm always for strict checking/comparison, because you cannot be wrong.

SleeplessByte:
@MeerKatDev I absolutely 100% agree with you that using implicit true-and-falseness is not as strict as... well doing the strict comparison. I generally enjoy more strict checks, because it protects me from some mistakes calling the functions, and it is just clearer what is tested.

However, I don't think the stricter checks are _more right_ or _more idiomatic_. It's stylistic in the context of these exercises. Maybe that's the missing link here? 

>  you don't know if the return type is a boolean or an object.

Isn't that the point of truthiness and falsiness tho? We don't care if it's a boolean, as long as it can be interpreted like one? If a function returns `true`, `false` or `nil`, then it's up to the intended behaviour if the check should be strict, I think.

```elixir
if maybe_boolean do
  # something if maybe_boolean == true
```

The above would work fine if we only care about `maybe_boolean` being true. I don't think this is better or worse than explictly checking: ```not is_nil(maybe_boolean) and maybe_boolean``` or something similar that's actually valid Elixir. 

However, if `nil` means "absence of a value" and should maybe have special behaviour, I'd expect to see:

```elixir
if is_nil(maybe_boolean) do
  # special case for maybe_boolean == nil 
  # if this returns early or errors, then we can continue below
  
if maybe_boolean do
  # etc
```

Am I wrong thinking the _correctness_ of strict checking is contextual and not idiomatic-based?

MeerKatDev:
> However, if nil means "absence of a value" 

That is what I'm contesting. We are not checking falsiness, we are checking if something is `nil` here. Normally I wouldn't say anything, but I see sometimes bad habits that can be avoided, and non-strict checking, IMHO, is a bad habit, which beginners shouldn't pick up. I mean, they might like that at some point, but I work in several languages and there are a series of prime principles which I follow and that work well in any programming language, so I personally tend to stick to certain rules whatever I write.

> However, I don't think the stricter checks are more right or more idiomatic

I meant that, specifically in Elixir, I _rarely_ saw such a non-specific check (I'd get mad, so I'd remember), and for that reason "it just doesn't belong to elixir", because much more often people use pattern matching or `is_nil`.

It's instead (or it was?) a common thing in js, because we didn't want to write shenanigans of the like:
`if (sth != undefined && sth != null && sth != NaN && sth != false)`, while you could just write `if (sth) {` instead.
People started to use that in other languages. And then there was mayhem.

SleeplessByte:
> We are not checking falsiness, we are checking if something is nil here (so, it's contextual, as you say) .

Except that I don't see this in the exercise's problem description or tests. In other words: the exercise doesn't break if `is_nil` is not used, and that seems to be intended (read: the context).

It sounds mostly that you feel that using truthiness and falsiness is bad-habit, which is an opinion many developers will share with you. However, for each of the languages I mentioned, there are examples of _using_ this language feature in the language docs itself (including Elixir), which further establishes that it's as intended.

This leads me to the conclusion that, whilst I empathize with what you're saying, ánd understand _why_ you're saying it, that going one route over the other is a stylistic choice and not one of idiomatic use. That doesn't mean your opinion is wrong or invalid, or that Tim, Angelika, or I are against it, but it means that we made a conscious decision to allow all stylistic choices in this exercise.

However, I believe Exlir will have or already has specific _concept_ exercises that teach about nilness, optional/maybe/none and/or truthiness and falsiness. It's also perfectly fine as a **mentor** to voice your opinion during a mentoring session, but that's different than hard-coding it into the exercise.

MeerKatDev:
First of all @SleeplessByte , please note that we are talking about some function used strictly in the tests, and not something about the implementation of the exercise. Just saying that if some student was to go in the test file and check it out, they might take that as an example.

 And I feel like we are not understanding each other here, I don't see how this was the message taken from my comments?

> you feel that using truthiness and falsiness is bad-habit

I'm pretty confident that there is something that is always _more idiomatic_ than something else, which is what I wanted to do here. These approaches don't mean the same, so it's not a stylistic choice.

But I think we went to greater lengths than this was supposed to be. I just wanted to point out that while you're free to merge what and how you prefer, I might at least get a hint on why my last comment was not something worthy addressing. I have some good years of experience with this language, but this doesn't seem like a good reason to listen.

SleeplessByte:
I can't speak for everyone else, but in general, no response doesn't mean I'm not listening.

Just because we're talking about the `assert` inside the test file doesn't change what I said or how I feel about this particular issue. The outcome is the same. The tests will run the same. The Elixir documentation has examples of using truthiness, which make it okay to use in my opinion. The discussion I _think_ we're having is still the underlying "using truthiness is bad or not bad", and that is a stylistic choice, unless the code paths would be different if you were explicit versus implicit. 

Are you okay with us not agreeing here, but us (you and me) still having the same intention / end goal?

angelikatyborska:
@MeerKatDev I can 100% assure you that using `if x` to guard against `nil` values is perfectly fine. I have been writing production Elixir code for 3 years now, and I see it everywhere. I experienced a big share of awful bugs in my Elixir code, but it was never because if `if x` :).

I understand that it's hard to believe that if you personally really dislike this construct. Here are some examples from very famous Elixir code bases that I found for you:

Mix:

https://github.com/elixir-lang/elixir/blob/76d245b6081c53228bf99fc1494add5de7872065/lib/mix/lib/mix/dep/loader.ex#L72
https://github.com/elixir-lang/elixir/blob/76d245b6081c53228bf99fc1494add5de7872065/lib/mix/lib/mix/shell/io.ex#L15
https://github.com/elixir-lang/elixir/blob/76d245b6081c53228bf99fc1494add5de7872065/lib/mix/lib/mix/scm/git.ex#L10

ExUnit:

https://github.com/elixir-lang/elixir/blob/master/lib/ex_unit/lib/ex_unit/assertions.ex#L153
https://github.com/elixir-lang/elixir/blob/master/lib/ex_unit/lib/ex_unit/callbacks.ex#L207
https://github.com/elixir-lang/elixir/blob/master/lib/ex_unit/lib/ex_unit/doc_test.ex#L210
https://github.com/elixir-lang/elixir/blob/master/lib/ex_unit/lib/ex_unit/runner.ex#L245

Ecto:

https://github.com/elixir-ecto/ecto/blob/67a7f5795376e1c1a03b2e7dc4ba4f97da9eb7c5/lib/mix/ecto.ex#L26
https://github.com/elixir-ecto/ecto/blob/67a7f5795376e1c1a03b2e7dc4ba4f97da9eb7c5/lib/ecto/repo/supervisor.ex#L12
https://github.com/elixir-ecto/ecto/blob/26b441371d715497f47c0c5fc535042d9a84c301/lib/ecto/repo/preloader.ex#L63
https://github.com/elixir-ecto/ecto/blob/67a7f5795376e1c1a03b2e7dc4ba4f97da9eb7c5/lib/ecto/exceptions.ex#L43

Phoenix:

https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/config.ex#L100
https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/controller.ex#L1278
https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/router.ex#L674
https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/endpoint/cowboy2_adapter.ex#L70
https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/endpoint/supervisor.ex#L412

That being said, I agree that the test suite should give students a good example. I think a mere `nil` check in that case is not enough. The code requires an enumerable, and we know for sure that it's should be not any enumerable, but a list. Thus my PR: https://github.com/exercism/elixir/pull/588

MeerKatDev:
Thanks @angelikatyborska for the PR :) even better!

That said, here I'm not contesting the use of the `if` construct at all, I use `if/else` myself all the time. What I was contesting is its use to check against non-boolean types - `if object do ..` vs `if bool do ..`, and I think that the second should be always preferred. But I thought that was clear.

iHiD:
I'm gonna jump in here as we're not getting anywhere productive, and we all have lots of more important things to be cracking on with for getting v3 launched 🙂 

I'll summarise where we're at and draw a line under it...

I think it's clear that @MeerKatDev doesn't like conditional-testing for truthiness in Elixir and prefers strict boolean testing. For my own interest, I've gone and checked the Elixir docs, and they explicitly make it clear that it's valid in Elixir [1]. The examples that @angelikatyborska points to are also lots of uses of `if` checking for truthiness (`nil` not a boolean). So however strong that @MeerKatDev's opinion is, it's not a strong enough to be considered the "idiomatic Elixir opinion" as both the docs and large, popular codebases are clearly fine with it. So then it becomes a decision that is down to the maintainer team to make, and in this case, Tim and Angelika are our senior Elixir Maintainers, so their judgement becomes Exercism's position on this. So we'll stick with it being ok to use conditionals with truthiness on the Elixir Track 🙂  

Thanks for your thoughts @MeerKatDev - it's important that people bring their different experiences in programming in languages to Exercism to make sure that we're making things idiomatic where possible - and do the others for the healthy debate 🙂 

Let's all leave this here and crack on 💙 

---

[1] https://hexdocs.pm/elixir/Kernel.html#if/2

Screenshot 2021-02-02 at 14 51 35

PR #589, opened 2021-02-03
Remove Elixir < 1.5 chunk/chunk_every fix in `largest-series-product`

angelikatyborska:
I noticed this confusing snippet in the example solution for `largest-series-product`. I think it was a fix necessary for the change that happened between Elixir 1.4 and 1.5. That's a long time ago, and our CI runs 1.8 - 1.11, so I think we can get rid of it.


neenjaw (approved):
This seems reasonable to me

PR #593, opened 2021-02-05
Move and reformat hints

angelikatyborska:


exercises/practice/secret-handshake/hints/hints.md
```diff
@@ -1,20 +0,0 @@
...
-Check: 0b11010
-
-Flags &&& Check: 0b11010 (All checked bits are set)
-
-Another:
-
-Flags: 0b11011
-Check: 0b10110
-
-Flags &&& Check: 0b10010 (Third bit not set)
```
angelikatyborska, line 20: I don't know how to format this whole text as a list item that would fit on the website. AFAIK sublists and code blocks are not allowed in hints.

ErikSchierboom: I'm not sure actually. I _think_ our only requirement is for the hints to be in a list. I don't know for sure though.

ErikSchierboom (approved):
LGTM!

exercises/practice/secret-handshake/hints/hints.md
angelikatyborska: I had to remove sublists from concept exercise hints before because they would break the website.

ErikSchierboom: Ah okay, I didn't know. Sorry

PR #594, opened 2021-02-06
Add track-specific docs

angelikatyborska:
According to https://github.com/exercism/docs/blob/aa78d01ec239bab1c9a66a9288e7464ba62ecfbc/anatomy/tracks/shared-files.md#track-specific-files

### DIFF
--- a/exercises/shared/.docs/debug.md
+++ b/exercises/shared/.docs/debug.md
@@ -1,0 +1,9 @@
+# Debugging
+
+To help with debugging, you can use the fact that all console output will be captured and shown in the test results window. You can output any value by passing it to `IO.inspect`:
+
+```elixir
+"hello"
+|> String.upcase()
+|> IO.inspect()
+```
--- a/exercises/shared/.docs/debug.md
+++ b/exercises/shared/.docs/debug.md
@@ -1,9 +1,12 @@
 # Debugging
 
-To help with debugging, you can use the fact that all console output will be captured and shown in the test results window. You can output any value by passing it to `IO.inspect`:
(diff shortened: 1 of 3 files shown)


exercises/shared/.docs/debug.md
```diff
@@ -0,0 +1,9 @@
+# Debugging
+
+To help with debugging, you can use the fact that all console output will be captured and shown in the test results window. You can output any value by passing it to `IO.inspect`:
+
+```elixir
+"hello"
+|> String.upcase()
+|> IO.inspect()
```
angelikatyborska, line 8: I think this is the best example of how `IO.inspect` gets used in the wild, but it's not the simplest. Would it be better to give a simpler example, without pipes or the `String` module?

neenjaw: or maybe put the inspect between two functions and comment how it allows us to introspect the intermediate value?

neenjaw (approved):

exercises/shared/.docs/tests.md
```diff
@@ -0,0 +1,30 @@
...
+
+```elixir
+# @tag :pending
+test "shouting" do
+  assert Bob.hey("WATCH OUT!") == "Whoa, chill out!"
+end
+```
+
+Or, you can enable all the tests by commenting out the
+`ExUnit.configure` line in the test suite.
```
neenjaw, line 26: Maybe should point them to the `test_helper.exs` file?

angelikatyborska: Good idea, the current wording in misleading.

neenjaw (approved):

PR #595, opened 2021-02-06
New concept exercise `dates-and-time`

### LINKED ISSUE #584: [v3] Mention `:calendar` module in the `erlang-libraries` concept
More than two exercises in our track rely on that module, so it would be necessary to mention it, at the very least in `about.me` of the `erlang-libraries` concept.

### LINKED ISSUE #586: [v3] Implement new Concept Exercise: `dates-and-time`
## Learning objectives

- the `Date` module
- the `Time` module
- the `DateTime` module
- the `NaiveDateTime` module
- the Erlang `:calendar` module
- comparing dates/datetimes, moving datetimes, changing timezones maybe?
## Out of scope

?

## Concepts

- `dates-and-time` (or a nicer name that describes the same set of ideas :) )

## Prerequisites

- `erlang-libraries`
- other, depending on the story

## Practice exercises

Those practice exercises should have `dates-and-time` in their prerequisites and as `practices`:
- `meetup`
- `gigasecond`

angelikatyborska:
closes #586, closes #584

This is WIP. I first want to get early feedback on the story.


config.json
```diff
@@ -481,6 +481,21 @@
           "pids"
         ],
         "status": "wip"
+      },
+      {
+        "slug": "library-fees",
+        "name": "Library Fees",
+        "uuid": "b5629003-638c-47c0-ad76-a0769f285053",
+        "concepts": [
+          "dates-and-time"
```
angelikatyborska, line 490: I have been wondering if `dates-and-time` is the best name for this. Alternatives that I can think of:
- `date-and-time` (I would prefer the concept names to be in plural, like `atoms` and `tuples`, but `time` in this context has no plural and that makes `dates-and-time` sound weird?)
- `time-and-date`
- `calendar` (but it might be confused with the `Calendar` behaviour or the Erlang `:calendar` module
- `datetime`
- three separate concepts: `date`, `time`, and `naive-date-time`?

exercises/concept/library-fees/.docs/instructions.md
```diff
@@ -0,0 +1,61 @@
+Your librarian friend has asked you to extend her library software to automatically calculate late fees.
+Her current system stores the exact date and time of a book checkout as an ISO8601 datetime string.
+She runs a local library in a small town in ..., which uses the GMT timezone (UTC +0), and doesn't need to worry about other timezones.
```
angelikatyborska, line 3: @iHiD I'm not sure which country to choose here. It doesn't matter for the exercise at all, but I didn't want to be Europe-centric yet again... Due to my white privilege and not paying attention in geography class, I don't know much about Africa. I would want to avoid accidentally naming a country with gross human right violations, and where a woman running a library on her own wouldn't even be possible. Here are the GMT African countries:
- Mali
- Morocco
- Western Sahara
- Mauritania
- The Gambia
- Senegal
- Guinea-Bissau
- Guinea
- Sierra Leone
- Liberia
- Burkina Faso
- Côte d'Ivoire
- Ghana

Any recommendations? Or should I just stick to what I know and choose Northern Ireland or something?

exercises/concept/library-fees/.meta/exemplar.ex
```diff
@@ -0,0 +1,41 @@
...
+  def calculate_late_fee(checkout, return, rate) do
+    checkout_datetime = datetime_from_string(checkout)
+    planned_return_datetime = return_datetime(checkout_datetime)
+    actual_return_datetime = datetime_from_string(return)
+
+    days_late = days_late(planned_return_datetime, actual_return_datetime)
+    rate = if monday?(actual_return_datetime), do: 0.5 * rate, else: rate
+    trunc(days_late * rate)
+  end
+end
```
angelikatyborska, line 41: I was wondering if this is maybe a bit too much? I could easily remove either the Monday condition or the before noon condition to make the exercise a bit simpler. I included the Monday condition to teach about `Date.day_of_week()` and the noon condition to force the student to write a time literal with the `~T` sigil and to do anything related to time.

neenjaw:
I like the story, I think it is a good idea to use another country, but I'm not sure I can help you decide on a particular country because I, like you, am naive to the situation of women proprietors in those countries. 

I think it's a reasonable exercise and task since naive date times have good library functions for this with a bit of math to convert days to seconds to calc the offset

exercises/concept/library-fees/.docs/instructions.md
iHiD: Thanks for thinking/caring! :)

Mali and Ghana are solid choices. I believe Senegal is making good strides forward with Human Rights. I don't know about the others.

concepts/erlang-libraries/about.md
```diff
@@ -51,3 +52,5 @@ Elixir interoperability with Erlang libraries is not limited to Erlang's standar
 [erl-queue]: http://erlang.org/doc/man/queue.html
 [erl-crypto]: http://erlang.org/doc/man/crypto.html
 [erl-zip]: http://erlang.org/doc/man/zip.html
+[erl-calendar]: https://erlang.org/doc/man/calendar.html
+[erl-calendar-iso-week-number]: https://erlang.org/doc/man/calendar.html#iso_week_number-1
```
angelikatyborska, line 56: After the addition of `Date.to_gregorian_days` and `DateTime.to_gregorian_seconds` in Elixir 1.11, I think `iso_week_number` is the only remaining Erlang calendar function that doesn't have an Elixir equivalent.

config.json
```diff
@@ -481,6 +481,22 @@
...
+        ],
+        "prerequisites": [
+          "integers",
+          "floating-point-numbers",
+          "tuples",
+          "atoms",
+          "if"
```
neenjaw, line 497: It's mentioned in the concept but not in the introduction about the erlang calendar module. From what I see in the exercise this isn't required in the exercise, but perhaps if there will be a locking component to concept progression then the erlang modules should be a pre-req to the concept?

exercises/concept/library-fees/.meta/exemplar.ex
neenjaw: I don't think it is too much from the perspective of any one function, but it is a longer exercise overall than some of our other ones. Maybe that is what you are sensing?

neenjaw (approved):
I like it, I hear your concern about it being "too much" and my take on it is that it just feels like a longer exercise with 6 steps. The instruction progression should prepare them for that final step though.

exercises/concept/library-fees/.meta/exemplar.ex
angelikatyborska: Possibly. After I thought about it again, the exercise feels simple enough, I think there aren't any tricky steps.

config.json
angelikatyborska: Yeah, the `calendar` module is not necessary at all for this exercise. I found that the Erlang `calendar` module is way less useful than I thought it would be, in that the Elixir stdlib already has a lot. My assumption was that it's ok for a `concept/about.md` to violate prerequisites a bit in the service of completeness. For example, in `processes/about.md`, `Agent` and `GenServer` are briefly mentioned. If you think that in this case it's too much, I think I would rather just remove the mention of `:calendar` and `to_erl` from this file and not have a prerequisite on `erlang-libraries`.

neenjaw: You're more in touch with that than me, and if that is allowable in the other circumstances I think this is probably quite fine in this case.

angelikatyborska: After a second thought, I decided that the Erlang date format is not so important in this context and just removed the mention from the `about.md` file.

Issue #581, opened 2021-01-30
Bring all exercises up to date with the current canonical data


angelikatyborska:
For every practice exercise, we need to:
- Remove the outdated `/.meta/version` file.
- Update the `/.meta/tests.toml` file (see: https://github.com/exercism/problem-specifications#track-test-data) so that it accurately represents which of the canonical tests we implement. Canonical data can be found at: https://github.com/exercism/problem-specifications/tree/main/exercises/
- Optionally, if one of the canonical tests is not implemented for the Elixir version of the exercise, it can be added, depending on our own judgement. If a very similar test already exists in the Elixir version of the exercise, it could be replaced by the canonical test.

This can be done in chunks, by different people.

Current progress:

- [x] accumulate
- [x] acronym
- [x] all-your-base https://github.com/exercism/elixir/pull/597
- [x] allergies
- [x] alphametics
- [x] anagram
- [x] armstrong-numbers
- [x] atbash-cipher
- [x] bank-account
- [x] beer-song
- [x] binary ~https://github.com/exercism/elixir/pull/604~ https://github.com/exercism/elixir/pull/620 **deprecated**
- [x] binary-search
- [x] binary-search-tree https://github.com/exercism/elixir/pull/605
- [x] bob
- [x] bowling
- [x] change https://github.com/exercism/elixir/pull/606
- [x] clock https://github.com/exercism/elixir/pull/607
- [x] collatz-conjecture
- [x] connect https://github.com/exercism/elixir/pull/608
- [x] crypto-square
- [x] custom-set https://github.com/exercism/elixir/pull/609
- [x] diamond https://github.com/exercism/elixir/pull/610
- [x] difference-of-squares https://github.com/exercism/elixir/pull/611
- [x] diffie-hellman
- [x] dnd-character https://github.com/exercism/elixir/pull/612
- [x] dominoes
- [x] dot-dsl
- [x] etl https://github.com/exercism/elixir/pull/613
- [x] flatten-array https://github.com/exercism/elixir/pull/614
- [x] forth https://github.com/exercism/elixir/pull/615
- [x] gigasecond https://github.com/exercism/elixir/pull/616
- [x] grade-school https://github.com/exercism/elixir/pull/663
- [x] grains
- [x] grep https://github.com/exercism/elixir/pull/617
- [x] hamming https://github.com/exercism/elixir/pull/619
- [x] hello-world
- [x] hexadecimal https://github.com/exercism/elixir/pull/620 **deprecated**
[...]

PR #597, opened 2021-02-11
Bring `all-your-base` up to date with canonical data

angelikatyborska:
I'm taking a stab at https://github.com/exercism/elixir/issues/581

I'm not sure if I'm doing this right... all of the tests from canonical data were already there (https://github.com/exercism/problem-specifications/blob/9b30575a999d56b217f04510846c85d825c2a057/exercises/all-your-base/canonical-data.json) but some of them had different return values than suggested. If I am interpreting this correctly: https://github.com/exercism/problem-specifications/blob/9b30575a999d56b217f04510846c85d825c2a057/exercises/all-your-base/canonical-data.json#L244

This should return or raise a language-specific error-thingy. I think in Elixir that should be an error tuple. I don't know if this was already considered and rejected in favor of `nil`. According to the `version` file, this exercise was already supposed to be up to date.

But I don't know if I could just leave `true` values in `tests.toml` without modifying the exercise... Changing `true` to `false` for test cases that were implemented but had a different return value would also be weird.


neenjaw (approved):
does the tests.toml file need to be touched for this?

angelikatyborska:
No, all the test cases from the current problem specification were already implemented, and there are no extra ones. So `tests.toml` for this exercise is up to date as it is now.

neenjaw:
I think the first 10-20 exercises were almost all brought up to the canonical spec when we (me and @Br1ght0ne) tried to do this before.

PR #607, opened 2021-02-15
Bring `clock` up to date with canonical data

angelikatyborska:
Two test cases were missing, and the example solution turned out to be incorrect after adding the new test case.


exercises/practice/clock/.meta/example.ex
```diff
@@ -25,26 +25,29 @@ defmodule Clock do
...
+        %Clock{hour: hour - 24, minute: minute}
+        |> rollover()
+
+      {hour, minute} when hour < 0 ->
+        %Clock{hour: hour + 24, minute: minute}
+        |> rollover()
+
+      _ ->
+        clock
+    end
```
angelikatyborska, line 48: I just copy-pasted part of my own solution to clock instead of trying to figure out where the bug was.

neenjaw (approved):

PR #615, opened 2021-02-16
Bring `forth` up to date with canonical data

angelikatyborska:
WIP: Missing an example solution that will pass the tests.


angelikatyborska:
@neenjaw Would you be able to help debug this GHA failure? I don't see any useful errors in the output. I don't know GHA very well... on CircleCI, I would have just "re-run job with ssh" and had access to the machine, but here I have no idea how to gain access, or how to try out this exercise on Elixir 1.6.0 + OTP 19.

I tried running the oldest Elixir still available on Dockerhub (1.6.6 + OTP 20) and there the tests passed.

NobbZ:
Not sure why it is firing only for 19/1.6, but it seems as if it is related to how the bash builtin does its argument parsing…

I can reproduce it locally in a bash:

```text
$ printf "-- %s" "foo"
bash: printf: --: invalid option
printf: usage: printf [-v var] format [arguments]
$ printf -- "-- %s" "foo"
-- foo
```

So adding a simple `--` in front of the format specifier (or removing the double dashes at all) should make it work.

The most compatible way is probably removing the dashes at all.

NobbZ:
Please be also aware of the fact that `actions/setup-elixir` has been deprecated in favor of the [`erlef/setup-elixir`](https://github.com/erlef/setup-elixir) action.

angelikatyborska:
@NobbZ Thanks for the heads up about the action! I'll attempt to update it here: https://github.com/exercism/elixir/pull/678 it doesn't seem to work out of the box.

angelikatyborska:
It looks like all the changes to the CI made the tests magically pass 🙃

neenjaw (approved):

PR #616, opened 2021-02-16
Bring `gigasecond` up to date with canonical data

angelikatyborska:


exercises/practice/gigasecond/test/gigasecond_test.exs
```diff
@@ -2,20 +2,30 @@ defmodule GigasecondTest do
   use ExUnit.Case
 
   # @tag :pending
-  test "from 4/25/2011" do
+  test "from 2011-04-25 00:00:00" do
```
angelikatyborska, line 5: More international and programmer-friendly date format...

exercises/practice/gigasecond/lib/gigasecond.ex
```diff
@@ -3,8 +3,7 @@ defmodule Gigasecond do
   Calculate a date one billion seconds after an input date.
   """
   @spec from({{pos_integer, pos_integer, pos_integer}, {pos_integer, pos_integer, pos_integer}}) ::
-          :calendar.datetime()
-
+          {{pos_integer, pos_integer, pos_integer}, {pos_integer, pos_integer, pos_integer}}
```
angelikatyborska, line 6: The input and output are of the same type, so they should be the same in the type spec. I think `{{pos_integer, pos_integer, pos_integer}, {pos_integer, pos_integer, pos_integer}}` is way uglier to write, but more clear to beginners. I'm not sure...

neenjaw: 🤷🏻 either or, I'm not sure that detail is limiting.

exercises/practice/gigasecond/test/gigasecond_test.exs
neenjaw: nice usability change

neenjaw (approved):

PR #620, opened 2021-02-18
Deprecate `binary` and `hexadecimal` in favor of `all-your-base`

angelikatyborska:
5 years late... but let's do this 😂. According to 
- https://github.com/exercism/problem-specifications/issues/279
- https://github.com/exercism/problem-specifications/blob/main/exercises/binary/.deprecated
- https://github.com/exercism/problem-specifications/blob/main/exercises/hexadecimal/.deprecated

`binary`, `trinary`, `octal`, and `hexadecimal` should all be considered deprecated, in favor of `all-your-base`


angelikatyborska:
@ErikSchierboom Is this a correct way to deprecate an exercise? Setting the status key, nothing else?

ErikSchierboom (approved):
@angelikatyborska Yes, that is the correct way to do it. Note that we haven't yet implemented the code to do anything with the status, so it won't have any effect right now, but it will.

PR #626, opened 2021-02-18
Bring `list-ops` up to date with canonical data

angelikatyborska:
- Renamed function `count` -> `length`
- Renamed function `reduce` -> `foldl`
- New function `foldr`
- Rearranged test suite to use `describe` blocks
- Almost all test cases stay unchanged, I'm keeping our custom test cases for "huge" lists, including for `foldr`
- Increased exercise difficulty because of the addition of `foldr`


neenjaw (approved):

frerich:
I think I lack some context. I wonder _why_ this renaming is done. Maybe the commit log

> Bring `list-ops` up to date with canonical data

could be made a little more elaborate? What 'canonical data' is meant here?

frerich:
Another question, which is most certainly triggered by my lack of understanding how the repository is structured than by this PR itself: what is the purpose of the `exercises/practice/list-ops/.meta/example.ex` file?

angelikatyborska:
@frerich 

> What 'canonical data' is meant here?

There is this repository called "problem specifications" (https://github.com/exercism/problem-specifications). This repository is important to the whole Exercism platform. It defines the list of all practice exercises, for example `list-ops` (https://github.com/exercism/problem-specifications/tree/main/exercises/list-ops). Each exercise is defined by:
- its metadata https://github.com/exercism/problem-specifications/blob/main/exercises/list-ops/metadata.yml
- description, which is the first document that the user will read after starting the exercise https://github.com/exercism/problem-specifications/blob/main/exercises/list-ops/description.md
- canonical data, which is a list of all test cases that the exercise should implement https://github.com/exercism/problem-specifications/blob/main/exercises/list-ops/canonical-data.json 

It's up to the track to decide which test cases from canonical data make sense and how exactly to implement them (e.g. "errors" might mean different things in different exercises).

> I wonder _why_ this renaming is done.

The simplest answer to that question is: because canonical-data says so. My personal approach to the problem of how much the Elixir track should differ from the Exercism-wise problem specifications is: as little as possible. There are many more people making sure that problem specifications make sense than there are people maintaining the Elixir track, so I both trust them to make better judgements than I can, and I don't want to carry that extra burden of defining non-Elixir-specific details of exercises.

Looking at the history of the problem specifications repository, `list-ops` always included `foldl` and `foldr`. Why Elixir had a `reduce` instead? No idea.

> what is the purpose of the `exercises/practice/list-ops/.meta/example.ex` file?

The only purpose of that file is to prove that the exercise can be solved. It doesn't need to be pretty, it doesn't need to be idiomatic, it most likely wouldn't make it into the mentoring notes :). It just has to pass all of the tests.

frerich:
Thanks for the elaborate response -- that helps a lot! I think your motivation is perfectly plausible! Just one thought:

I think that this exercise teaches some very useful lessons, two of them being:

a) there are many functions which seem different on the surface (e.g. `count` and `map` and `filter`) but which are all just variations on the same general theme of recursion
b) in many cases, you don't need to resort to manual recursion but can use `Enum.reduce` (I believe `Enum.reduce` is more commonly used than `List.foldl` or `List.foldr`, but I may be mistaken)

I really liked how this exercise asked students to define `reduce` since that maps directly to the `Enum.reduce` function they would come across in 'real life Elixir code'..

With `foldl` and `foldr`, this is maybe a little less clear -- i.e. mentoring would probably end with "Oh, and by the way: everything you learned about foldl actually applies to `Enum.reduce`, too!". Also, I noticed that _all_ students I mentored on this exercise so far instinctively defined `reduce` as `foldl`. Pointing out that

* there is a second way to define `reduce`
* the alternative way actually avoids having to use `reverse` when building lists

always seemed to be a nice light-bulb moment. So many asking to implement `foldr` and `foldl` right away is a bit of a spoiler. :-)

That being said, I fully agree with your arguments about maintaining the track - I feel they outweigh mine, and I think this PR could go in as it is.

angelikatyborska:
@frerich Thank you a lot for the input 🙏. I agree with your observations about mentoring this exercise, I noticed the same things. I'm hoping implementing `foldl` and `foldr` will be even more illuminating to student, I'm looking forward to reports on how they will handle the new part of the exercise 😏.

PR #628, opened 2021-02-18
Bring `markdown` up to date with canonical data

angelikatyborska:
4 new test cases.

WIP: the initial solution and the example solution both need to be fixed not to fail the new test cases.


neenjaw (approved):

neenjaw:
I like the additional test cases, looking forward to the improved problem/example

exercises/practice/markdown/lib/markdown.ex
```diff
@@ -70,10 +70,9 @@ defmodule Markdown do
...
   defp patch(l) do
-    String.replace_suffix(
-      String.replace(l, "<li>", "<ul>" <> "<li>", global: false),
-      "</li>",
-      "</li>" <> "</ul>"
-    )
+    String.replace(l, "<li>", "<ul>" <> "<li>", global: false)
+    |> String.reverse()
+    |> String.replace(String.reverse("</li>"), String.reverse("</li></ul>"), global: false)
+    |> String.reverse()
```
angelikatyborska, line 76: @neenjaw What do you think about this change to the "pre-refactor" boilerpalte? It's a bit ridiculous, but it fixes the tests and I think makes it clear it needs to be refactored 😁

neenjaw: Ya, it is certainly clear that it could be improved :P

I like it ✅

PR #634, opened 2021-02-20
Bring `nth-prime` up to date with canonical data

angelikatyborska:
One new test case that makes the existing solution way too slow :)


neenjaw (approved):
This looks good, I didn't test the new example solution, but it does look like it would operate much faster than the previous since it only uses the previously found primes to check for divisibility.

Is it fast enough to satisfy your comment?

angelikatyborska:
Yes, this one is fast enough

PR #639, opened 2021-02-20
Bring `phone-number` up to date with canonical data

angelikatyborska:
This exercise gets a complete makeover. Canonical data defines many new corner cases. After seeing how many there are, I decided to drop the additional functions that Elixir had (`pretty`, `area_code`) because we would need to repeat all of that error checking in the tests for those functions too.

With the new corner cases, the solution becomes more complex and has new prerequisites. I took one shortcut for the corner cases, described in `design.md`.

To mark even more that this is a new, canonical-data-complaint version, it gets a new module name and function name, to match canonical data.


exercises/practice/phone-number/.meta/example.ex
```diff
@@ -1,106 +1,80 @@
...
   end
 
-  defp valid?(num) do
-    Regex.match?(~r/^1?(?:[2-9][0-9]{2}){2}\d{4}$/, num)
-    |> (fn
-          true -> num
-          false -> false
-        end).()
+  defp remove_separators(number) do
+    String.replace(number, ~r/[\s+.()-]/, "")
```
neenjaw, line 26:
```suggestion
    String.replace(number, ~r/[^0-9]/, "")
```
Maybe this would be less prone to breaking if the testcases change in the future?

neenjaw (approved):
Big change, but it looks good.  The clean/validate are the crux of this problem, so I think it is okay to focus on them.

exercises/practice/phone-number/.meta/example.ex
angelikatyborska: We can't strip all non-digits at this point because we need to check later if, after removing those few symbols, there are any not allowed symbols left. The list of symbols that are allowed, but should be removed (so - this regex), didn't really change.

neenjaw: Ah, ok!

PR #643, opened 2021-02-20
Bring `pythagorean-triplet` up to date with canonical data

angelikatyborska:
WIP: Missing a solution

A few new test cases, including one for very big numbers. They make the current solution way too slow. To enable the student to explore different optimizations, the additional max/min arguments were removed.


MeerKatDev:
The solution in `.meta` passes all the tests well, so I tried to optimise it in terms of execution speed. On my laptop, the first run took 28 sec. Removing `Enum.to_list` from the for-comprehension (~8s less, not needed) and moving the `pythagorean?` body to the guard (~1s less), I could take it down to 20s, and I cannot find any other way to improve the speed. After some debugging, it seems that the offending line is this `y <- if((sum - 2 * x) > x, do: x..(sum - 2 * x), else: [])`, and I think it's already the optimal way.

The only thing I could think of that could speed it up would be to parallelise it, but I suppose it's way too much hassle for this case. I'm not sure whether recursion would improve anything.

angelikatyborska:
I tried a pretty simple task approach:
```elixir
  def generate(sum) do
    Enum.map(sum..1, fn x ->
      Task.async(fn ->
        for y <- if(sum - 2 * x > x, do: Enum.to_list(x..(sum - 2 * x)), else: []),
            z <- if(sum - x - y != 0, do: [sum - x - y], else: []),
            pythagorean?([x, y, z]),
            do: [x, y, z]
      end)
    end)
    |> Task.await_many()
    |> Enum.reduce([], fn list, acc -> list ++ acc end)
  end
```

I benchmarked it against the previous solution for `Triplet.generate(30000)` and it turns out to be 3x faster:
```


Name                  ips        average  deviation         median         99th %
task               0.0795        12.58 s     ±0.00%        12.58 s        12.58 s
traditional        0.0234        42.79 s     ±0.00%        42.79 s        42.79 s

Comparison: 
task               0.0795
traditional        0.0234 - 3.40x slower +30.21 s
```

I think that's good enough :) the whole test suite runs in under 15 seconds now on my machine. I think this could be a nice exercise to teach about parallelization using tasks.

MeerKatDev:
Great, I was too lazy to write and think about it myself :) does the performance improve without using `Enum.to_list` ?

I read somewhere (cannot find it now) that in some future version there should appear parallel for comprehensions. It would be a great addition!

angelikatyborska:
> does the performance improve without using Enum.to_list ?

Yes!

I run another benchmark, using those four different implementations:
```elixir
  def generate_no_task_with_to_list(sum) do
    for x <- Enum.to_list(1..sum),
        y <- if((sum - 2 * x) > x, do: Enum.to_list(x..(sum - 2 * x)), else: []),
        z <- if((sum - x - y) != 0, do: [sum - x - y], else: []),
        pythagorean?([x, y, z]),
        do: [x, y, z]
  end

  def generate_no_task_no_to_list(sum) do
    for x <- 1..sum,
        y <- if((sum - 2 * x) > x, do: x..(sum - 2 * x), else: []),
        z <- if((sum - x - y) != 0, do: [sum - x - y], else: []),
        pythagorean?([x, y, z]),
        do: [x, y, z]
  end

  def generate_with_task_with_to_list(sum) do
    Enum.map(sum..1, fn x ->
      Task.async(fn ->
        for y <- if(sum - 2 * x > x, do: Enum.to_list(x..(sum - 2 * x)), else: []),
            z <- if(sum - x - y != 0, do: [sum - x - y], else: []),
            pythagorean?([x, y, z]),
            do: [x, y, z]
      end)
    end)
    |> await()
    |> Enum.reduce([], fn list, acc -> list ++ acc end)
  end

  def generate_with_task_no_to_list(sum) do
    Enum.map(sum..1, fn x ->
      Task.async(fn ->
        for y <- if(sum - 2 * x > x, do: x..(sum - 2 * x), else: []),
            z <- if(sum - x - y != 0, do: [sum - x - y], else: []),
            pythagorean?([x, y, z]),
            do: [x, y, z]
      end)
    end)
    |> await()
    |> Enum.reduce([], fn list, acc -> list ++ acc end)
  end
```

Those are the results:
```
Name                                         ips        average  deviation         median         99th %
generate - with task, no to list          0.0888        11.26 s     ±0.00%        11.26 s        11.26 s
generate - with task, with to list        0.0742        13.49 s     ±0.00%        13.49 s        13.49 s
generate - no task, no to list            0.0362        27.61 s     ±0.00%        27.61 s        27.61 s
generate - no task, with to list          0.0293        34.14 s     ±0.00%        34.14 s        34.14 s

Comparison: 
generate - with task, no to list          0.0888
generate - with task, with to list        0.0742 - 1.20x slower +2.23 s
generate - no task, no to list            0.0362 - 2.45x slower +16.35 s
generate - no task, with to list          0.0293 - 3.03x slower +22.88 s
```

neenjaw (approved):
Nice benchmarks

PR #659, opened 2021-02-21
Bring `word-count` up to date with canonical data

angelikatyborska:
7 new tests from the canonical data, 2 new tests from me to help mentors out with noticing unhandled corner cases ("with quotations at the end at the beginning of the sentence" and "Polish").

The new tests from canonical data add a lot of complexity by testing apostrophes and quotations. I recall this exercise used to do that in the past because I had an old solution submitted that covered those test cases, but then it was simplified. I couldn't find why it was simplified. My guess is that because it was chosen as a core exercise. If that's the case, I think it can go back to being more complex 😅.


neenjaw:
I'm okay with adding these in to cover these cases, there is a simpler solution though than the multistep regex replace by using the `\b` metacharacter match for a word boundary (thanks @iHiD for schooling me on this in ruby 😄), I've opened a PR against this pr with my solution to use if you wish. #660

neenjaw (approved):
Looks good, only suggestion is to consider a different solution. Feel free to iterate on it if you think it needs tweaking.

PR #660, opened 2021-02-21
word-count-solution-suggestion

neenjaw:
Was looking over this and I remembered some other tracks have a different way of dealing with canonical spec test cases.  Thoughts?

### DIFF
--- a/exercises/practice/word-count/.meta/example.ex
+++ b/exercises/practice/word-count/.meta/example.ex
@@ -9,7 +9,7 @@ defmodule WordCount do
     sentence
     |> normalize()
     |> split
-    |> Enum.frequencies()
+    |> frequencies()
   end
 
   defp normalize(string) do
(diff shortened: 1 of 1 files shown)


neenjaw:
Feel free to iterate, merge, or close as you wish :)

neenjaw:
Right, `Enum.frequencies` isn't available before `1.10`. I guess two ways to go about it:

1) We could drop pre-1.10 support ( did a reasonable search of distro packages, and most are pointing to 1.11 by default now ), simplifying our CI runs
2) add in the previous `reduce > Map.update` step back in

angelikatyborska:
Amazing 🤩 I wish my regex game was so strong. I will gladly accept this clearly superior solution, and for the sake of the students I'm very glad this is possible.

Elixir 1.10 is only 1 year old, I don't feel comfortable dropping support for < 1.10 just yet. Right now, we're starting testing from 1.8, which was released in January 2019, a mere 2 years ago. I feel like that already is pretty young.

I tweaked the solution for < 1.10.

PR #664, opened 2021-02-23
Added 'forth' example solution which passes all tests

frerich:
This adds an example solution to `exercises/practice/forth/.meta/example.ex` which passes all tests.


angelikatyborska:
Amazing 🤩 thank you! I'll merge and take care of the CI failures.

PR #666, opened 2021-02-23
Bring `palindrome-products` up to date with canonical data (tests passing) 

MeerKatDev:
```
PalindromeProductsTest [test/palindrome_products_test.exs]
  * test largest palindrome from single digit factors (4.3ms) [L#12]
  * test smallest palindrome from triple digit factors (129.4ms) [L#33]
  * test largest palindrome from triple digit factors (125.7ms) [L#40]
  * test empty result if no palindrome in the range, smaller range (0.00ms) [L#67]
  * test largest palindrome from double digit factors (1.3ms) [L#26]
  * test smallest palindrome from double digit factors (1.0ms) [L#19]
  * test largest palindrome from four digit factors (17973.3ms) [L#54]
  * test smallest palindrome from four digit factors (18231.1ms) [L#47]
  * test raises argument error if max < min (2.3ms) [L#73]
  * test smallest palindrome from single digit factors (0.03ms) [L#5]
  * test empty result if no palindrome in the range (0.00ms) [L#61]


Finished in 36.5 seconds
11 tests, 0 failures
```

Taken from [here](https://exercism.io/tracks/elixir/exercises/palindrome-products/solutions/f307194ada734640a2777328915688b1), I changed only `palindrome?`

### DIFF
--- a/exercises/practice/palindrome-products/.meta/example.ex
+++ b/exercises/practice/palindrome-products/.meta/example.ex
@@ -2,27 +2,25 @@ defmodule PalindromeProducts do
   @doc """
   Generates all palindrome products from an optionally given min factor (or 1) to a given max factor.
   """
-  @spec generate(non_neg_integer) :: map
   @spec generate(non_neg_integer, non_neg_integer) :: map
   def generate(max_factor, min_factor \\ 1)
-  def generate(max_factor, min_factor) when max_factor < min_factor do
-    raise ArgumentError
-  end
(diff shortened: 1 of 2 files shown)


angelikatyborska:
Awesome! That's much faster than the current solution. 

Could you run `mix format` on your code and then move the changes to `.meta/example.ex`? `lib/palindrome_products.ex` is the boilerplate that the student downloads, so it need to remain like it was.

PR #668, opened 2021-02-24
Added example solution to 'wordy' exercise

frerich:


angelikatyborska:
Closing because we got two PRs for the same thing, and the other PR was faster 😅 https://github.com/exercism/elixir/pull/667

But your work is definitely not wasted because after v3 launch you will have one of the very few `wordy` solutions that still pass the tests 😎

PR #677, opened 2021-02-27
Run doctests on CI

angelikatyborska:
I noticed that we have some doctests that are never run, and thus they become out of date. We cannot include doctests in the test files downloaded by students, but we can include them when running tests on the CI.

This tiny change will also catch instances of the test module and lib module having different names, as it extracts the module-under-test name from the test module name.

Failures at the moment:
```
10 tests failing:
 - binary-search
 - difference-of-squares
 - etl
 - isbn-verifier
 - largest-series-product
 - markdown
 - nth-prime
 - rna-transcription
 - space-age
 - transpose
```

I will fix them in this PR.

### DIFF
--- a/.github/workflows/pr.ci.yml
+++ b/.github/workflows/pr.ci.yml
@@ -56,13 +56,8 @@ jobs:
           mkdir -p priv/plts
           mix dialyzer --plt
 
-        # Replace <code-extensions> with file extensions that should trigger this check. Replace with .* to allow anything.
-      - name: Run exercism/elixir ci pre-check (stub files, config integrity) for changed exercises
-        run: |
-          PULL_REQUEST_URL=$(jq -r ".pull_request.url" "$GITHUB_EVENT_PATH")
-          curl --url $"${PULL_REQUEST_URL}/files" --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' | \
-            jq -c '.[] | select(.status == "added" or .status == "modified") | select(.filename | match("\\.(ex|exs|md|json)$")) | .filename' | \
-            xargs -r bin/pr-check.sh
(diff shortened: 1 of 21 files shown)


.github/workflows/pr.ci.yml
```diff
@@ -56,13 +56,8 @@ jobs:
           mkdir -p priv/plts
           mix dialyzer --plt
 
-        # Replace <code-extensions> with file extensions that should trigger this check. Replace with .* to allow anything.
-      - name: Run exercism/elixir ci pre-check (stub files, config integrity) for changed exercises
-        run: |
-          PULL_REQUEST_URL=$(jq -r ".pull_request.url" "$GITHUB_EVENT_PATH")
-          curl --url $"${PULL_REQUEST_URL}/files" --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' | \
-            jq -c '.[] | select(.status == "added" or .status == "modified") | select(.filename | match("\\.(ex|exs|md|json)$")) | .filename' | \
-            xargs -r bin/pr-check.sh
```
angelikatyborska, line 65: IMO this check doesn't make much sense because the scripts `bin/pr-check.sh` and `bin/pr.sh` don't have an option to only work for selected exercises or files, they always check all of the files when run.

Right now, `bin/pr-check.sh` and `bin/pr.sh` won't run if no `ex|exs|md|json` were changed, which means they won't run if the scripts themselves were changed! Like in the first commit in this PR.

I think the most reasonable and safe choice that doesn't require a lot of extra work is to always run all checks.

angelikatyborska:
I just realized that my changes will only verify doctests from example solution. It won't actually find broken doctests in the empty solutions that students download 🤦. Still, better than than nothing? I'm not sure how to ensure that doctests in empty solutions are correct.

neenjaw (approved):

PR #678, opened 2021-02-27
Update setup-elixir action

angelikatyborska:
`actions/setup-elixir` has been deprecated in favor of the [`erlef/setup-elixir`](https://github.com/erlef/setup-elixir) action.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Use Elixir
-        uses: actions/setup-elixir@v1
+        uses: erlef/setup-elixir@v1
         with:
           otp-version: ${{matrix.otp}}
           elixir-version: ${{matrix.elixir}}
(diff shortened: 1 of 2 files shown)


NobbZ (commented):
Is manually installing `hex` and `rebar` in the dependency installation step really necessary?

The `erlef/setup-elixir` has the following options:

https://github.com/erlef/setup-elixir/blob/main/action.yml#L12-L17

Anyway, it seems as if the SSL version in the `ubuntu-latest` isn't supported by OTP 19.

Therefore I am not sure if either would work at all.

Though when elixir 1.12 gets released, 1.6 will be end of life according to regular policy.

Sadly I was unable to find an official policy about OTP end of life, though I found some announcement that was [about 21.3 reaching EOL](https://groups.google.com/g/rabbitmq-users/c/v3K5nZNsfwM). Therefore I think we can safely assume 19 and 20 are EOL as well.

Is there a need to support this meany versions?

neenjaw:
> Is there a need to support this meany versions?

I don't think so, personally.

angelikatyborska:
I exchanged OTP 19 with OTP 20 and now everything works 🎉

neenjaw (approved):

PR #679, opened 2021-03-01
Add fetch_configlet.sh

angelikatyborska:
Copied from https://github.com/exercism/github-actions/blob/main/configlet-ci/fetch-configlet, added a case statement to allow choosing the platform.

To be used by maintainers manually, when they want to lint the track before submitting a PR, for example.

### DIFF
--- a/bin/fetch_configlet.sh
+++ b/bin/fetch_configlet.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+set -euo pipefail
+
+curlopts=(
+  --silent
+  --show-error
+  --fail
+  --location
+  --retry 3
+)
+
+get_download_url() {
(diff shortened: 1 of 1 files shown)


ErikSchierboom (commented):
@angelikatyborska While using the GHA script works with this modification, we actually have a script version for tracks: https://github.com/exercism/configlet/blob/main/scripts/fetch-configlet

That script has the platform option built-in as well as some other features.

angelikatyborska:
Perfect! I guess I didn't look hard enough for it :D

ErikSchierboom (approved):
Great!

ErikSchierboom:
> Perfect! I guess I didn't look hard enough for it :D

Well, you'd have to know, so it is more on our side :)

PR #683, opened 2021-03-05
Update bob stub

angelikatyborska:
I didn't notice this before, but for some reason `bob`'s stub is the only one that hardcodes the usage of `cond` and uses `raise`. All of our other stubs for practice exercises leave out the functions completely empty. I don't see why `bob` should be different 🤷

### DIFF
--- a/exercises/practice/bob/.meta/example.ex
+++ b/exercises/practice/bob/.meta/example.ex
@@ -1,22 +1,5 @@
 defmodule Bob do
-  @doc """
-  Answers to `hey` like a teenager.
-
-  ## Examples
-
-    iex> Bob.hey("")
-    "Fine. Be that way!"
-
-    iex> Bob.hey("Do you like math?")
-    "Sure."
-
-    iex> Bob.hey("HELLO!")
-    "Whoa, chill out!"
-
-    iex> Bob.hey("Coding is cool.")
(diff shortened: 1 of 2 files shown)


neenjaw (approved):

exercises/practice/bob/.meta/example.ex
```diff
@@ -1,22 +1,5 @@
...
-    "Sure."
-
-    iex> Bob.hey("HELLO!")
-    "Whoa, chill out!"
-
-    iex> Bob.hey("Coding is cool.")
-    "Whatever."
-  """
-
+  @spec String.t() :: String.t()
```
neenjaw, line 2:
```suggestion
  @spec hey(String.t()) :: String.t()
```

exercises/practice/bob/lib/bob.ex
```diff
@@ -1,7 +1,5 @@
 defmodule Bob do
+  @spec String.t() :: String.t()
```
neenjaw, line 2:
```suggestion
  @spec hey(String.t()) :: String.t()
```

neenjaw:
Good idea, I fixed the `@spec` which was missing the function name. 😄

angelikatyborska:
Oops :D thanks a lot!

PR #689, opened 2021-03-14
fix spec

herminiotorres:
The square method waits for a tuple with an atom and an integer or a tuple with an atom and a string on the test.
And for the total method waits for a tuple with an atom and an integer.

### DIFF
--- a/exercises/practice/grains/lib/grains.ex
+++ b/exercises/practice/grains/lib/grains.ex
@@ -2,14 +2,14 @@ defmodule Grains do
   @doc """
   Calculate two to the power of the input minus one.
   """
-  @spec square(pos_integer) :: {:ok, pos_integer} | {:error, String.t()}
+  @spec square(pos_integer()) :: {:ok, pos_integer()} | {:error, String.t()}
   def square(number) do
   end
 
   @doc """
   Adds square of each number from 1 to 64.
   """
-  @spec total :: {atom, pos_integer}
+  @spec total :: {:ok, pos_integer()}
   def total do
   end
 end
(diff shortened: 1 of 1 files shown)


neenjaw:
Do you think it might be more descriptive to have it `{:ok, pos_integer()} | {:error, String.t()}`?

herminiotorres:
> Do you think it might be more descriptive to have it `{:ok, pos_integer()} | {:error, String.t()}`?

Good point, I will send this changing, thanks to the point that!

exercises/practice/grains/lib/grains.ex
```diff
@@ -2,14 +2,14 @@ defmodule Grains do
...
-  @spec square(pos_integer) :: pos_integer
+  @spec square(pos_integer) :: {:ok, pos_integer} | {:error, String.t()}
   def square(number) do
   end
 
   @doc """
   Adds square of each number from 1 to 64.
   """
-  @spec total :: pos_integer
+  @spec total :: {atom, pos_integer}
```
neenjaw, line 12:
```suggestion
  @spec total :: {:ok, pos_integer()}
```

neenjaw (changes requested):

exercises/practice/grains/lib/grains.ex
neenjaw: Maybe the same thing here?

neenjaw:
Also, with concept exercises (which are new) we have made a conscious choice to use the parentheses with typespecs since that is what you see in [official documentation](https://hexdocs.pm/elixir/typespecs.html).  They are also present in ~40% of the practice exercises (from what I saw doing a quick grep of the source files). So please add them.

We should probably go through and normalize them across all of the exercises.

neenjaw (approved):

neenjaw:
Nice work! Thanks for this patch, really appreciate your help to make the track a better place!! 💙

neenjaw:
Once v3 is released you'll get credit for this PR on the website too!

PR #693, opened 2021-03-20
New concept exercise `city-office` (`docs` and `typespecs`)

angelikatyborska:
WIP!


neenjaw (commented):
🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 So good, great investigation to discover this method for testing. Love the exercise, I think it will be awesome.

Where were you thinking about having it in the track? early after we introduce major value types? Then we could go back an add in `@spec`s to other exercises.

angelikatyborska:
> Where were you thinking about having it in the track? early after we introduce major value types? 

The decision to include a struct in this exercise unfortunately forces rather late in the tree, after structs. That's not great, I might want to remove the `Address` struct and replace it with something else that deserves its own custom type, maybe map? Maps are much earlier in the concept tree.

> Then we could go back an add in @specs to other exercises.

I don't have a strong opinion for or against this, but if some exercises got specs, then they would have to require the typespecs concept. If we don't do that, then the concept remains entirely optional 🤔

neenjaw:
If you rebase on main, it will then drop the 1.6 ci check. 😄

angelikatyborska:
I'll sneak in some more reputation points in this PR 🐍 I think every PR that adds a whole exercise should a major one.

neenjaw (approved):

neenjaw:
Looks great!

PR #694, opened 2021-03-21
Drop 1.6 from CI matrix

neenjaw:
Introducing a concept exercise for documentation and typespecs requires use of a function not present in 1.6.

1.6 is reasonably old at this point and not part of any default package repo that I have been able to find, therefore I think it is safe to drop this version from being supported if it allows us to have a better concept course of study.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,10 +64,9 @@ jobs:
 
     strategy:
       matrix:
-        #Note: pick a canonical set of supported versions
-        elixir: ['1.6.0', '1.7.0']
-        otp: ['20.0']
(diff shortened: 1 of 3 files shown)


angelikatyborska:
It also needs an update in the file `.github/workflows/ci.yml` and in the README in the setup section

neenjaw:
Made sure to add you as a co-author for that commit for the other files.

angelikatyborska:
git voodoo 😮

angelikatyborska (approved):

neenjaw:
`git commit --amend` then just added the co-authored trailer :P

PR #699, opened 2021-03-28
shorten practices entries for `enum` and `strings`

neenjaw:
Guildline for a pratice exercise `practices` entry to only use the concept slug maximum 8-10 times accross the track.  the `prerequisites` key may still use the concept slug as many times as neccessary.

This PR reduces the number of times the enum and string concepts were being used across the practice exercises.

My understanding is that practices key is more for presentational use than how it affects progression.

### DIFF
--- a/config.json
+++ b/config.json
@@ -930,7 +930,7 @@
           "multiple-clause-functions",
           "binaries"
         ],
-        "practices": ["recursion", "lists"],
+        "practices": ["recursion"],
         "difficulty": 3
       },
       {
@@ -954,7 +954,7 @@
           "enum",
           "recursion"
         ],
-        "practices": ["enum", "recursion"],
+        "practices": ["enum"],
         "difficulty": 2
       },
       {
@@ -1211,7 +1211,7 @@
           "enum",
           "strings"
         ],
-        "practices": ["recursion", "tail-call-recursion"],
+        "practices": ["tail-call-recursion"],
         "difficulty": 3,
         "status": "deprecated"
       },
@@ -1292,7 +1292,7 @@
           "recursion",
           "ranges"
         ],
-        "practices": ["strings", "charlists"],
+        "practices": ["charlists"],
         "difficulty": 5
       },
       {
--- a/config.json
+++ b/config.json
@@ -470,7 +470,7 @@
           "structs",
           "default-arguments"
         ],
-        "practices": ["pattern-matching", "structs"],
+        "practices": ["structs", "multiple-clause-functions"],
         "difficulty": 6
       },
       {
@@ -484,7 +484,7 @@
           "multiple-clause-functions",
           "pattern-matching"
         ],
-        "practices": ["lists", "recursion", "tail-call-recursion"],
+        "practices": ["lists", "tail-call-recursion"],
         "difficulty": 6
       },
       {
@@ -502,7 +502,7 @@
           "case",
           "if"
         ],
-        "practices": ["strings", "regular-expressions"],
+        "practices": ["regular-expressions"],
         "difficulty": 5
       },
       {
@@ -527,7 +527,7 @@
           "pattern-matching",
           "multiple-clause-functions"
         ],
-        "practices": ["structs", "recursion"],
+        "practices": ["structs"],
         "difficulty": 8
       },
       {
@@ -849,7 +849,7 @@
           "maps",
           "ranges"
         ],
-        "practices": ["multiple-clause-functions", "recursion"],
+        "practices": ["multiple-clause-functions"],
         "difficulty": 3
       },
       {
@@ -862,7 +862,7 @@
           "multiple-clause-functions",
           "guards"
         ],
-        "practices": ["integers", "recursion", "guards"],
+        "practices": ["integers", "guards"],
         "difficulty": 2
       },
       {
@@ -910,7 +910,7 @@
           "guards",
           "nil"
         ],
-        "practices": ["lists", "recursion", "nil"],
+        "practices": ["lists", "nil"],
         "difficulty": 3
       },
       {
@@ -930,7 +930,7 @@
           "multiple-clause-functions",
           "binaries"
         ],
-        "practices": ["recursion", "pattern-matching"],
+        "practices": ["recursion", "lists"],
         "difficulty": 3
       },
       {
@@ -1004,7 +1004,7 @@
           "ranges",
           "module-attributes-as-constants"
         ],
-        "practices": ["guards", "maps", "module-attributes-as-constants"],
+        "practices": ["guards", "module-attributes-as-constants"],
         "difficulty": 3
       },
       {
@@ -1025,7 +1025,7 @@
           "errors",
           "exceptions"
         ],
-        "practices": ["integers", "recursion", "ranges", "errors"],
+        "practices": ["integers", "ranges", "errors"],
         "difficulty": 2
       },
       {
@@ -1114,11 +1114,7 @@
           "tail-call-recursion",
           "enum"
         ],
-        "practices": [
-          "integers",
-          "multiple-clause-functions",
-          "pattern-matching"
-        ],
+        "practices": ["integers", "multiple-clause-functions"],
         "difficulty": 2
       },
       {
@@ -1188,13 +1184,7 @@
           "recursion",
           "tail-call-recursion"
         ],
-        "practices": [
-          "atoms",
-          "tuples",
-          "pattern-matching",
-          "recursion",
-          "tail-call-recursion"
-        ],
+        "practices": ["atoms", "tuples", "tail-call-recursion"],
         "difficulty": 1
       },
       {
@@ -1248,7 +1238,7 @@
         "name": "Transpose",
         "uuid": "ce270a34-add1-422c-bb86-53b310f05e27",
         "prerequisites": ["strings", "enum"],
-        "practices": ["strings"],
+        "practices": ["enum"],
         "difficulty": 5
       },
       {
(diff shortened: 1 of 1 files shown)


angelikatyborska:
> Guildline for a pratice exercise practices entry to only use the concept slug maximum 8-10 times accross the track.

Where does that come from? I heard that we should try to find at least a few exercises that practice a concept, but not that we have to limit how many.

If we're limited to 10 exercises per concept, what is the strategy for choosing them? Only the easiest ones, only the hardest ones, or try to find a mixture of easy and hard?

If this isn't a set-in-stone decision yet, then I would personally not to be limited in this way, and if the number of exercises per concept is too high, I would expect the website to handle it somehow.

iHiD:
> Where does that come from

3-8 is recommended [in the docs](https://github.com/exercism/docs/blob/c44e6305bc1e7626c1c60b31bc9634271ac82d80/anatomy/tracks/practice-exercises.md#practices). I think a hard-cap at 10 will be the sensible number (which is where Tim got that from because I said that to him yesterday)

---

The main reason is to give the user an ability to get through the Concepts and feel a sense of completion, while leaving them with more exercises to "play" with afterwards. I feel like we want to make the idea of "Mastering a Concept" (doing its Concept Exercise and all it's "practices") something that you can do in ~10 exercises, not 30, otherwise I think it will be an overwhelming experience.

For me this is a key value in having the seperate `prerequisites` key and the `practices` key.

Open to discussion as ever though :)

angelikatyborska:
I see. I completely missed this requirement before 👀. I understand and I'll comply 🤓 
 
@iHiD can you also give your opinion on this question:

> If we're limited to 10 exercises per concept, what is the strategy for choosing them? Only the easiest ones, only the hardest ones, or try to find a mixture of easy and hard?

And a new question:

What if having to limit a concept to max 10 exercises will cause some exercises to have empty "practices"? Is that acceptable?

iHiD:
Thanks 💙 

> If we're limited to 10 exercises per concept, what is the strategy for choosing them? Only the easiest ones, only the hardest ones, or try to find a mixture of easy and hard?

I feel like a variety is best. Ideally I'd say things that use Strings (e.g.) in different ways. So maybe something that does building, something that iterates through strings, something that looks at replacing chars, etc. And a mixture of easy/hard would b great too 🙂 

> What if having to limit a concept to max 10 exercises will cause some exercises to have empty "practices"? Is that acceptable?

Yes. That's totally fine. Those will just get unlocked by the relevant pre-reqs but won't be associated with that Concept on the map. I think of them as "bonus exercises".

neenjaw:
My strategy for this PR? I did a regex search, first for both concepts, I removed them from the `practices` list for an exercise when that exercise had more than 3 concepts listed.  That got me to ~12-15 each.  For exercises that only practised `enum` or `strings` I left those lists alone. For exercises that had 1-3 concepts being practised it was just an (imperfect, I'm sure) judgement call.

angelikatyborska:
> My strategy for this PR? I did a regex search, first for both concepts, I removed them from the practices list for an exercise when that exercise had more than 3 concepts listed. That got me to ~12-15 each. For exercises that only practised enum or strings I left those lists alone. For exercises that had 1-3 concepts being practised it was just an (imperfect, I'm sure) judgement call.

I like that approach 👍 

## A short analysis of the PR

### Before the changes:

Strings had 30 "practices":
```
strings - 30
[
  'word-count',
  'bob',
  'beer-song',
  'markdown',
  'two-fer',
  'pig-latin',
  'protein-translation',
  'run-length-encoding',
  'series',
  'scrabble-score',
  'twelve-days',
  'matching-brackets',
  'anagram',
  'matrix',
  'phone-number',
  'isbn-verifier',
  'binary',
  'transpose',
  'ocr-numbers',
  'kindergarten-garden',
  'diamond',
  'luhn',
  'rail-fence-cipher',
  'hexadecimal',
  'parallel-letter-frequency',
  'simple-cipher',
  'crypto-square',
  'atbash-cipher',
  'grep',
  'alphametics'
]
```

Enum had 34 "practices":
```
enum - 34
[
       'rna-transcription',         'resistor-color',
        'nucleotide-count',              'raindrops',
                  'series',         'scrabble-score',
        'sum-of-multiples',                'isogram',
                 'anagram',                'hamming',
                  'matrix',             'tournament',
                  'grains',                 'change',
               'transpose',            'ocr-numbers',
     'kindergarten-garden',                'diamond',
                    'luhn',        'scale-generator',
       'rail-fence-cipher',        'perfect-numbers',
     'palindrome-products',       'pascals-triangle',
           'spiral-matrix',          'saddle-points',
                 'connect',            'minesweeper',
               'allergies',          'crypto-square',
  'largest-series-product',                  'poker',
   'difference-of-squares',            'alphametics'
]
```

### After the changes

string has 11 "practices"

```
strings - 11
[
  'markdown',
  'pig-latin',
  'run-length-encoding',
  'anagram',
  'phone-number',
  'transpose',
  'ocr-numbers',
  'kindergarten-garden',
  'crypto-square',
  'atbash-cipher',
  'alphametics'
]
```

enum has 8 "practices"

```
enum - 8
[
  'rna-transcription',
  'resistor-color',
  'raindrops',
  'sum-of-multiples',
  'anagram',
  'hamming',
  'connect',
  'minesweeper'
]
```

### Conclusion for this PR (my review)

#### Strings

I really like which exercises are leftover for "strings", but we need to remove 3 more to meet the 8 threshold. I would suggest removing:
- `markdown` - it's very hard and we could keep it as a regex exercise
- `transpose` - it's much more enum-heavy than string-heavy (we can either bring it back to practice "enum" or leave empty prerequisites?)
- `kindergarten-garten` - no hard feelings against it, but something else has to be removed, and this one is IMO better suited for practicing maps than strings

#### Enum

Looks good, but maybe we can do a bit better? Add more variety? Here are my suggestions (totally optional, feel free to ignore if you don't agree):
- remove `anagram` and let it be a string-only exercise (although it's a bit boring for both)
- remove `rna-transcription`, keep it charlist-only because the enum component is extremely simple and easily replaced with recursion
- add something more fun/complex, e.g. `diamond`, `scale-generator`, `saddle-points`

### Conclusion for a new PR

There are still 6 concepts that are above the 8 exercise threshold. @neenjaw do you want to take care of that or should I?

```
integers - 9
[
  'collatz-conjecture',
  'nth-prime',
  'leap',
  'all-your-base',
  'change',
  'luhn',
  'armstrong-numbers',
  'largest-series-product',
  'clock'
]

maps - 10
[
  'nucleotide-count',
  'scrabble-score',
  'grade-school',
  'matrix',
  'say',
  'tournament',
  'binary-search-tree',
  'kindergarten-garden',
  'custom-set',
  'saddle-points'
]

pattern-matching - 16
[
  'robot-simulator',
  'bowling',
  'protein-translation',
  'triangle',
  'matching-brackets',
  'tournament',
  'all-your-base',
  'meetup',
  'simple-linked-list',
  'ocr-numbers',
  'perfect-numbers',
  'prime-factors',
  'queen-attack',
  'poker',
  'alphametics',
  'wordy'
]

ranges - 13
[
  'rotational-cipher',
  'pangram',
  'sum-of-multiples',
  'nth-prime',
  'isbn-verifier',
  'grains',
  'diamond',
  'rail-fence-cipher',
  'hexadecimal',
  'diffie-hellman',
  'palindrome-products',
  'pascals-triangle',
  'difference-of-squares'
]

recursion - 21
[
      'roman-numerals',           'list-ops',
              'zipper',             'strain',
          'accumulate',        'twelve-days',
  'collatz-conjecture',      'flatten-array',
   'matching-brackets',            'hamming',
           'nth-prime',      'binary-search',
  'binary-search-tree', 'simple-linked-list',
              'binary',    'perfect-numbers',
       'prime-factors',   'pascals-triangle',
       'spiral-matrix',              'sieve',
            'dominoes'
]

regular-expressions - 11
[
  'word-count',
  'markdown',
  'forth',
  'acronym',
  'run-length-encoding',
  'isogram',
  'phone-number',
  'isbn-verifier',
  'parallel-letter-frequency',
  'grep',
  'wordy'
]
```

neenjaw:
👍🏻 I can clean up those other exercises too.  Good catch. I'll work your suggestions in.

iHiD:
(I think the "hard limit" will be 10 and the recommended range 3-8, so `kindergarten-garten` could stay if you wanted 🙂)

neenjaw:
I think I've resolved the occurance

config.json
```diff
@@ -930,7 +930,7 @@
           "multiple-clause-functions",
           "binaries"
         ],
-        "practices": ["recursion", "pattern-matching"],
+        "practices": ["recursion", "lists"],
```
angelikatyborska, line 933: Is the `lists` there on purpose? They're not even in `prerequisites`.

config.json
```diff
@@ -1306,7 +1292,7 @@
           "recursion",
           "ranges"
         ],
-        "practices": ["charlists", "ranges"],
+        "practices": ["strings", "charlists"],
```
angelikatyborska, line 1295: I'm not sure if this adds `strings` on purpose or by accident. My suggestion was to make this an `enum` exercise. The example solution doesn't even use a single `String` function, only `Enum.join`.

iHiD:
(I'm unsubscribing. Ping me again if you need anything 🙂)

angelikatyborska (commented):
There's still 2 too many occurrences of `recursion`:
```
recursion - 12
[
  'roman-numerals',
  'strain',
  'accumulate',
  'matching-brackets',
  'hamming',
  'binary-search',
  'binary-search-tree',
  'binary',
  'prime-factors',
  'pascals-triangle',
  'spiral-matrix',
  'sieve'
]
```

 I think `hamming` might be a good candidate for removal because the example solution doesn't even use recursion (how did I arrive that this is a recursion exercise? not sure...). Also `binary` because it's deprecated.

neenjaw:
Ya, I added those because I thought that it was a better alternative to ranges in the moment.  I don't really have strong feelings about this. So those two additions are removed.

angelikatyborska (approved):

PR #700, opened 2021-03-28
Add test for `list-ops`

marcandre:
I noticed the test suite did not catch an error I made because all tests have functions that return actual booleans.

### DIFF
--- a/exercises/practice/list-ops/test/list_ops_test.exs
+++ b/exercises/practice/list-ops/test/list_ops_test.exs
@@ -76,6 +76,11 @@ defmodule ListOpsTest do
     test "huge list" do
       assert L.filter(Enum.to_list(1..1_000_000), &odd?/1) == Enum.map(1..500_000, &(&1 * 2 - 1))
     end
+
+    @tag :pending
+    test "truthy values filter the list" do
+      assert L.filter([true, false, nil, 0, 1, ""], & &1) == [true, 0, 1, ""]
+    end
   end
 
   describe "foldl" do


exercises/practice/list-ops/test/list_ops_test.exs
```diff
@@ -76,6 +76,11 @@ defmodule ListOpsTest do
     test "huge list" do
       assert L.filter(Enum.to_list(1..1_000_000), &odd?/1) == Enum.map(1..500_000, &(&1 * 2 - 1))
     end
+
+    @tag :pending
+    test "non boolean results" do
```
neenjaw, line 81:
```suggestion
    test "truthy values filter the list" do
```

neenjaw (changes requested):
Hey! Awesome contribution! Thanks!

Do you think we could rename the test to better describe what behaviour it is testing for? I left a suggestion, but doesn't have to be exactly that if there is something better that you can come up with.

exercises/practice/list-ops/test/list_ops_test.exs
marcandre: Sure, sounds good. Done!

neenjaw: Cool, this looks good, there is one check failing, can you run `mix format` from that exercise directory?

neenjaw (approved):
Looks good, I can merge it once the checks pass

marcandre:
Format fixed

neenjaw:
Awesome, thanks for making the track a better place!

marcandre:
Thanks for the rapid feedback, and for your contributions to this great track!

PR #701, opened 2021-03-31
Use `Map` instead of `List` to store vertices in `dot-dsl`

marcandre:
I enjoyed doing the `dot-dsl` exercise quite a bit.

I realize the focus was on macro & ast handling, not on the actual graph, but I thought it was quite odd that the nodes of a graph would be stored in a list, and not in a map.

The test suite doesn't check what happens if a node is specified more than once, but it still makes no sense to have a graph with multiple nodes with the same label. By contrast, duplicate edges are not an issue.

It also seems natural that any actual use of the graph would require efficient lookup of nodes.

I don't know if that kind of trivial change is worth it or not, but I thought I'd submit it nevertheless. I won't mind if the powers that be prefer to keep the exercise as is.

### DIFF
--- a/exercises/practice/dot-dsl/.meta/example.ex
+++ b/exercises/practice/dot-dsl/.meta/example.ex
@@ -1,5 +1,5 @@
 defmodule Graph do
-  defstruct attrs: [], nodes: [], edges: []
+  defstruct attrs: [], nodes: %{}, edges: []
 end
 
 defmodule Dot do
@@ -9,10 +9,10 @@ defmodule Dot do
   defmacro graph(do: ast) do
     g = do_graph(ast)
 
-    Macro.escape(%Graph{
-      attrs: Enum.sort(g.attrs),
-      nodes: Enum.sort(g.nodes),
-      edges: Enum.sort(g.edges)
+    Macro.escape(%{
+      g
+      | attrs: Enum.sort(g.attrs),
+        edges: Enum.sort(g.edges)
     })
   end
 
@@ -37,13 +37,13 @@ defmodule Dot do
   end
 
   defp do_stmt({atom, _, nil}, g) when is_atom(atom) and atom != :-- do
-    %{g | nodes: [{atom, []} | g.nodes]}
+    %{g | nodes: Map.put(g.nodes, atom, [])}
   end
 
   defp do_stmt(stmt = {atom, _, [kws]}, g)
        when is_atom(atom) and atom != :-- and is_list(kws) do
     if Keyword.keyword?(kws) do
-      %{g | nodes: [{atom, kws} | g.nodes]}
+      %{g | nodes: Map.put(g.nodes, atom, kws)}
     else
       raise_invalid_stmt(stmt)
     end
--- a/exercises/practice/dot-dsl/lib/graph.ex
+++ b/exercises/practice/dot-dsl/lib/graph.ex
@@ -1,3 +1,3 @@
 defmodule Graph do
-  defstruct attrs: [], nodes: [], edges: []
+  defstruct attrs: [], nodes: %{}, edges: []
 end
--- a/exercises/practice/dot-dsl/test/dot_test.exs
+++ b/exercises/practice/dot-dsl/test/dot_test.exs
@@ -25,7 +25,7 @@ defmodule DotTest do
 
   @tag :pending
   test "graph with one node" do
-    assert %Graph{nodes: [{:a, []}]} ==
+    assert %Graph{nodes: %{a: []}} ==
              exprt(
                Dot.graph do
                  a
@@ -35,7 +35,7 @@ defmodule DotTest do
 
   @tag :pending
   test "graph with one node with keywords" do
-    assert %Graph{nodes: [{:a, [color: :green]}]} ==
+    assert %Graph{nodes: %{a: [color: :green]}} ==
              exprt(
                Dot.graph do
                  a(color: :green)
@@ -67,7 +67,7 @@ defmodule DotTest do
   test "graph with attributes" do
     assert %Graph{
              attrs: [bar: true, foo: 1, title: "Testing Attrs"],
-             nodes: [{:a, [color: :green]}, {:b, [label: "Beta!"]}, {:c, []}],
+             nodes: %{a: [color: :green], b: [label: "Beta!"], c: []},
              edges: [{:a, :b, [color: :blue]}, {:b, :c, []}]
            } ==
              exprt(


neenjaw:
Thanks for submitting this PR, I think it is important to try to improve the exercises, as they aren't perfect by any means.

So Elixir, Python, and Rust are the only tracks that have this exercise (from a brief and limited search), and all of them currently store the nodes in a list.  That's not to say that it is the best choice, so I tried to experiment a bit with graphviz to see if it is even legal syntax to name a single node twice -- and it turns out it is: https://bit.ly/2O9gFEz (online graphviz playground).

I'm not sure what the right choice is here, but with the example solution changes, it has the potential to lose information because a repeating node key will overwrite a node value. I think to do this implementation correctly you would need to do some sort of value merging if the node key is repeated, and that might raise the bar to high for a single exercise.  🤷🏻 I think, like you acknowledged, the focus of this exercise is on Macros and ast handling so I don't want to rob of that experience with too much more complexity.

That all said, I think it would be useful if other could share their thoughts on this potential advancement of this problem.

cc @angelikatyborska @NobbZ @devonestes @Cohen-Carlisle

neenjaw:
Oh, one thing that might help with the conversation if this exercise is already too hard: @iHiD is there a metric to see how many students have attempted this solution to how many have completed it?

marcandre:
Indeed, in my solution I did not attempt to merge the keywords (as they probably should be). It wouldn't be very difficult to do though. Just to be clear, my point isn't about the solution, but about the exercise itself, which forces a solution from the wrong model (imo).

It seems acceptable that an exercise doesn't test all cases (e.g. repeated nodes), or if the test is a bit too strict (e.g. it requires sorting the attributes), but having the wrong model feels different (and easily fixed).

BTW, the only exercises rated "hard" are:
* `zipper`: 338 published solutions
* `forth`: 176 published solutions
* `bowling`: 176 published solutions
* `dot-desl`: 88 published solutions

`dot-dsl` comes much later in the list of exercises though. In any case, it's difficult to conclude anything from these numbers (e.g. I didn't do `bowling` because it doesn't seem challenging enough...)

marcandre:
Here's [an additional test](https://github.com/marcandre/elixir/pull/1/files) for a repeated node and the small required change to the solution.

angelikatyborska:
@marcandre Could you provide sources for this argument?

> but it still makes no sense to have a graph with multiple nodes with the same label

I would say quite the opposite, I don't see a reason not to allow multiple nodes with the same label. But my computer science knowledge is limited, so I might be totally wrong.

marcandre:
Maybe I should have used "id" instead of "label". The "id" has to be unique. If we say there's an edge from "a" to "b", the "a" and "b" must refer *uniquely* to a node.

In Tim's example, the node "start" is used twice, but there's only one vertex in the resulting graph (as it should be).

angelikatyborska:
Understood 👍 Indeed, _id_ is a better name. But still, the DOT language does allow declaring the same node (with the same id) or the same edge multiple times and it's not a syntax error. If we switched to a map, we would not only lose the information about duplicated node declarations, but also about the declaration order. I still think for those reasons a list/keyword list is a better choice. The better choice when trying to describe the DOT language, maybe not the better choice when designing your own custom DSL for graphs :). But my assumption is that this exercise is about mimicking the DOT language, not about teaching graphs or setting good practices for working with graphs.

marcandre:
> If we switched to a map, we would not only lose the information about duplicated node declarations, but also about the declaration order

If instead of using `Map.put` we use `Map.update` (as [I show here](https://github.com/marcandre/elixir/pull/1/files)), then we loose no information and we actually get to merge it properly.
The declaration order is not important in a graph. For example, I changed modified Tim's example [by putting "start" at the end](https://bit.ly/2PpqvT8) and you obtain the same graph.

angelikatyborska:
> The declaration order is not important in a graph.

Again, I think we disagree about the point of the exercise. For me, this exercise is about representing the DOT language, not _any_ graph. And the declarations in the dot files have an order 🤷 

I'll let @neenjaw make a decision about this, but my opinion remains unchanged (keep keyword list of nodes, add a test that duplicate node declarations are valid).

marcandre:
>  I think we disagree about the point of the exercise.

Ah, thanks for explaining. Yes, we disagree on that 😄 

I would point out that:
1) The exercise states "Our DSL is similar to the Graphviz dot language in that our DSL will be used to create *graph data structures*"
2) The current test already requires us to reorder the nodes, edges and attributes; order is already lost.

I'm off to bed 😴

neenjaw:
Woah! I missed lots of conversation! 😛 

> > I think we disagree about the point of the exercise.
> 
> Ah, thanks for explaining. Yes, we disagree on that

I think acknowledging the area of disagreement is good, then gives us an opportunity to find the solve to this.

> BTW, the only exercises rated "hard" are:
> 
> * `zipper`: 338 published solutions
> * `forth`: 176 published solutions
> * `bowling`: 202 published solutions
> * `dot-desl`: 88 published solutions
> 
> `dot-dsl` comes much later in the list of exercises though. In any case, it's difficult to conclude anything from these numbers (e.g. I didn't do `bowling` because it doesn't seem challenging enough...) and they probably were not created at the same time.

Our difficulty rating has been an ongoing issue to get right (Many are currently wrongly classified). The difference between the first three in your list and the last is that the former all appear in the core track. Which is why I listed other difficult exercises (discovered by my anecdotal experience) to explore not the publishing rate but the attempted-but-gave-up rate.

---

At any rate, any solution to this needs to be a simplification/abstraction of the full dot language.  The question is about where to draw that balance between accuracy to the subject and abstraction to focus on the concept.  This is a question we have talked at length about when developing the new concept exercises.

I do think this exercise is difficult at present as it is.  I don't think this addition makes inordinately more difficult, but if this solution already has a low attempted:completed ratio, I would rather keep it as is.  I'm going to off-github have a conversation with @ihid about this and will come back with some data to support a conclusion. Sorry to keep you in a moment of limbo @marcandre

iHiD:
Sorry - my SQL didn't include some considerations for legacy data. I've updated the numbers above.

neenjaw:
Thanks @iHiD! Yes, those numbers are similar to my expectations from having solved them.  I think there are additional factors which influence completion rates (placement in list of exercises, whether or not they are core exercises), but I do think difficulty is a potential problem with this exercise.

So I am not inclined to make this problem any harder than it currently is, even if that means it has to suffer with a design problem where it uses a list rather than a map.

So I have an ask and a proposal:
1) @marcandre, are there ways you think we could simplify this problem? drop requirements, drop features, lower the bar
2) @marcandre, how do you feel about perhaps creating a followup exercise to a simplified DOT-DSL where we can incorporate your idea, once a person is exposed to metaprogramming and macros?  It would be similar to how we have `Resistor Color` and `Resistor Color Duo` (and even Trio on some tracks)

marcandre:
> So I have an ask and a proposal:
> 
>     1. @marcandre, are there ways you think we could simplify this problem? drop requirements, drop features, lower the bar

Are already not required:
- node repeated (this whole thread)
- auto-creation of nodes when referred to in an edge
- better matching (not dependent on things being sorted)

I think it is difficult to drop requirements this way.

What could be done is to provide a `Graph` module with an API to:
1) add an edge
2) specify a node, potentially with keywords
3) specify keywords for the graph
4) compare two graphs

This would make the exercise be more about the macro/AST and less about building the structure.

>     2. @marcandre, how do you feel about perhaps creating a followup exercise to a simplified DOT-DSL where we can incorporate your idea, once a person is exposed to metaprogramming and macros?  It would be similar to how we have `Resistor Color` and `Resistor Color Duo` (and even Trio on some tracks)

So maybe that Graph module could be an exercise in itself (I'd say difficulty = medium), maybe with a way to access a solved exercise for those that would only want to tackle the dot-dsl?

marcandre:
Or things can simply stay as they are. Even if some people might feel they are asked to write the wrong code, but it's clear it's all just for fun.

neenjaw:
Yes, I think you're right, it's hard to make the exercise easier by making the graph more trivial, and a good option to refocus on the Macro/AST is to do what you said, provide a complete rudimentary Graph module which then the student just needs to plug in to.

I think this is a really cool idea -- metaprogramming in Elixir is a reasonably popular topic because of excellent libraries (Ecto, Phoenix, Etc.) which make use of it (I've used it to write parts of our solution static analysis engine for elixir in v3).  It would be nice to give more exposure to that within the track.

> Or things can simply stay as they are. Even if some people might feel they are asked to write the wrong code, but it's clear it's all just for fun.

I hear ya, it is all just for fun. 🧑‍💻 🎉 I don't want to put you in a position to commit more time than you wanted to (with it just being a small commit that started this) but I do want to again acknowledge that I appreciate your engagement in this discussion because I do think there are areas that the track can improve.

I'm going to slate this on my to-do list, I don't have a very definite timeline for completion, but if you want to contribute to a re-write of this exercise I would accept your collaboration on it. If that's more than you signed up for, thats ok too. 💙

marcandre:
> It would be nice to give more exposure to that within the track.

I'm not sure I follow... Providing a basic `Graph` module wouldn't add any meta-programming to the `dot-dsl` exercise. It would just reduce a bit the non-meta-programming part.

> if you want to contribute to a re-write of this exercise

Just to be clear, the idea would be to keep the exercise more or less "as is", but with the student being able to use some provided functions to build the `%Graph`, right?
If so, I could definitely provide a PR for this

neenjaw:
> > It would be nice to give more exposure to that within the track.
> 
> I'm not sure I follow... Providing a basic `Graph` module wouldn't add any meta-programming to the `dot-dsl` exercise. It would just reduce a bit the non-meta-programming part.

My thought is that by providing the `Graph`, it makes the exercise more approachable, so that people people submit/complete it. 

> > if you want to contribute to a re-write of this exercise
> 
> Just to be clear, the idea would be to keep the exercise more or less "as is", but with the student being able to use some provided functions to build the `%Graph`, right?
> If so, I could definitely provide a PR for this

Yes, that is what I meant.

marcandre:
👍 , I hope to have a PR ready sometime next week.

marcandre:
@neenjaw I've opened #708 instead.

PR #708, opened 2021-04-08
Provide `Graph` API for `dot-dsl`

marcandre:
This PR proposes to provide a `Graph` API for the `dot-dsl` exercise (see discussion in #701 )


marcandre:
No idea why CI is failing

neenjaw:
🤩 wow! Awesome! Just has a quick peek and looks awesome.

I think it is failing unrelated to you. Configlet get updated async from this repo, so it's spec has probably changed. I can look at that part. I'll look more closely at things tonight.

marcandre:
One thing: I had to duplicate the `graph.ex` content into `.meta/example.ex`, not sure if there's a better way. Felt dirty copy-pasting that 😆

neenjaw:
@marcandre Excellent work, this is very nice. I made a few very small changes not directly on the Graph API at all.  I have updated the testing script to use the (relatively) new `.meta/config` file, so we don't need to blindly wipe the files in `lib` to test.  So now the duplicated `Graph` code in the `example.ex` file isn't neccessary.

I put out a ping to @ErikSchierboom about the configlet issue.  I think this is an unrelated issue to your changes, so if you are okay with the commit I added, this is good to merge.

neenjaw (approved):

marcandre:
Cool, glad you removed the duplication 👍

ErikSchierboom:
Yep, the linting error is an unrelated issue.

angelikatyborska:
@marcandre I see that you're doing an unrelated fix to the `phone.ex` file name. That's awesome! Could you also fix the corresponding test file name? It's `phone_test.exs` but should be `phone_number_test.exs` according to the module name defined in the file.

@neenjaw Did you check if those changes require updating `practices` and `prerequisites` for this exercise? My guess would be that we need to remove both `maps` and `keyword-lists` from `practices`.

ErikSchierboom:
@neenjaw The configlet issue should be fixed: https://github.com/exercism/configlet/releases/tag/4.0.0-alpha.11

marcandre:
> @marcandre I see that you're doing an unrelated fix to the `phone.ex` file name. That's awesome! Could you also fix the corresponding test file name? It's `phone_test.exs` but should be `phone_number_test.exs` according to the module name defined in the file.

That's @neenjaw's doing

neenjaw:
It is, I'll fix it

neenjaw:
Fixed!! (sorry for the delay) but this brought to light some fragility in our ci system which now leverages the exercise configurations to work more robustly.

neenjaw:
Thanks again @marcandre! This makes this exercise better!

angelikatyborska:
@neenjaw maybe I missed it, but I think you didn't reply to my question before merging the PR:

> @neenjaw Did you check if those changes require updating practices and prerequisites for this exercise? My guess would be that we need to remove both `maps` and `keyword-lists` from practices.

neenjaw:
Nope forgot about this comment as I was working through the other details.

neenjaw:
@angelikatyborska, I looked at the root config.  `maps` aren't in the pre-reqs or practices list. The exercise still involves some vestigial keyword list use, so I think it's okay to keep it as is.

angelikatyborska:
@neenjaw Ooops, I wrote `maps`, but meant `structs`. Agreed on the `keyword-lists`, but it looks to me like `structs` should be removed from `practices` because all struct-related operations are now done for the student in the new functions added to the graph module.

PR #709, opened 2021-04-09
Test test

neenjaw:


ee7:
I've pushed a new release of `configlet`.

If you restart CI on this PR, the `configlet lint` job should now pass. Sorry that it failed before.

See https://github.com/exercism/configlet/pull/272.

neenjaw:
No prob @ee7! it's kind of an edge case! Thanks for all your hard work!!

PR #714, opened 2021-04-11
[German Sysadmin] Make sentence slightly friendlier

iHiD:
This is me testing the new GHA comments flow 🎉 

I don't expect this PR to be merged (unless you like the change, in which case more rep for me++)

### DIFF
--- a/exercises/concept/german-sysadmin/.docs/instructions.md
+++ b/exercises/concept/german-sysadmin/.docs/instructions.md
@@ -35,7 +35,7 @@ There are 4 non-Latin characters in the German alphabet, and all of them have co
 | ü                | ue               |
 | ß                | ss               |
 
-Extend the `sanitize/1` function. It should substitute German characters according to the table. Assume all usernames are already downcase.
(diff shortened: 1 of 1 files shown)


angelikatyborska:
Interesting that this didn't work 🤔 I did a test earlier that did work (https://github.com/exercism/elixir/pull/713).

The output has no errors, just says:
> matched snippet ids: 

Which would suggest the action is fine, but configuration file is not doing what I wanted it to do.

angelikatyborska:
You could set `ACTIONS_STEP_DEBUG` as a secret in this repo and rerun the action, this would give us more logs.

neenjaw:
I think you have to pass the dot option to `Minimatch.match`: [docs](https://www.npmjs.com/package/minimatch#dot)

angelikatyborska:
I created an issue so that we don't forget: https://github.com/exercism/pr-commenter-action/issues/1 

I'll try to work on this soon.

angelikatyborska (approved):

iHiD:
Wahoo. I did Elixir! 🎉

PR #718, opened 2021-04-18
Remove all skips from concept exercise tests

angelikatyborska:
It's not yet documented in the docs repo, but concept exercise tests should not be skipped at all (https://exercism-team.slack.com/archives/CR91YFNG3/p1617566538207100). This matters for students solving the exercises on their own machines, not via the website.


neenjaw (approved):
🤷🏻

PR #719, opened 2021-04-25
Explain `defp apply_discount/2` in `freelancer-rates`

angelikatyborska:
As I was solving this exercise again, I noticed that there is this private function in the boilerplate:
```elixir
  defp apply_discount(before_discount, discount) do
    raise "Please implement the apply_discount/2 function"
  end
```

Since it's a private function, it has no tests. The exercise steps do not mention this function at all. If the student followed all of the steps, they would be left this with private function not implemented.

To nudge the student into implementing this private function and to explain why, I added a paragraph to step 2.

Alternatively, we could make the function public, add tests, and insert a new step between current step 1 and 2 that tells the student to implement the function.

What do you think?

### DIFF
--- a/exercises/concept/freelancer-rates/.docs/hints.md
+++ b/exercises/concept/freelancer-rates/.docs/hints.md
@@ -9,12 +9,16 @@
 
 - [Basic arithmetic operations][kernel-arithmetic-operators] where one argument is an integer, and the other is a float, will return a float.
 
-## 2. Calculate the monthly rate, given an hourly rate and a discount
+## 2. Calculate a discounted price
+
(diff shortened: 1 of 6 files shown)


neenjaw:
I think we should either make it public or make it optional with something like `Applying a discount is an operation will also be needed in the next step. Implement it in a private function so that it can be reused.  A private function stub, `apply_discount/2`, has been provided for you.
`

angelikatyborska:
I'll make it public and add tests for it. I think concept exercises should be very clearly defined, without optional steps 🤔

neenjaw (approved):

PR #730, opened 2021-05-09
 Annotate every concept exercise test with @task_id

### LINKED ISSUE #729: Annotate every concept exercise test with @task_id
In theory, every test in a concept exercise's test suite belongs to a single task. They were already written in a way so that the tests are ordered by task.

Make this relationship explicit by annotating every test with `@task_id`, where the id is the task's number. 

This is optional. If there are concept exercises where some or all tests cannot be clearly assigned to a single task, tag them explicitly with `@task_id nil`.

Related test runner PR: https://github.com/exercism/elixir-test-runner/pull/36/files

angelikatyborska:
Closes #729

Approach: if there is a test that tests "everything together", I'm putting it as part of the last task. I am assuming that tasks need to be done in order, which is technically true in many exercises.

I skipped freelancer-rates and did it in this PR instead https://github.com/exercism/elixir/pull/719 because it adds a new task. I also skipped mensch-aergere-dich-nicht because it's deprecated.

### DIFF
--- a/exercises/concept/community-garden/test/community_garden_test.exs
+++ b/exercises/concept/community-garden/test/community_garden_test.exs
@@ -1,27 +1,32 @@
 defmodule CommunityGardenTest do
   use ExUnit.Case
 
+  @task_id 1
   test "start returns an alive pid" do
     assert {:ok, pid} = CommunityGarden.start()
     assert Process.alive?(pid)
   end
 
+  @task_id 2
   test "when started, the registry is empty" do
     assert {:ok, pid} = CommunityGarden.start()
     assert [] == CommunityGarden.list_registrations(pid)
   end
 
+  @task_id 3
   test "can register a new plot" do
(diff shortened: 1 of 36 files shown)


exercises/concept/community-garden/test/community_garden_test.exs
```diff
@@ -1,40 +1,40 @@
...
 
+  @task_id 4
   test "can release a plot" do
     assert {:ok, pid} = CommunityGarden.start()
     assert %Plot{} = plot = CommunityGarden.register(pid, "Johnny Appleseed")
     assert :ok = CommunityGarden.release(pid, plot.plot_id)
     assert [] == CommunityGarden.list_registrations(pid)
   end
 
-  test "can get a specific registration by id" do
```
angelikatyborska, line 32: I don't see a difference between this test and the next test. I think it might be an accident that there are two so similar tests?

neenjaw (approved):

PR #735, opened 2021-05-15
Make headings consistent in introductions

angelikatyborska:
Some of the existing `introduction.md` files didn't follow the heading pattern:
```
# Introduction

## Concept name

### Concept subheading
```

I'm fixing that in this PR.

It's important to note that all h1 are removed when the document is displayed on the website and replaced with "Introduction", thus following this pattern in all the files makes them more consistent with what you will see on the website.

### DIFF
--- a/exercises/concept/lasagna/.docs/introduction.md
+++ b/exercises/concept/lasagna/.docs/introduction.md
@@ -2,6 +2,8 @@
 
 ## Basics
 
+### Variables
+
 Elixir is a dynamically-typed language, meaning that the type of a variable is only checked at runtime. Using the match `=` operator, we can bind a value of any type to a variable name:
 
 ```elixir
--- a/exercises/concept/lasagna/.docs/introduction.md
+++ b/exercises/concept/lasagna/.docs/introduction.md
@@ -1,4 +1,6 @@
-# Basics
+# Introduction
+
+## Basics
 
 Elixir is a dynamically-typed language, meaning that the type of a variable is only checked at runtime. Using the match `=` operator, we can bind a value of any type to a variable name:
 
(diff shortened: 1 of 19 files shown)


angelikatyborska:
The CI has the same failure as in https://github.com/exercism/elixir/pull/734, that PR solves it.

neenjaw:
Where is the spec for this change? I can't find it in the docs.

angelikatyborska:
It's not really a new change, a lot of existing `introduction.md` files already follow this pattern. The structure of h2-h6 within a `introduction.md` is up to us, about h1 being removed it says here: https://github.com/exercism/docs/blob/main/building/markdown/markdown.md#L11-L13

neenjaw:
Yes, I already knew it was being removed, but by adding the concept name to a second level title, does it display on the website in these sections as expected? Or will it be displayed redundantly?

angelikatyborska:
By adding a concept name as h2 we will make sure that it will appear on the website. There won't be any duplication of heading because the website doesn't add concept name headings to concept _exercise_ introduction, it only adds a "Introduction" heading.

For example, before the PR, Lasagna has this `introduction.md`:
```
# Basics
(...)

## Modules
(...)
```

which on the website is rendered as (note the concept name is not there):
```
## Introduction
(...)

### Modules
(...)
```

Screen Shot 2021-05-16 at 16 43 54


After the PR, Lasagna will have this `introduction.md`:
```
# Introduction

## Basics
(...)

### Modules
(...)
```

which on the website will be rendered as:
```
## Introduction

### Basics
(...)

#### Modules
(...)
```

exercises/concept/lasagna/.docs/introduction.md
```diff
@@ -1,4 +1,6 @@
-# Basics
+# Introduction
+
+## Basics
```
neenjaw, line 3: So with this addition it makes the website look like this:

Screenshot from 2021-05-16 11-53-53

I don't think 'Basics' makes sense as the first h3 to describe the content.  If anything we should, depending on the exercise and introduction associated, make this:

```suggestion
## Basics

### Variables
```

or leave the concept name out and just use heading to describe the content

neenjaw (approved):
Made one suggestion comment, I like the idea overall and don't want to block you from merging when you think it is ready

exercises/concept/lasagna/.docs/introduction.md
angelikatyborska: I like the suggestion of adding a "Variables" heading 👍

PR #736, opened 2021-05-18
New concept exercise `need-for-speed` (`alias` and `import`)

angelikatyborska:
A new concept exercise that teaches people about `alias` and `import`.

PS: Please ignore the error in the branch name 🤦. Somehow I keep mixing up "import" and "require" when typing too fast.


exercises/concept/need-for-speed/test/need_for_speed_test.exs
```diff
@@ -0,0 +1,144 @@
...
+             Contestants:
+
+               1. #{IO.ANSI.green()}The Grasshopper#{IO.ANSI.default_color()}
+               Distance: 0 meters
+               Battery: Battery at 100%
+
+             """
+    end
+
+    @task_id nil
```
angelikatyborska, line 105: Task ids are nil because the test suite won't even run without having solved all of the tasks.

concepts/import/about.md
```diff
@@ -0,0 +1,19 @@
+# Import
+
+The [special form `import`][import] allows you to use functions from an outside module without using the module's name.
```
angelikatyborska, line 3:
```suggestion
The [special form `import`][import] allows you to use functions and macros from an outside module without using the module's name.
```

neenjaw (approved):
Looks great, I think this is probably the most straightforward way to teach these concepts

PR #737, opened 2021-05-19
Run concept exercise tests in predictable order

angelikatyborska:
Concept exercises do not have skips, so for a student on their own machine all tests will run at once, in random order. For learning purposes, it would be better if the tests always run in the same order as they're defined in in the file.

I asked Jeremy and Erik about this on Slack (https://exercism-team.slack.com/archives/CR91YFNG3/p1621353127004100) and they both confirmed that would be best.

For practice Exercises, I think it makes sense to keep the random order because:
- the student themselves control unskipping
- it's the Elixir way to randomize test order :)
- there's no right way of approaching an exercise, whereas concept exercise clearly have tasks in order


neenjaw (approved):
Looks good to me!

PR #749, opened 2021-06-05
Try a more precise cache key

angelikatyborska:
## The problem
Our CI check started randomly failing on commits unrelated to anything Elixir:

Screen Shot 2021-06-05 at 12 06 41

What changed? New patch versions of the langauges have been released between those commits.

### Last passing commit

Screen Shot 2021-06-05 at 12 08 06

### First failing commit

Screen Shot 2021-06-05 at 12 08 14

### The cause

Our old cache key creation method doesn't include patch version of the languages, and thus outdated PLTs are being used and causing the CLI failure:

Screen Shot 2021-06-05 at 11 31 02

## The solution

This PR uses exact Erlang and Elixir versions for the cache key. It's the same method I have been using with CircleCI with my other projects for 2 years now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,12 +29,22 @@ jobs:
           otp-version: ${{matrix.otp}}
           elixir-version: ${{matrix.elixir}}
 
+      - name: Set cache key
+        id: set_cache_key
+        run: |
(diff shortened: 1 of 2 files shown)


angelikatyborska:
WIP

neenjaw (approved):

PR #756, opened 2021-06-12
[#726] New practice exercise `satellite`

jiegillet:
One more. I really enjoyed coming up with a clean algorithm for this one!

The [metadata file](https://github.com/exercism/problem-specifications/tree/main/exercises/satellite) was missing a source and URL.

### DIFF
--- a/config.json
+++ b/config.json
@@ -2381,6 +2381,21 @@
         "difficulty": 9
       },
       {
+        "slug": "satellite",
+        "name": "Satellite",
+        "uuid": "a07e01e6-7cea-4db9-a514-7c95788e90a9",
+        "prerequisites": [
+          "pattern-matching",
+          "tuple",
+          "cond",
+          "recursion"
+        ],
+        "practices": [
+          "recusrion"
+        ],
+        "difficulty": 6
+      },
+      {
         "slug": "zebra-puzzle",
         "name": "Zebra Puzzle",
         "uuid": "0a561b08-6e61-4415-a3f1-e7d411a4a0e2",
--- a/exercises/practice/satellite/lib/satellite.ex
+++ b/exercises/practice/satellite/lib/satellite.ex
@@ -1,0 +1,14 @@
+defmodule Satellite do
+  @typedoc """
+  A tree, which can be empty, or made from a left branch, a node and a right branch
+  """
+  @type tree :: {} | {tree, any, tree}
+
+  @doc """
+  Build a tree from the elements given in a pre-order and in-order style
+  """
+  @spec build_tree(preorder :: [any], inorder :: [any]) :: tree | {:error, String.t()}
+
+  def build_tree(preorder, inorder) do
+  end
+end
--- a/exercises/practice/satellite/lib/satellite.ex
+++ b/exercises/practice/satellite/lib/satellite.ex
@@ -7,7 +7,7 @@ defmodule Satellite do
   @doc """
   Build a tree from the elements given in a pre-order and in-order style
   """
-  @spec build_tree(preorder :: [any], inorder :: [any]) :: tree | {:error, String.t()}
+  @spec build_tree(preorder :: [any], inorder :: [any]) :: {:ok, tree} | {:error, String.t()}
 
   def build_tree(preorder, inorder) do
   end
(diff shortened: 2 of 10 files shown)


neenjaw (approved):
Looks good by me!

config.json
```diff
@@ -2380,6 +2380,21 @@
...
       },
+      {
+        "slug": "satellite",
+        "name": "Satellite",
+        "uuid": "a07e01e6-7cea-4db9-a514-7c95788e90a9",
+        "prerequisites": [
+          "pattern-matching",
+          "tuple",
+          "cond",
+          "recursion"
```
angelikatyborska, line 2391: Let's add:
- `atoms` and `lists` because arguments passed to `build_tree` are lists of atoms
- `enum` for a standard uniqueness check (`(list -- Enum.uniq(list)) == []`) that doesn't require knowing about `MapSet`
```suggestion
          "recursion",
          "atoms",
          "lists",
          "enum"
```

config.json
```diff
@@ -2380,6 +2380,21 @@
         ],
         "difficulty": 9
       },
+      {
+        "slug": "satellite",
+        "name": "Satellite",
+        "uuid": "a07e01e6-7cea-4db9-a514-7c95788e90a9",
+        "prerequisites": [
+          "pattern-matching",
+          "tuple",
```
angelikatyborska, line 2389: The correct concept name is:
```suggestion
          "tuples",
```

config.json
```diff
@@ -2380,6 +2380,21 @@
...
+          "cond",
+          "recursion"
+        ],
+        "practices": [
+          "recusrion"
```
angelikatyborska, line 2394: Typo :)
```suggestion
          "recursion"
```

exercises/practice/satellite/lib/satellite.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Satellite do
+  @typedoc """
+  A tree, which can be empty, or made from a left branch, a node and a right branch
+  """
+  @type tree :: {} | {tree, any, tree}
+
+  @doc """
+  Build a tree from the elements given in a pre-order and in-order style
+  """
+  @spec build_tree(preorder :: [any], inorder :: [any]) :: tree | {:error, String.t()}
```
angelikatyborska, line 10: I forgot to pay attention to this in the previous PRs: the more "Elixir way" would be to return:
```
{:ok, tree} | {:error, String.t()}
```
Usually functions that return an error tuple when something goes wrong, return an ok-tuple when everything is fine (a few examples: [File.read](https://hexdocs.pm/elixir/File.html#read/1), [DateTime.new](https://hexdocs.pm/elixir/DateTime.html#new/2), [Ecto.Repo.insert](https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert/2)). It would be nice if our exercises followed that convention too.

It would be great if you did that change in POV too. Better in a separate PR, that will give you extra reputation points 😬 (reputation is a new feature in v3, it's just some points that you can collect but not really use for anything).

angelikatyborska (commented):
❤️ 🧡 💛 💚 💙 💜

config.json
jiegillet: So many typos... -_- 

I wonder about atoms, because there is no need to handle what the list contains, the tests use atoms, but it could be anything else.

exercises/practice/satellite/lib/satellite.ex
jiegillet: OK. Actually at some point I had an implementation that used `{:ok, tree}` and all the checks along the way used `with` instead of using `MapSet` but I found it a bit bulky. It was the better design in the end.

I'll revisit POV too, what wouldn't I do for useless internet points :)

config.json
angelikatyborska: Right, but there's also the necessity of returning either `:ok` or `:error`, so the student solving the exercise must know how to write an atom, and that's taught in the `atoms` concept. In practice, having `atoms` as a prerequisite doesn't change anything because `atoms` are a direct prerequisite of `tuples`, and an indirect prerequisite of almost everything else :)

angelikatyborska:
And another one getting merged 🚀 🎉

PR #757, opened 2021-06-12
Fix protein translation spec

angelikatyborska:
The spec of `of_rna` was not correct (because in case of an error, a string is returned, not a list of strings), the spec of `of_codon` was just not specific enough IMO :)

### DIFF
--- a/exercises/practice/protein-translation/lib/protein_translation.ex
+++ b/exercises/practice/protein-translation/lib/protein_translation.ex
@@ -2,7 +2,7 @@ defmodule ProteinTranslation do
   @doc """
   Given an RNA string, return a list of proteins specified by codons, in order.
   """
-  @spec of_rna(String.t()) :: {atom, list(String.t())} | {:error, String.t()}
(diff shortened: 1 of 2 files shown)


exercises/practice/protein-translation/lib/protein_translation.ex
```diff
@@ -2,7 +2,7 @@ defmodule ProteinTranslation do
   @doc """
   Given an RNA string, return a list of proteins specified by codons, in order.
   """
-  @spec of_rna(String.t()) :: {atom, list(String.t())}
+  @spec of_rna(String.t()) :: {atom, list(String.t())} | {:error, String.t()}
```
jiegillet, line 5:
```suggestion
  @spec of_rna(String.t()) :: {:ok, list(String.t())} | {:error, String.t()}
```

angelikatyborska: Yes! I missed that one, thanks :)

neenjaw (approved):

PR #759, opened 2021-06-12
Add script that checks concepts

angelikatyborska:
We need to fix two concepts:

```
pattern-matching ⨯
  Too many exercises practice this concept.
  Practiced by (11): [alphametics, bowling, meetup, perfect-numbers, poker, pov, protein-translation, queen-attack, tournament, triangle, wordy]


recursion ⨯
  Too many exercises practice this concept.
  Practiced by (11): [accumulate, binary-search, binary-search-tree, matching-brackets, pascals-triangle, pov, prime-factors, roman-numerals, sieve, spiral-matrix, strain]

```

### DIFF
--- a/bin/check_concepts.exs
+++ b/bin/check_concepts.exs
@@ -1,0 +1,128 @@
+Mix.install([
+  {:jason, "~> 1.2"}
+])
+
+{opts, _} = OptionParser.parse!(
+  System.argv(),
+  [strict: [verbose: :boolean, no_warn: :boolean], aliases: [v: :verbose]]
+)
+
+concept_practices_exercise_limit = 10
+
+config =
+  "./config.json"
+  |> File.read!()
+  |> Jason.decode!()
+
+concept_slugs =
+  config["concepts"]
+  |> Enum.map(& &1["slug"])
+  |> Enum.sort()
+
+print_exercises =
+  fn exercises ->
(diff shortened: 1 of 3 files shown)


neenjaw:
Awesome! Love the Mix.install,  Could make those practice changes here to get it passing before merge.

bin/check_concepts.exs
```diff
@@ -0,0 +1,128 @@
+Mix.install([
+  {:jason, "~> 1.2"}
+])
```
neenjaw, line 3: 😍

neenjaw (approved):

angelikatyborska:
Ok, here are my proposed changes:
- POV doesn't practice pattern matching, only recursion
- Accumulate doesn't practice recursion, only lists and anonymous functions

I don't have a great justification for this choice... my only goal was to avoid exercises that don't practice anything (so for example I wouldn't remove recursion from Roman Numerals, because then Roman Numerals don't practice anything).

neenjaw:
I think this is a product problem that will have to be addressed at somepoint because having more exercises in a track means these collisions will happen more

angelikatyborska:
Oh, we need one more change after adding Satellite:
- Prime Factors no longer practice recursion because they're more specific than that, they practice tail-call recursion

PR #760, opened 2021-06-13
[#722] New practice exercise `yacht`

elibdev:
This is my first contribution to this project, so I'm not sure if I've done everything right. I haven't edited the track config because I'm not sure how the difficulty would be rated. I would be happy to make any necessary changes to this PR. Thanks!


exercises/practice/yacht/mix.exs
```diff
@@ -0,0 +1,27 @@
...
+      extra_applications: [:logger]
+    ]
+  end
+
+  # Run "mix help deps" to learn about dependencies.
+  defp deps do
+    [
+      {:jason, "~> 1.2"}
+    ]
+  end
```
neenjaw, line 26: We can't use external dependencies in the test runner presently. See note down below about this

exercises/practice/yacht/.meta/gen_tests_toml.exs
```diff
@@ -0,0 +1,48 @@
+# Generate tests.toml file from canonical-data.json
```
neenjaw, line 1: Interesting approach here, we haven't done anything like this yet to generate test files or test toml configurations so I think this might not work with our current set up for v3. Let me take a think what might need to change to accommodate this method.

exercises/practice/yacht/.meta/example.ex
```diff
@@ -0,0 +1,76 @@
...
+  def score("fives", dice), do: score(5, dice)
+  def score("sixes", dice), do: score(6, dice)
+
+  def score(number, dice) when is_integer(number) do
+    Enum.count(dice, &(&1 == number)) * number
+  end
+
+  def score("full house", dice) do
+    full_house =
+      Enum.frequencies(dice)
```
angelikatyborska, line 21: This function is only available from Elixir 1.10. One of the challenges of the example solution is that it needs to work in older versions too, currently we start testing from 1.7. I'm going to update the documentation to make this more explicit.

We had this problem with another exercise already 🙂 here:
https://github.com/exercism/elixir/blob/0caae7f55f150e0af192d7720eb0c10907572ae5/exercises/practice/word-count/.meta/example.ex#L27-L33

angelikatyborska (commented):
Thank you a lot for the contribution 🎉. I am really sorry that I didn't write exhaustive documentation right away, which caused you to try very reasonable and clever automated approaches that need to be undone. I hope you have the time and motivation to do the changes, we would love to merge this PR with a new exercise 💜 

> I haven't edited the track config because I'm not sure how the difficulty would be rated.

Judging by your example solution, I would say... a 5? It's just a guess, it doesn't have to be perfect 😬 

For "prerequisites", I would suggest:

```
"lists",
"cond",
"case",
"if",
"multiple-clause-functions",
"pattern-matching",
"guards",
"enum"
```

And for "practices", I would suggest "enum", but you would also need to remove "enum" from "practices" of another exercise, for example... Zebra Puzzle? There is a limit that a concept can be practiced by maximum 10 exercises.

exercises/practice/yacht/mix.exs
```diff
@@ -0,0 +1,27 @@
+defmodule Yacht.MixProject do
+  use Mix.Project
+
+  def project do
+    [
+      app: :yacht,
+      version: "0.1.0",
+      # elixir: "~> 1.12",
```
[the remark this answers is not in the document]
elibdev: Got it, I'll edit this and make sure it works with 1.8. Thanks!

exercises/practice/yacht/test/yacht_test.exs
```diff
@@ -0,0 +1,16 @@
...
+
+  for test_case <- @data["cases"] do
+    test test_case["description"] do
+      assert Yacht.score(
+               unquote(test_case["input"]["category"]),
+               unquote(test_case["input"]["dice"])
+             ) == unquote(test_case["expected"])
+    end
+  end
+end
```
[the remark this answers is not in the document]
elibdev: Thanks for explaining the context around this and why this isn't a valid approach for the way the exercises work. I will definitely update this PR to include a test suite with the tests all explicitly written out so it will work as expected for students. I am interested in helping with that test file generation project, but I understand how it's separate and outside the scope of the exercises.

exercises/practice/yacht/mix.exs
elibdev: No problem, I will update the PR without the dependency. I mostly did this as a fun exercise for myself so please don't worry about it not being documented explicitly (the contributing documentation is very helpful). I had a feeling it might not fit in the project as-is but I'm happy to make the changes so it does.

exercises/practice/yacht/.meta/gen_tests_toml.exs
elibdev: Great, I'll delete this. Thanks!

exercises/practice/yacht/.meta/example.ex
elibdev: Got it, I'll update my solution to not use `Enum.frequencies/1` and will test it on Elixir 1.7 to make sure it works with that as well.

elibdev:
Thanks for all the feedback! I'll make the necessary changes that you've helpfully laid out. For the track config "practices" would it be better if I changed my solution to not use the `Enum` module as much and use `for` instead, if there are too many exercises that require `Enum`?

angelikatyborska:
> For the track config "practices" would it be better if I changed my solution to not use the Enum module as much and use for instead, if there are too many exercises that require Enum?

No, don't worry about it! I think it's the nature of Exercism exrcises and Elixir's standard library that `Enum` plays a big role in almost all of them 😉. As far as I know, only a small subset of operations on enumerables is easier to achieve with `for` than with the `Enum` module (e.g. creating cartesian products). For this exercise, `Enum` is better suited, so let's keep it.

elibdev:
A couple things I wasn't sure about:
- In the track config, I just generated a new UUID v4 fo the exercise. Is that exercise UUID supposed to come from somewhere instead?
- I put the "practices" as "multiple-clause-matching" so we don't have to change an existing Enum exercise. I think for this exercise the multiple clause matching does apply but I'm not sure if that's the right one to pick.

exercises/practice/yacht/test/yach_test.exs
```diff
@@ -0,0 +1,143 @@
+defmodule YachtTest do
```
angelikatyborska, line 1: There's a typo in the file name

exercises/practice/yacht/lib/yacht.ex
```diff
@@ -0,0 +1,8 @@
+defmodule Yacht do
+  @doc """
+  Calculate the score of 5 dice using the given category's scoring method.
+  """
+  @spec score(category :: String.t(), dice :: [integer]) :: integer
+  def score(category, dice) do
```
angelikatyborska, line 6: What would you think about making the first argument an atom instead? It serves a bit as a "enum", it doesn't need to be uppercased, split, concatenated or anything, so it doesn't really need to be a string.

```suggestion
  @type category ::
          :ones
          | :twos
          | :threes
          | :fours
          | :fives
          | :sixes
          | :full_house
          | :four_of_a_kind
          | :little_straight
          | :big_straight
          | :choice
          | :yacht

  @doc """
  Calculate the score of 5 dice using the given category's scoring method.
  """
  @spec score(category :: category(), dice :: [integer]) :: integer
```

angelikatyborska (approved):
Excellent ❤️ I left a small (optional) suggestion. The `configlet / lint` is failing due to a typo in a filename, nothing serious.

> Is that exercise UUID supposed to come from somewhere instead?

No, generating a new unique value is correct 👍 

> I think for this exercise the multiple clause matching does apply but I'm not sure if that's the right one to pick.

Sounds like a good pick to me! It's really hard to tell at the beginning. For existing exercises, we analyzed some community solutions to see which approaches people use the most. Here, we only have one solution... but I feel good about suggesting to students that they _should_ solve this one with multiple clause functions 😄

exercises/practice/yacht/lib/yacht.ex
elibdev: I like that much better!

angelikatyborska:
All green = merge 🚀 

Thank you again 💜. The next exercise won't be available on exercism.io until v3 is launched, but you will be able to check it out on v3 staging (https://exercism.lol/tracks/elixir).

If you enjoyed adding a new exercise, I added a few more issues about more practice exercises 😇

elibdev:
Thank you for all your help, and I will check out the new exercise issues! 💜

PR #762, opened 2021-06-14
Set a specific range of supported versions

angelikatyborska:
Let's say we support exactly 6 latest minor Elixir versions. That's more specific than "at least 2 years".

Also fixes a mistake from my last PR where I called them "major" Elixir versions 🙃.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,15 @@
 # Exercism Elixir Track
 
-![build status](https://travis-ci.org/exercism/elixir.svg?branch=master)
+![GitHub branch checks state](https://img.shields.io/github/checks-status/exercism/elixir/main)
+![GitHub contributors](https://img.shields.io/github/contributors-anon/exercism/elixir)
+![GitHub commit activity](https://img.shields.io/github/commit-activity/m/exercism/elixir)
 
-Exercism Exercises in Elixir
+[Exercism Exercises in Elixir](https://exercism.io/my/tracks/elixir)
 
 ## Setup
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,14 +1,15 @@
 # Exercism Elixir Track
 
-![build status](https://travis-ci.org/exercism/elixir.svg?branch=master)
+![GitHub branch checks state](https://img.shields.io/github/checks-status/exercism/elixir/main)
```
angelikatyborska, line 3: I got this from https://shields.io/category/build but I'm not sure if it works... it says "checks: pending" for me. But the old one definitely doesn't work anymore 😬

neenjaw: If you go into the repo actions tab, select a workflow, then click the `...` there is a badge button there, where a modal will pop up with that badge snippet, like this one for configlet

```
[![configlet](https://github.com/exercism/elixir/actions/workflows/configlet.yml/badge.svg)](https://github.com/exercism/elixir/actions/workflows/configlet.yml)
```

neenjaw (approved):
Left a small comment about the badge.

README.md
angelikatyborska: Right, but that forces me to choose a workflow. I'm looking for a badge that would show the "checks" status, which includes adding together the statuses of many workflows. Basically a badge equivalent of this green tick :)

Screen Shot 2021-06-14 at 10 17 43

neenjaw: Sounds good, let's try it then 🤷‍♂️👍

angelikatyborska: Looks like "just merging" didn't fix the problem 😬 I will have to debug it later

PR #769, opened 2021-06-16
[#765] New practice exercise `complex-numbers`

jiegillet:
New one!

Inspired by @elibdev I have written (and committed, I know it doesn't belong in this PR) a quick and dirty script `generate_tests.exs`. This goes through `canonical-data.json` and build the test and lib files. Of course, it is still required to edit by hand, but I it was way easier than copy/pasting stuff from 31 tests. 

I have not tested it very much on other exercises, and even here it has limitations (this `canonical-data.json ` had nested `describe` which is not allowed in Elixir, so I fixed that by hand), but if it's something that is worth pursuing, I'm willing to make it better. It could also generate other files pretty easily. Let me know what you think. I will personally keep using it for more exercises.


config.json
```diff
@@ -2430,6 +2430,21 @@
...
+        "slug": "complex-numbers",
+        "name": "Complex Numbers",
+        "uuid": "0cc86a92-7118-4cb9-8417-db4557baf364",
+        "prerequisites": [
+          "pattern-matching",
+          "erlang-libraries",
+          "floating-point-numbers"
+        ],
+        "practices": [
+          "pattern-matching",
```
jiegillet, line 2443: Should I remove pattern matching? It's not crucial at all. Maybe replace by `erlang-libraries` because of the `:math` functions?

angelikatyborska:
Thank you again! I'll try to review until the end of the week, I'm a bit busy in the next few days.

angelikatyborska:
> I have not tested it very much on other exercises, (...) but if it's something that is worth pursuing, I'm willing to make it better.

It's definitely something that we would want at some point. Ideally, it would allow you to automatically implement new tests for all exercise. Tests often get added to canonical data after the exercise has been around for while. You get new tests by running `bin/configlet sync`, which should add them to the `tests.toml` files. A real test generator would read `tests.toml` files and on based on their contents and `canonical-data.json` files update the test suite. Different exercises are slightly incompatible with `canonical-data.json` in different ways, so it's possible that it would still require some manual work each time.

> I will personally keep using it for more exercises.

Good plan! Let's keep it in the repository so that other people can use it too, and that it serves as a reference for when we will be building a proper test generator.

angelikatyborska (commented):
Looks great 🙂 we just need to update "practice" to get the checks passing and move the generator script somewhere else.

exercises/practice/complex-numbers/.docs/instructions.md
```diff
@@ -0,0 +1,29 @@
...
+
+Dividing a complex number `a + i * b` by another `c + i * d` gives:
+`(a + i * b) / (c + i * d) = (a * c + b * d)/(c^2 + d^2) + (b * c - a * d)/(c^2 + d^2) * i`.
+
+Raising e to a complex exponent can be expressed as `e^(a + i * b) = e^a * e^(i * b)`, the last term of which is given by Euler's formula `e^(i * b) = cos(b) + i * sin(b)`.
+
+Implement the following operations:
+ - addition, subtraction, multiplication and division of two complex numbers,
+ - conjugate, absolute value, exponent of a given complex number.
+
```
neenjaw, line 27: I would like to see a description of how the complex pair is represented with a tuple to represent the real and imaginary part.

Ex) "For this exercise, a complex number will be a tuple-pair where the number `4+3i` would be represented by `{4, 3}`."

neenjaw (approved):
NIce work implementing this exercise, will be a good addition.

Left a comment about the description, I think it'd be useful to explain how we are representing the complex number with a tuple-pair.

I also agree with @angelikatyborska's suggestions.

exercises/practice/complex-numbers/generate_tests.exs
```diff
@@ -0,0 +1,158 @@
+# Generates test cases and basic template, which will need to be adapted by hand
+# Pass the name of the exercise (e. g, "complex-numbers") as an argument
```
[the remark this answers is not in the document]
jiegillet: Yes, it should definitely be in `bin`, I just left it here for now for discussion. 
I will probably expand it a bit to generate all of the other files, might as well,  and rename it `bootstrap_practice_exercise.exs` or something to be more explicit. If it can serve as a basis for a better test generator later, that would be good.

exercises/practice/complex-numbers/generate_tests.exs
```diff
@@ -0,0 +1,158 @@
...
+    """
+  end
+
+  def print_test_case(
+        %{
+          "description" => description,
+          "property" => property,
+          "input" => %{} = input,
+          "expected" => expected
+        } = c,
```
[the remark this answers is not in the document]
jiegillet: Good to know, thank you! As of now, it would print the test in the following way
```elixir
test "disallow empty first strand" do
# Normalises error messages
# -- reimplements --
# db92e77e-7c72-499d-8fe6-9354d2bfd504
...
```
Which is not that bad, since there needs to be a manual check anyway.

exercises/practice/complex-numbers/.docs/instructions.md
jiegillet: Good idea, thank you!

jiegillet: All the documentation I saw discouraged me from changing the description of the exercise, so I updated the `@typedoc` instead.

neenjaw: You can put it in an append file: https://github.com/exercism/docs/blob/main/building/tracks/practice-exercises.md#documentation-files

angelikatyborska (approved):

exercises/practice/complex-numbers/.docs/instructions.md
angelikatyborska: Just a word of explanation: we're not supposed to change anything in `instructions.md` because there will be a script soon that allows synchronizing it with the problem specifications repository in case of changes, and then track-specific changes would get overwritten. That's why there's an append file for this.

PR #773, opened 2021-06-20
[#763] New practice exercise `affine-cipher`

jiegillet:
One more exercise.

All files (minus `config.json` and `tests.toml`) were generated automatically from `bin/boostrap_practice_exercise.exs`. Of course, the tests needed a bunch of manual input, but usually at the level of search and replace, so it's quite valuable. 

I tried to be a little bit clever about some things (for example `@spec` generation for the `lib` file) but I hope that contributors understand that they should not trust these guesses too much. Is it better to make them obviously wrong?  
Please test that script by running `elixir bin/boostrap_practice_exercise.exs darts` (or whichever exercise is not done yet, I have not tested for those yet) and inspect the generated files. I'm sure there is room for improvement. 

Now to think of it, should I make a different PR for this?


config.json
```diff
@@ -2448,6 +2448,25 @@
...
+          "pattern-matching",
+          "atoms",
+          "enum",
+          "integers",
+          "if",
+          "strings",
+          "tuples"
+        ],
+        "practices": [
+          "enum",
```
jiegillet, line 2466: Sorry, I ran the checks on a slightly older branch that didn't have the concept tracking. 
I'll remove `enum`. 
Maybe I could add `pipe-operator` instead? `encode` is just one clean pipe.

angelikatyborska:
> Now to think of it, should I make a different PR for this?

Yes, a separate PR with just the bootstrap script would be great so that we don't need to block merging the exercise if the script need improvements or the other way around.

I'll review only the exercise on this PR and wait for a new PR with the script 🙂

config.json
```diff
@@ -2464,6 +2464,26 @@
           "multiple-clause-matching"
         ],
         "difficulty": 5
+      },
+      {
+        "slug": "affine-cipher",
+        "name": "Affine Cipher",
+        "uuid": "18d12bac-ead0-456c-bcef-ccb2fd06fe72",
+        "prerequisites": [
+          "pattern-matching",
```
angelikatyborska, line 2473: Pattern matching needs some friends to reach its full potential 😁 
```suggestion
          "pattern-matching",
          "guards",
          "multiple-clause-functions",
```

config.json
```diff
@@ -2464,6 +2464,26 @@
...
+          "atoms",
+          "pipe-operator",
+          "enum",
+          "integers",
+          "if",
```
angelikatyborska, line 2478: Let's force people to know all conditional expressions and choose the appropriate one on their own (usually those can be interchangeable anyway)
```suggestion
          "if",
          "cond",
          "case",
```

config.json
```diff
@@ -2464,6 +2464,26 @@
...
+          "strings",
+          "tuples"
```
angelikatyborska, line 2480: The key is a map, so we also need to require maps.

For the code that actually changes the letter to a different one, we need either charlists (your approach) or list comprehensions (alternative approach)
```elixir
      message =
        encrypted
        |> String.replace(@ignored, "")
        |> to_charlist()
        |> Enum.map(fn
          digit when ?0 <= digit and digit <= ?9 -> digit
          char -> Integer.mod(mmi * (char - ?a - b), @alphabet_size) + ?a
        end)
        |> to_string
```

```elixir
      message = String.replace(encrypted, @ignored, "")

      message =
        for << char <- message >>, into: <<>> do
          case char do
            char when ?0 <= char and char <= ?9 -> <<char>>
            char -> <<Integer.mod(mmi * (char - ?a - b), @alphabet_size) + ?a>>
          end
        end
```

```suggestion
          "tuples",
          "maps",
          "list-comprehensions",
          "charlists"
```

config.json
```diff
@@ -2464,6 +2464,26 @@
...
+        ],
+        "practices": [
+          "pipe-operator",
+          "strings"
+        ],
```
angelikatyborska, line 2485: There is a lot going on in this exercise, so I find it really hard to say what "practices" would be best. I think those make sense.

angelikatyborska (commented):
Excellent work as always. You're unstoppable with the new exercises 😁.

I left a request to extend prerequisites and some optional feedback for the stub.

exercises/practice/affine-cipher/lib/affine_cipher.ex
```diff
@@ -0,0 +1,17 @@
+defmodule AffineCipher do
+  @doc """
+  Encode an encrypted message using a key
+  """
+  @spec encode(key :: %{a: integer, b: integer}, message :: String.t()) ::
+          {:ok, String.t()} | {:error, String.t()}
+  def(encode(%{a: a, b: b}, message)) do
```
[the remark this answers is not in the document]
jiegillet: I have no idea how that happened :)

angelikatyborska (approved):
Looks great 👍 I'll wait a bit before merging to see if @neenjaw wants to add a review too.

neenjaw:
Probably don't have time to look at this today, I trust your judgement @angelikatyborska :D

angelikatyborska:
No worries! I understand that sometimes less review work is better, so I'll just merge and @neenjaw can submit a patch PR in case there's something he really wants changed 😀

PR #774, opened 2021-06-21
Bootstrap practice exercises

jiegillet:
This script will generate all files (minus config.json and tests.toml) for a practice exercise. Of course, the tests need a bunch of manual input, but usually at the level of search and replace, so it's quite valuable.

I tried to be a little bit clever about some things (for example @spec generation for the lib file) but I hope that contributors understand that they should not trust these guesses too much. Is it better to make them obviously wrong?
Please test that script by running `elixir bin/boostrap_practice_exercise.exs darts` (or whichever exercise) and inspect the generated files. I'm sure there is room for improvement.

### DIFF
--- a/bin/bootstrap_practice_exercise.exs
+++ b/bin/bootstrap_practice_exercise.exs
@@ -5,6 +5,11 @@
 # $ elixir bin/boostrap_practice_exercise.exs complex-numbers
 # Pass the name of the exercise (e. g., "complex-numbers") as an argument
 
+Mix.install([
+  {:jason, "~> 1.2"},
+  {:toml, "~> 0.6"}
+])
+
 defmodule Generate do
   def explore_properties(%{"cases" => cases}) when is_list(cases) do
     Enum.map(cases, &Generate.explore_properties/1)
@@ -120,11 +125,6 @@ defmodule Generate do
     """
   end
 end
-
-Mix.install([
-  {:jason, "~> 1.2"},
-  {:toml, "~> 0.6"}
-])
 
 [exercise] = System.argv()
 
--- a/bin/bootstrap_practice_exercise.exs
+++ b/bin/bootstrap_practice_exercise.exs
@@ -12,34 +12,53 @@ defmodule Generate do
     |> Enum.map_join("_", &String.downcase/1)
   end
 
-  def get_properties(%{"cases" => cases}) when is_list(cases) do
-    Enum.map(cases, &Generate.get_properties/1)
-    |> Enum.concat()
-    |> Enum.uniq()
+  def explore_properties(%{"cases" => cases}) when is_list(cases) do
+    Enum.map(cases, &Generate.explore_properties/1)
+    |> Enum.reduce(
+      %{},
+      &Map.merge(&1, &2, fn _prop, p1, p2 -> %{p1 | error: p1.error or p2.error} end)
+    )
   end
 
-  def get_properties(%{"property" => property, "input" => %{} = input}),
-    do: [
-      {property,
-       Enum.map(input, fn
-         {var, %{} = val} -> {var, Enum.map(val, fn {v, _} -> v end)}
-         {var, _} -> {var, nil}
-       end)}
-    ]
+  def explore_properties(%{"property" => property, "input" => input, "expected" => expected}),
+    do: %{
+      property => %{
+        name: Generate.to_snake_case(property),
+        variables:
+          if match?(%{}, input) do
+            Enum.map(input, fn
+              {var, %{} = val} -> {var, Enum.map(val, fn {v, _} -> v end)}
+              {var, _} -> {var, nil}
+            end)
+          else
+            [{"input", nil}]
+          end,
+        error: match?(%{"error" => _err}, expected)
+      }
+    }
 
-  def get_properties(%{"property" => property}), do: [{property, "input"}]
-  def get_properties(_data), do: []
+  def explore_properties(_data), do: %{}
 
-  def print_property({property, variables}) do
+  def print_property(%{name: name, variables: variables, error: error}) do
+    return_type = if error, do: "{:ok, TODO.t()} | {:error, String.t()}", else: "TODO.t()"
+
+    variable_types =
+      Enum.map_join(variables, ", ", fn
+        {var, nil} ->
+          "#{var} :: TODO.t()"
+
+        {var, sub_vars} ->
+          "#{var} :: %{#{Enum.map_join(sub_vars, ", ", &(&1 <> " : TODO.t()"))}}"
+      end)
+
+    variable_list = Enum.map_join(variables, ", ", fn {var, _} -> var end)
+
     """
      @doc \"\"\"
-     insert function description here
+     TODO: add function description and replace types in @spec
      \"\"\"
-     @spec #{property}(#{Enum.map_join(variables, ", ", fn
-      {var, nil} -> "#{var} :: String.t()"
-      {var, sub_vars} -> "#{var} :: %{#{Enum.map_join(sub_vars, " : String.t(), ", fn v -> v end)} : String.t()}"
-    end)}) :: {:ok, String.t()} | {:error, String.t()}
-     def #{property}(#{Enum.map_join(variables, ", ", fn {var, _} -> var end)}) do
+     @spec #{name}(#{variable_types}) :: #{return_type}
+     def #{name}(#{variable_list}) do
      end
     """
   end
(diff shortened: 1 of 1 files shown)


bin/bootstrap_practice_exercise.exs
```diff
@@ -0,0 +1,258 @@
...
+     @doc \"\"\"
+     insert function description here
+     \"\"\"
+     @spec #{property}(#{Enum.map_join(variables, ", ", fn
+      {var, nil} -> "#{var} :: String.t()"
+      {var, sub_vars} -> "#{var} :: %{#{Enum.map_join(sub_vars, " : String.t(), ", fn v -> v end)} : String.t()}"
+    end)}) :: {:ok, String.t()} | {:error, String.t()}
+     def #{property}(#{Enum.map_join(variables, ", ", fn {var, _} -> var end)}) do
+     end
+    """
```
angelikatyborska, line 44: > Is it better to make them obviously wrong?

Maybe it's better to somehow mark them as requiring manual changes? Maybe something like this? (and similar in other places that also have most-likely-wrong defaults)
```suggestion
    """
     @doc \"\"\"
     TODO: insert function description here
     \"\"\"
     @spec #{property}(#{Enum.map_join(variables, ", ", fn
      {var, nil} -> "#{var} :: TODO.t()"
      {var, sub_vars} -> "#{var} :: %{#{Enum.map_join(sub_vars, " : TODO.t(), ", fn v -> v end)} : TODO.t()}"
    end)}) :: TODO.t()
     def #{property}(#{Enum.map_join(variables, ", ", fn {var, _} -> var end)}) do
     end
    """
```

bin/bootstrap_practice_exercise.exs
```diff
@@ -0,0 +1,258 @@
...
+          "expected" => expected
+        } = c,
+        module
+      ) do
+    """
+    @tag :pending
+    test \"#{description}\" do
+      #{Generate.print_comments(c, ["description", "property", "input", "expected", "uuid"])}
+      #{print_input(input)}
+      output = #{module}.#{Generate.to_snake_case(property)}(#{print_variables(input)})
```
angelikatyborska, line 99: It would be good if changing to snake case happened as soon as possible, right after fetching and decoding the JSON, and not when printing the test case. For example, you forgot to use it when printing the stub 😉 so for an exercise like `allergies` (https://github.com/exercism/problem-specifications/blob/e2779fc84ff07cb5bd8ff89f38e78af6d17b5056/exercises/allergies/canonical-data.json#L14), you get a stub like this:
```elixir
def allergicTo(item, score) do
end
```

but a function call in tests like this:
```elixir
output = Allergies.allergic_to(item, score)
```

It's also necessary to change properties to snake case, for example: https://github.com/exercism/problem-specifications/blob/daf620d47ed905409564dec5fa9610664e294bde/exercises/all-your-base/canonical-data.json#L20-L22

bin/bootstrap_practice_exercise.exs
```diff
@@ -0,0 +1,258 @@
...
+
+  def print_input(input), do: "input = #{inspect(input)}"
+
+  def print_variables(%{} = input),
+    do: Enum.map_join(input, ", ", fn {variable, _value} -> variable end)
+
+  def print_variables(input), do: input
+
+  def print_expected(%{"error" => err}), do: "{:error, #{inspect(err)}}"
+  def print_expected(expected), do: "{:ok, #{inspect(expected)}}"
```
angelikatyborska, line 74: Would it be possible (and not too difficult) to make this more intelligent, and only use `{:ok, value}` if any error case exists in the problem specification for this property, and otherwise just use the plain value? There are a lot of exercises without any error handling.

bin/bootstrap_practice_exercise.exs
```diff
@@ -0,0 +1,258 @@
...
+
+# .formatter.exs
+format = """
+# Used by "mix format"
+[
+  inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}"]
+]
+"""
+
+Mix.Generator.create_file("exercises/practice/#{exercise}/.formatter.exs", format)
```
angelikatyborska, line 140: I had no idea about this module 😍 awesome

angelikatyborska (approved):
I run it on a few existing exercises and left some (hopefully realistic) improvement ideas based on the output.

This is excellent 🎉. I guess with this script you'll churn through all the remaining "new practice exercise" issues in no time. I have to create more 😅.

neenjaw:
I'd like to look over this one before merge, but might need to wait a couple days if that's okay.  I have all next week off, so should be able to catch up on this stuff. :)

bin/bootstrap_practice_exercise.exs
jiegillet: It is quite nice. And actually it wouldn't be much more work to make this script into a `Mix.Task`. It would be kind of cool: `mix ex.practice.gen darts`.

bin/bootstrap_practice_exercise.exs
jiegillet: Done :)

bin/bootstrap_practice_exercise.exs
```diff
@@ -0,0 +1,282 @@
...
+
+    #{Generate.print_comments(data, ["cases", "exercise"])}
+    #{Enum.map_join(data["cases"], "\n\n", &Generate.print_test_case(&1, properties, module))}
+  end
+  """
+  |> String.replace("@tag", "# @tag", global: false)
+
+path = "exercises/practice/#{exercise}/test/#{exercise_path}_test.exs"
+Mix.Generator.create_file(path, test_file)
+System.cmd("mix", ["format", path])
```
jiegillet, line 282: Do you know if there is a way to call `mix format` from within a `.exs` script without having to use `System.cmd`? I searched for it, but couldn't find it.  
Although maybe it's best to run it this way so that it includes the  format in `.formatter.exs`.

angelikatyborska: Every mix task is just a module with a `run` function, so it should be possible to run `Mix.Tasks.Format.run([])`. It should read `.formatter.exs` from the current directory anyway.

mix format documentation: https://hexdocs.pm/mix/master/Mix.Tasks.Format.html
and implementation: https://github.com/elixir-lang/elixir/blob/2a4312412be39303af08ccd0cef81b5bbcd55eac/lib/mix/lib/mix/tasks/format.ex#L1

jiegillet:
> I guess with this script you'll churn through all the remaining "new practice exercise" issues in no time. I have to create more 😅.

I compiled a list of all the exercises we don't have yet (separating the ones with and without tests), do you want me to share it in an issue? Or do you prefer individual ones?

bin/bootstrap_practice_exercise.exs
```diff
@@ -0,0 +1,283 @@
...
+# Run the following command from the root of the repo:
+# $ elixir bin/boostrap_practice_exercise.exs complex-numbers
+# Pass the name of the exercise (e. g., "complex-numbers") as an argument
+
+defmodule Generate do
+  def to_snake_case(lower_camel_case) do
+    lower_camel_case
+    |> String.split(~r/(?=[A-Z\d])/)
+    |> Enum.map_join("_", &String.downcase/1)
+  end
```
neenjaw, line 13: We can use `Macro.underscore` for this if we are using it just for identifiers

bin/bootstrap_practice_exercise.exs
```diff
@@ -0,0 +1,283 @@
...
+      assert output == expected
+    end
+    """
+  end
+end
+
+Mix.install([
+  {:jason, "~> 1.2"},
+  {:toml, "~> 0.6"}
+])
```
neenjaw, line 133: I don't know if there is really a standard established for mix install, but can we hoist this to the top of the file?

bin/bootstrap_practice_exercise.exs
```diff
@@ -0,0 +1,283 @@
...
+
+  # Run "mix help deps" to learn about dependencies.
+  defp deps do
+    [
+      # {:dep_from_hexpm, "~> 0.3.0"},
+      # {:dep_from_git, git: "https://github.com/elixir-lang/my_dep.git", tag: "0.1.0"}
+    ]
+  end
+end
+"""
```
neenjaw, line 194: Minor: if we can't run mix format on this file, can we indent here 2 spaces for this binding?

bin/bootstrap_practice_exercise.exs
```diff
@@ -0,0 +1,283 @@
...
+
+Mix.Generator.create_file("exercises/practice/#{exercise}/mix.exs", mix)
+
+# test/test_helper.exs
+test_helper = """
+ExUnit.start()
+ExUnit.configure(exclude: :pending, trace: true)
+"""
```
neenjaw, line 202: Minor, same indent as mentioned above

neenjaw (approved):
Left a few minor comments, cool stuff!

angelikatyborska:
> I compiled a list of all the exercises we don't have yet (separating the ones with and without tests), do you want me to share it in an issue? Or do you prefer individual ones?

There is a dashboard for this already: https://tracks.exercism.io/elixir/main/unimplemented

Sadly it doesn't mention which exercises don't have tests defined, but there's only a few left unimplemented so we can manage.

Normally only maintainers deal with exercise selection, but since you're working faster than I can review or write new issues, I want to share with you my thoughts on implementing those missing exercises:

1. I don't want any of the exercises that don't have `canonical-data.json` (error handling, hangman, ledger, lens-person, paasio, robot-name, tree building) because that puts on me and Tim the responsibility to figure out which tests cases are necessary for the story, to think of the corner cases etc., it just sounds like too much work 😅 
2. I don't want any of the exercises that are ridiculously easy in a high level programming language (reverse string, maybe micro blog). That might not apply to square root because I think square root could forbid using the standard library and demand a manual implementation.
3. I don't want high scores because it will conflict with a similar concept exercise that we already have

The rest can be implemented and your PRs with them will be very welcome if you still have the time and energy 🙂

bin/bootstrap_practice_exercise.exs
jiegillet: At the end we run mix format on everything, it should take care of this one too.

bin/bootstrap_practice_exercise.exs
```diff
@@ -0,0 +1,277 @@
...
+
+# Run "mix help deps" to learn about dependencies.
+defp deps do
+  [
+    # {:dep_from_hexpm, "~> 0.3.0"},
+    # {:dep_from_git, git: "https://github.com/elixir-lang/my_dep.git", tag: "0.1.0"}
+  ]
+  end
+end
+"""
```
angelikatyborska, line 188: I think this is what @neenjaw means:
```suggestion
mix = """
defmodule #{module}.MixProject do
  use Mix.Project

  def project do
    [
      app: :#{exercise_snake_case},
      version: "0.1.0",
      # elixir: "~> 1.8",
      start_permanent: Mix.env() == :prod,
      deps: deps()
    ]
  end

  # Run "mix help compile.app" to learn about applications.
  def application do
    [
      extra_applications: [:logger]
    ]
  end

  # Run "mix help deps" to learn about dependencies.
  defp deps do
    [
      # {:dep_from_hexpm, "~> 0.3.0"},
      # {:dep_from_git, git: "https://github.com/elixir-lang/my_dep.git", tag: "0.1.0"}
    ]
  end
end
"""
```

bin/bootstrap_practice_exercise.exs
angelikatyborska: Here the indentation looks good already 🤔 meaning: level 0, there's no indentation.

bin/bootstrap_practice_exercise.exs
jiegillet: How obvious. Thank you :)

angelikatyborska:
I'm going to make a call that all the feedback has been considered and applied, and now it's merge time 🎉

PR #775, opened 2021-06-21
[#766] New practice exercise `circular-buffer`

jiegillet:
One more.

I know I'm overdoing it a little, I have time off this week. I get kind of addicted to these kinds of problems, and if me solving them can help out... 😅 

No rush for the review, please take your time.

### DIFF
--- a/config.json
+++ b/config.json
@@ -2464,6 +2464,32 @@
           "multiple-clause-matching"
         ],
         "difficulty": 5
+      },
+      {
+        "slug": "circular-buffer",
+        "name": "Circular Buffer",
+        "uuid": "535d64c9-95f7-4f59-b7b6-d459337b82b0",
+        "prerequisites": [
+          "atoms",
+          "tuples",
+          "lists",
+          "structs",
+          "agent",
+          "processes",
+          "pids",
+          "case",
+          "cond",
+          "if",
+          "multiple-clause-functions",
+          "pattern-matching",
+          "guards",
+          "errors"
+        ],
+        "practices": [
+          "processes",
+          "errors"
+        ],
+        "difficulty": 8
       }
     ],
     "foregone": [
--- a/config.json
+++ b/config.json
@@ -2474,6 +2474,7 @@
           "tuples",
           "lists",
           "structs",
+          "enum"
           "agent",
           "processes",
           "pids",
--- a/config.json
+++ b/config.json
@@ -2474,7 +2474,7 @@
           "tuples",
           "lists",
           "structs",
-          "enum"
+          "enum",
           "agent",
           "processes",
           "pids",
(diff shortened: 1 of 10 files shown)


neenjaw:
Nice work! This week is a bit busy for me, will try to review by this weekend at the latest

jiegillet:
Thank you and please take your time. I have a week off so I'm enjoying my free time.

exercises/practice/circular-buffer/test/circular_buffer_test.exs
```diff
@@ -0,0 +1,237 @@
...
+  test "can read an item just written" do
+    capacity = 1
+    {:ok, buffer} = CircularBuffer.new(capacity)
+
+    operations = [
+      %{"item" => 1, "operation" => "write", "should_succeed" => true},
+      %{"expected" => 1, "operation" => "read", "should_succeed" => true}
+    ]
+
+    Enum.each(operations, &test_operation(buffer, &1))
```
angelikatyborska, line 48: I think it might be better for students if the whole test suite was a little bit less DRY and a bit more explicit. Something like this, so you don't need to go back to the top of the file to see the expected result:
```suggestion
    assert CircularBuffer.write(buffer, 1) == :ok
    assert CircularBuffer.read(buffer) == {:ok, 1}
```

Especially for beginners, seeing this list of maps with operations might be confusing.

Otherwise, in a non-teaching context, this would be all fine by me :)

config.json
```diff
@@ -2464,6 +2464,32 @@
...
+          "case",
+          "cond",
+          "if",
+          "multiple-clause-functions",
+          "pattern-matching",
+          "guards",
+          "errors"
+        ],
+        "practices": [
+          "processes",
```
angelikatyborska, line 2489: Woohoo, we needed more exercises that practice processes 💜

config.json
```diff
@@ -2464,6 +2464,32 @@
...
+          "errors"
```
angelikatyborska, line 2490: @neenjaw Do you remember how we defined practicing/requiring "errors"? Is it about the ok/error-tuple pattern, or is it about using `raise` (`exceptions` is definitely about defining custom ones)?

config.json
```diff
@@ -2464,6 +2464,32 @@
...
+      },
+      {
+        "slug": "circular-buffer",
+        "name": "Circular Buffer",
+        "uuid": "535d64c9-95f7-4f59-b7b6-d459337b82b0",
+        "prerequisites": [
+          "atoms",
+          "tuples",
+          "lists",
+          "structs",
```
angelikatyborska, line 2476: I believe enum might also be useful (it is used in your solution)
```suggestion
          "structs",
          "enum"
```

angelikatyborska (approved):
Looks great to me, but let's wait a few days to see if @neenjaw wants to add anything

config.json
neenjaw: I think it is implemented to exercise `raise`

neenjaw: I'm not opposed to this addition though, it is an accepted way of handling errors

neenjaw (approved):

PR #776, opened 2021-06-22
[#768] New practice exercise `go-counting`

jiegillet:
Another one. Again, please take your time for the review, I'm just enjoying a week off.

It took me a while to solve, so I put a high difficulty, but maybe I overdid it?

### DIFF
--- a/exercises/practice/go-counting/lib/go_counting.ex
+++ b/exercises/practice/go-counting/lib/go_counting.ex
@@ -1,0 +1,21 @@
+defmodule GoCounting do
+  @type position :: {integer, integer}
+  @type owner :: %{owner: atom, territory: [position]}
+  @type territories :: %{white: [position], black: [position], none: [position]}
+
+  @doc """
+  Return the owner and territory around a position
+
+  """
+  @spec territory(board :: [String.t()], position :: position) ::
+          {:ok, owner} | {:error, String.t()}
+  def territory(board, {x, y} = pos) do
+  end
+
+  @doc """
+  Return all white, black and neutral territories
+  """
+  @spec territories(board :: [String.t()]) :: territories
+  def territories(board) do
+  end
+end
--- a/exercises/practice/go-counting/lib/go_counting.ex
+++ b/exercises/practice/go-counting/lib/go_counting.ex
@@ -5,7 +5,6 @@ defmodule GoCounting do
 
   @doc """
   Return the owner and territory around a position
-
   """
   @spec territory(board :: String.t(), position :: position) ::
           {:ok, owner} | {:error, String.t()}
--- a/exercises/practice/go-counting/lib/go_counting.ex
+++ b/exercises/practice/go-counting/lib/go_counting.ex
@@ -7,7 +7,7 @@ defmodule GoCounting do
   Return the owner and territory around a position
 
   """
-  @spec territory(board :: [String.t()], position :: position) ::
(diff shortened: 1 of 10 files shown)


exercises/practice/go-counting/test/go_counting_test.exs
```diff
@@ -0,0 +1,138 @@
...
+    assert Enum.sort(b1) == Enum.sort(b2)
+    assert Enum.sort(n1) == Enum.sort(n2)
+    assert Enum.sort(w1) == Enum.sort(w2)
+  end
+
+  def equal?(_a, _b), do: flunk("Unexpected output")
+
+  # @tag :pending
+  test "Black corner territory on 5x5 board" do
+    board = ["  B  ", " B B ", "B W B", " W W ", "  W  "]
```
angelikatyborska, line 19: Let's do multiline strings as input. It will be easier to read and not that much harder to implement a solution. In problem specifications, sometimes there is a comment that a list of strings is supposed to be a multiline string (for example: https://github.com/exercism/problem-specifications/blob/e2779fc84ff07cb5bd8ff89f38e78af6d17b5056/exercises/proverb/canonical-data.json#L4-L6). It's weird that it's missing for this exercise.

We will need this hack to keep trailing whitespace in a multiline string: https://github.com/exercism/elixir/blob/9565a2f8acfab7a4d716de750523d34084a765b4/exercises/practice/diamond/test/diamond_test.exs#L53-L106


Or maybe it won't be that much easier to read, in that case we should at least split the list items each in a new line, so that it kinda reads like a multiline string while still being a list 🙂

exercises/practice/go-counting/lib/go_counting.ex
```diff
@@ -0,0 +1,21 @@
+defmodule GoCounting do
+  @type position :: {integer, integer}
+  @type owner :: %{owner: atom, territory: [position]}
+  @type territories :: %{white: [position], black: [position], none: [position]}
```
angelikatyborska, line 4: 👍

exercises/practice/go-counting/test/go_counting_test.exs
```diff
@@ -0,0 +1,138 @@
...
+
+  def equal?({:error, err}, {:error, err}), do: assert(true)
+
+  def equal?({:ok, %{owner: owner, territory: t1}}, {:ok, %{owner: owner, territory: t2}}),
+    do: assert(Enum.sort(t1) == Enum.sort(t2))
+
+  def equal?(%{black: b1, none: n1, white: w1}, %{black: b2, none: n2, white: w2}) do
+    assert Enum.sort(b1) == Enum.sort(b2)
+    assert Enum.sort(n1) == Enum.sort(n2)
+    assert Enum.sort(w1) == Enum.sort(w2)
```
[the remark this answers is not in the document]
jiegillet: OK, let's assume coordinates are sorted, it makes everything cleaner.

exercises/practice/go-counting/test/go_counting_test.exs
jiegillet: Very good suggestion, thank you.
It was a bit awkward to deal with the leading/trailing white spaces, so instead I used "_" for empty spaces.

angelikatyborska: it looks great with "_", good idea!

angelikatyborska (approved):

neenjaw (approved):

PR #777, opened 2021-06-24
[#767] New practice exercise `darts`

jiegillet:
Another one. Not the hardest.

### DIFF
--- a/config.json
+++ b/config.json
@@ -2471,8 +2471,8 @@
         "uuid": "a68113f8-92be-40e0-a97f-eb4239303ef5",
         "prerequisites": [
           "cond",
-          "case",
-          "if"
+          "integers",
+          "floating-point-numbers",
         ],
         "practices": [
           "cond"
--- a/config.json
+++ b/config.json
@@ -2472,8 +2472,7 @@
         "prerequisites": [
           "cond",
           "case",
-          "if",
-          "erlang-libraries"
+          "if"
         ],
         "practices": [
           "cond"
--- a/config.json
+++ b/config.json
@@ -2472,7 +2472,7 @@
         "prerequisites": [
           "cond",
           "integers",
-          "floating-point-numbers",
+          "floating-point-numbers"
         ],
         "practices": [
           "cond"
--- a/config.json
+++ b/config.json
@@ -2466,6 +2466,21 @@
         "difficulty": 5
       },
       {
+        "slug": "darts",
+        "name": "Darts",
+        "uuid": "a68113f8-92be-40e0-a97f-eb4239303ef5",
+        "prerequisites": [
+          "cond",
+          "case",
+          "if",
+          "erlang-libraries"
+        ],
+        "practices": [
+          "cond"
+        ],
+        "difficulty": 2
+      },
+      {
         "slug": "affine-cipher",
         "name": "Affine Cipher",
         "uuid": "18d12bac-ead0-456c-bcef-ccb2fd06fe72",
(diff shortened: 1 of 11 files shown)


config.json
```diff
@@ -2465,6 +2465,21 @@
...
+          "cond",
+          "case",
+          "if",
+          "erlang-libraries"
+        ],
+        "practices": [
+          "cond"
+        ],
+        "difficulty": 2
+      },
```
angelikatyborska, line 2482: @neenjaw Did we agree recently that `Float.pow` and `Integer.pow` (added in Elixir 1.12) count as part of the integers and floats?

If that's the case, then we could say that this is a cond + integers + floats exercise (although we might need leaving a hint that sqrt(x) == pow(x, 0.5) for less math-savvy students).

We have a severe lack of very easy exercises that could be done after solving only very few learning exercises. If we made that decision, then this exercise will actually become the very first exercise that people learning Elixir can solve. All it takes it that they solve `lasagna` (basics), `freelancer-rates` (integers), `pacman-rules` (booleans) and `log-level` (cond)


For reference for @jiegillet (not sure if you have been on the staging website exercism.lol (sic)), here's our current concept tree (on the website you can browse it better than on a screenshot):

elixir concepts

exercises/practice/darts/lib/darts.ex
```diff
@@ -0,0 +1,10 @@
+defmodule Darts do
+  @type position :: {float, float}
```
angelikatyborska, line 2: There are test cases that pass integers in the position. `{number, number}`?

angelikatyborska (approved):
Perfect! I'm actually glad this is so easy. Maybe we can modify prerequisites to make this exercise unlocks much earlier (erlang libraries are pretty far into the concept tree). Let's see what @neenjaw thinks.

config.json
jiegillet: OK, let's leave out `erlang-libraries`. 
Technically there is also a solution that doesn't rely on square roots at all. Something like 
```elixir
radius_sq = x * x + y * y
cond do
  radius_sq > 10 * 10
  ...
```
Although I don't think less math-savvy students would think of that ^^

config.json
```diff
@@ -2465,6 +2465,20 @@
...
+        "name": "Darts",
+        "uuid": "a68113f8-92be-40e0-a97f-eb4239303ef5",
+        "prerequisites": [
+          "cond",
+          "case",
+          "if"
+        ],
+        "practices": [
+          "cond"
+        ],
```
angelikatyborska, line 2479: Ok, if we're leaving out erlang-libraries, I would want this to be a very early exercise which we can achieve by deciding that you're supposed to use `cond`, not `case` or `if or `multiple-clause-functions`
```suggestion
        "prerequisites": [
          "cond",
          "integers",
          "floating-point-numbers",
        ],
        "practices": [
          "cond"
        ],
```


I think it's a reasonable suggestion. It looks more crowded with `case`/`multiple-clause-functions` and `guards`:
```elixir
    cond do
      radius > 10 -> 0
      radius > 5 -> 1
      radius > 1 -> 5
      true -> 10
    end
```
```elixir
    case radius do
      radius when radius > 10 -> 0
      radius when radius > 5 -> 1
      radius when radius > 1 -> 5
      radius -> 10
    end
```
```elixir
    defp do_score(radius) when radius > 10, do: 0
    defp do_score(radius) when radius > 5, do: 1
    defp do_score(radius) when radius > 1, do: 5
    defp do_score(radius), do: 10
```

angelikatyborska:
Could you add a hints document that mentions that `sqrt(x) == pow(x, 0.5)` and maybe also links to instructions how to calculate the distance between two points? I know it's basic math, but since this is supposed to be a very exercise, I don't want anyone to get stuck because of that.

config.json
jiegillet: Sounds good to me

neenjaw (approved):

PR #778, opened 2021-06-24
[#764] New practice exercise `book-store`

jiegillet:
Last one of the open issues for new exercises.

In my example solution I use `Agent` to memoize the possible discounts, but it is not really necessary (the tests went from taking  0.1s to 0.06s after I added it), so I didn't include it in the requirements. 

Speaking of, `enum` and `recursion` are full, so I put `list-comprehensions` but it's a stretch. Something we could do is encourage people to use to memoization with processes in a hint/extra instructions so they can practice that.

### DIFF
--- a/config.json
+++ b/config.json
@@ -2500,6 +2500,7 @@
           "lists",
           "pattern-matching",
           "guards",
+          "multiple-clause-functions",
           "enum",
           "list-comprehensions",
           "maps",
--- a/config.json
+++ b/config.json
@@ -2491,6 +2491,26 @@
           "strings"
         ],
         "difficulty": 5
+      },
+      {
+        "slug": "book-store",
+        "name": "Book Store",
+        "uuid": "d6d9687c-8bf0-482e-8309-4f3d706ba753",
+        "prerequisites": [
+          "lists",
+          "pattern-matching",
+          "guards",
+          "enum",
+          "list-comprehensions",
+          "maps",
+          "pipe-operator",
+          "ranges",
+          "recursion"
+        ],
+        "practices": [
+          "list-comprehensions"
+        ],
+        "difficulty": 5
       }
     ],
     "foregone": [
--- a/config.json
+++ b/config.json
@@ -2510,7 +2510,7 @@
         "practices": [
           "list-comprehensions"
         ],
-        "difficulty": 5
+        "difficulty": 8
       }
     ],
     "foregone": [
--- a/exercises/practice/book-store/lib/book_store.ex
+++ b/exercises/practice/book-store/lib/book_store.ex
@@ -1,5 +1,6 @@
 defmodule BookStore do
-  @type book :: integer
+  @typedoc "A book is represented by its number in the 5-book series"
+  @type book :: 1 | 2 | 3 | 4 | 5
 
   @doc """
   Calculate lowest price (in cents) for a shopping basket containing books.
--- a/exercises/practice/book-store/lib/book_store.ex
+++ b/exercises/practice/book-store/lib/book_store.ex
@@ -1,0 +1,10 @@
+defmodule BookStore do
+  @type book :: integer
+
+  @doc """
+  Calculate lowest price (in cents) for a shopping basket containing books.
+  """
+  @spec total(basket :: [book]) :: integer
+  def total(basket) do
+  end
+end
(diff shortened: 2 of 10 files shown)


exercises/practice/book-store/test/book_store_test.exs
```diff
@@ -0,0 +1,123 @@
...
+  test "Three each of first 2 books and 2 each of remaining books" do
+    # Suggested grouping, [[1,2,3,4,5],[1,2,3,4,5],[1,2]].
+    basket = [1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 1, 2]
+
+    assert BookStore.total(basket) == 7520
+  end
+
+  @tag :pending
+  test "Four groups of four are cheaper than two groups each of five and three" do
+    # Suggested grouping, [[1,2,3,4],[1,2,3,5],[1,2,3,4],[1,2,3,5]].
```
angelikatyborska, line 118: 👍 leaving the comments in seems helpful

exercises/practice/book-store/lib/book_store.ex
```diff
@@ -0,0 +1,10 @@
+defmodule BookStore do
+  @type book :: integer
```
angelikatyborska, line 2: It took me a while to realize why a book is an integer. Maybe something more explicit with a typedoc could help?
```suggestion
  @typedoc "A book is represented by its number in the 5-book series"
  @type book :: 1 | 2 | 3 | 4 | 5
```

config.json
```diff
@@ -2491,6 +2491,26 @@
...
+          "guards",
+          "enum",
+          "list-comprehensions",
+          "maps",
+          "pipe-operator",
+          "ranges",
+          "recursion"
+        ],
+        "practices": [
+          "list-comprehensions"
```
angelikatyborska, line 2511: If nothing fits or everything that fits already has 10 exercises, we could also leave this empty.

angelikatyborska (approved):
> the tests went from taking 0.1s to 0.06s after I added it
I see a similar difference is speed on my machine. Still, 0.1s for the test suite is very fast anyway :)

> Something we could do is encourage people to use to memoization with processes in a hint/extra instructions so they can practice that.

I think this would only make sense if students struggle with making the solution as fast as yours is without Agent. I find it hard to predict right now if that will be the case. We will need to wait and see, and then ask the mentors what students do with this exercise.

If you're interested in making code run fast, we have an example solution to `pythagorean-triplet` that takes 10s on my machine to pass the tests. I wonder if it can be improved further? I think it was already improved from 40s.

jiegillet:
> I think this would only make sense if students struggle with making the solution as fast as yours is without Agent. I find it hard to predict right now if that will be the case. We will need to wait and see, and then ask the mentors what students do with this exercise.

Sounds good :)

> If you're interested in making code run fast, we have an example solution to `pythagorean-triplet` that takes 10s on my machine to pass the tests. I wonder if it can be improved further? I think it was already improved from 40s.

My solution ([a slight adaptation of this one](https://exercism.io/tracks/elixir/exercises/pythagorean-triplet/solutions/d328c30ee8d44d25a233a577e14c0dc7)) runs in 0.06 seconds (on the updated tests of this repo, not the ones in the link). Not bad :)
The trick is not really computational though, it's mathematical: [generating primitive pythagorean triples](https://en.wikipedia.org/wiki/Tree_of_primitive_Pythagorean_triples) makes it really easy. I also found a typo in the test, I'll do a quick PR.

config.json
```diff
@@ -2491,6 +2491,26 @@
...
         "difficulty": 5
+      },
+      {
+        "slug": "book-store",
+        "name": "Book Store",
+        "uuid": "d6d9687c-8bf0-482e-8309-4f3d706ba753",
+        "prerequisites": [
+          "lists",
+          "pattern-matching",
+          "guards",
```
angelikatyborska, line 2502:
```suggestion
          "pattern-matching",
          "guards",
          "multiple-clause-functions",
```

config.json
angelikatyborska: Or we can also leave lists comprehensions and see if people complain :)

angelikatyborska:
IMO good to merge. I'll check up with Tim if we wants to review too or now.

neenjaw (approved):

PR #779, opened 2021-06-25
`pythagorean-triplet` -- Fix test and propose new example solution

jiegillet:
I found a typo in the test, that's the main thing.

I'm also sharing my solution that passes the tests in 0.06s versus 10s for the existing one, but I know that the purpose of an example solution is to prove it can be done and not necessarily be efficient, so I don't feel too strongly about making that change.

### DIFF
--- a/exercises/practice/pythagorean-triplet/.docs/hints.md
+++ b/exercises/practice/pythagorean-triplet/.docs/hints.md
@@ -1,0 +1,5 @@
+## General
+
+- If checking through all sorts of triangles is too slow, consider that [right triangles grow on trees](triples). 
+
+[triples]: https://en.wikipedia.org/wiki/Tree_of_primitive_Pythagorean_triples
--- a/exercises/practice/pythagorean-triplet/.docs/hints.md
+++ b/exercises/practice/pythagorean-triplet/.docs/hints.md
@@ -1,5 +1,5 @@
 ## General
 
-- If checking through all sorts of triangles is too slow, consider that [right triangles grow on trees](triples). 
+- If checking through all sorts of triangles is too slow, consider that [right triangles grow on trees](triples).
 
 [triples]: https://en.wikipedia.org/wiki/Tree_of_primitive_Pythagorean_triples
--- a/config.json
+++ b/config.json
@@ -581,8 +581,7 @@
           "enum"
         ],
         "practices": [
-          "charlists",
-          "enum"
+          "charlists"
         ],
         "difficulty": 2
       },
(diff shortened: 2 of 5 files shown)


angelikatyborska:
> I'm also sharing my solution that passes the tests in 0.06s versus 10s for the existing one

WHAT 😮 what are you, a computer witch/wizard? This is amazing!

angelikatyborska:
Your solution is so superior to the old one that it doesn't even need processes to parallelize calculations :D

Now I need to update "practices" because they recommend "tasks" and "list-comprehensions", but that is obviously leading students in the wrong direction.

I think we will need to write mentoring notes for this exercise to explain https://en.wikipedia.org/wiki/Tree_of_primitive_Pythagorean_triples and how that's necessary for a fast solution

jiegillet:
Hahaha, sorry for the extra work :)

How do mentoring notes work? Do you mean hints?

angelikatyborska:
> Do you mean hints?

No, I do not mean hints, but now that you mention it, that's an even better idea!

Hints would be a document in the exercise's directory (documentation: https://github.com/exercism/docs/blob/main/building/tracks/practice-exercises.md#file-docshintsmd) and each student would see it while solving the exercise. Let's add a hint for this 🙂 

But to also answer the question, mentoring notes are stored in https://github.com/exercism/website-copy/tree/main/tracks/elixir/exercises and they are documents aimed at mentors. They usually list a good solution to an exercise (as opposed to the example solutions in this repository, which are just supposed to work, they don't need to be good). They also list some hints about what to pay attention to when mentoring a student in this exercise, popular mistakes that people make etc.

For mentoring notes to be useful in this situation, the student would need to request mentoring, and then the mentor would need to realize that the solution is too slow and look up the mentoring notes. That's a lot of extra steps... hints sound like a better idea.

jiegillet:
I added a hint, I hope it's not too humorous.
I'll let you add the mentoring notes if that's OK, because I'm not a mentor (although I'm considering it).

exercises/practice/pythagorean-triplet/.docs/hints.md
```diff
@@ -0,0 +1,5 @@
+## General
+
+- If checking through all sorts of triangles is too slow, consider that [right triangles grow on trees](triples).
```
angelikatyborska, line 3: I love it 😂

jiegillet, line 3:
```suggestion
- If checking through all sorts of triangles is too slow, consider that [right triangles grow on trees](triples).
```

angelikatyborska:
I pushed a small change to prerequisites and practices so that they make sense with the clearly superior new example.

I think you would be a great mentor! You should try it sometime. Trying to figure out the correct advice to give to students is quite similar to doing code reviews.

angelikatyborska:
Oh, and I think the hint is enough. If we have the hint, hopefully we don't need mentor notes.

angelikatyborska (approved):

PR #781, opened 2021-06-26
New practice exercise `linked-list`

jiegillet:
New one.

This may need a bit of discussion because implementing true doubly linked lists in Elixir is not possible due to its immutability, so I added an appendix to the instruction asking to simulate the behavior instead. 
The data structure that I used is the same as the one I used in `circular-buffer` (two lists at the start and end), minus the state.
Possibly students might do strange things, I've seen several people solve `simple-linked-list` by implementing lists with tuples, which misses the point completely.


exercises/practice/linked-list/.docs/instructions.append.md
```diff
@@ -0,0 +1,3 @@
+# Introduction append
+
+In immutable, strict languages such as Elixir, there is no way to create cyclical data structures, like doubly linked lists, since they require mutability. You must therefore simulate that behavior in a different way.
```
neenjaw, line 3: This could use a citation if we wanted, this would be a potential: https://elixirforum.com/t/cyclical-data-structures-in-elixir/2987/6

exercises/practice/linked-list/.meta/config.json
```diff
@@ -0,0 +1,17 @@
...
+      ".meta/example.ex"
+    ],
+    "solution": [
+      "lib/linked_list.ex"
+    ],
+    "test": [
+      "test/linked_list_test.exs"
+    ]
+  },
+  "blurb": "Implement a doubly linked list.",
```
neenjaw, line 15:
```suggestion
  "blurb": "Simulate a doubly linked list.",
```

exercises/practice/linked-list/lib/linked_list.ex
```diff
@@ -0,0 +1,52 @@
+defmodule LinkedList do
+  @type linked_list :: nil
```
neenjaw, line 2: I would consider giving the Linked list struct here in the stub exercise

exercises/practice/linked-list/test/linked_list_test.exs
```diff
@@ -0,0 +1,261 @@
...
+  test "push/pop respectively add/remove at the end of the list" do
+    linked =
+      LinkedList.new()
+      |> LinkedList.push(11)
+      |> LinkedList.push(13)
+
+    {value, linked} = LinkedList.pop(linked)
+    assert value == 13
+    {value, _} = LinkedList.pop(linked)
+    assert value == 11
```
neenjaw, line 24: @angelikatyborska how does the website work with multiple assertions? iirc correctly, didn't we have to limit it to one assertion per test?

neenjaw (approved):
Nice work, I think adding some clarifying comments that we are simulating this structure might help to guide the student

I have just a minor concern about the tests asserting more than once, not because that is wrong, but because I am unsure if we are able to handle that nicely on the website when its run with the test runner

exercises/practice/linked-list/.meta/config.json
angelikatyborska: Don't we need to use exactly the same blurb as in https://github.com/exercism/problem-specifications/blob/main/exercises/linked-list/metadata.yml#L2 ?

Configlet will soon have a feature that allows updating blurbs from problem specs. I can't wait to use it to add a lot of missing dots... it would overwrite a custom blurb.

exercises/practice/linked-list/test/linked_list_test.exs
angelikatyborska: No, no, it's completely fine. The only limitation we have is that concept exercise tests need to be self-contained within the `test do ... end` block because the file won't be shown in the editor.

exercises/practice/linked-list/.meta/config.json
neenjaw: Not sure about that tbh

exercises/practice/linked-list/test/linked_list_test.exs
neenjaw: ah, yes, that's the caveat

exercises/practice/linked-list/.meta/example.ex
```diff
@@ -0,0 +1,53 @@
...
+  """
+  @spec push(linked :: linked_list, value :: any) :: linked_list
+  def push(%LinkedList{back: back} = linked, value), do: %{linked | back: [value | back]}
+
+  @doc """
+  Remove a value at the back
+  """
+  @spec pop(linked :: linked_list) :: {any, linked_list}
+  def pop(%LinkedList{back: [value | back]} = linked), do: {value, %{linked | back: back}}
+  def pop(%LinkedList{front: front}), do: pop(%LinkedList{back: Enum.reverse(front)})
```
angelikatyborska, line 26: Reversing a part of the linked list when removing its first element somehow feels wrong. Like... against the whole concept of linked lists? I'm not an expert on this, I quit my CS course after a year 😊 

I need to think about this a bit longer. Double linked lists and Elixir, hmm. Could they be faked better with structs and maps? 

E.g. a linked list would be a struct with three fields: start, end, and nodes:
```
%LinkedList{
  start: 4,
  end: 19,
  nodes: %{
    4: %{value: "banana", next: 7, prev: 19},
    7: %{value: "apple", next: 19, prev: 4},
    19: %{value: "orange", next: 7, prev: 7},
  }
}
```

The numbers are some random identifiers.

neenjaw: Yea, that would probably more akin to how I might solve it but @jiegillet's solution still works, so still good by me

jiegillet: Well, of course you can use a map to simulate a linked list, or you could even more simply use a tuple:
```
linked = {"banana", "apple", "orange"}
```
and use `Tuple.insert_at`, `Tuple.delete_at` for `push` and `pop`, but using a very sophisticated built-in data structure defeats the purpose of implementing your own data structure. 

Using a singly linked list to implement a doubly linked list (or really, a queue) is a more fair game. Yes, indeed, I need to reverse lists once in a while, which is not in the spirit of singly linked lists (although I think if you study the complexity, it's still not so bad because you don't need to do that very often), but it's [standard](http://erlang.org/doc/man/queue.html).

It feels like students might be confused by that exercise, because many of them didn't study CS and even less studied functional data structures. So I'm thinking either we drop it, because doubly linked lists as they are defined are impossible in immutable languages, or we give more instructions on what we expect the student to do (after deciding what that is).

angelikatyborska: > because doubly linked lists as they are defined are impossible in immutable languages

In that case I would say we should drop it. We have enough nice exercises, and I don't want to encourage people to do things that don't make much sense in Elixir.

We already had a few complaints about `binary-search` because the only way to implement it while keeping its intended complexity is to use a tuple to store a collection, but you wouldn't do that in Elixir. Tuples are not for collections, but you can't do binary search on a list.

jiegillet: OK, sounds good to me. 
If there ever is an exercise to "build a queue" then we dig for this implementation.

angelikatyborska: To mark an exercise as "please do not implement now or ever", we would need to add it to the `"foregone"` list in `config.json`. There's already one exercise there.

angelikatyborska (approved):

angelikatyborska:
@neenjaw are you cool with forever banishing doubly linked lists from our track? Can I merge? :)

neenjaw:
Yelpers 👍🏻

Issue #560, opened 2021-01-28
[v3] Implement new Concept Exercise: `with`


angelikatyborska:
## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md)

## Goal

The goal of this exercise is to teach the student how the `with` special form works and when it is appropriate to use (e.g. to avoid nested `case`s)

## Learning objectives

- Know how to use `with` to chain matching clauses 
- Know how to use the `else` block in `with` to match on different kinds of errors
- Know when _not_ to use `with`

## Out of scope

?

## Concepts

- `with`

## Prerequisites

- `errors` (the `with` special form is most commonly used with error/ok-tuples and the exercise should teach that)
- `pattern-matching`
- other, depending on the story

## Resources to refer to

- https://hexdocs.pm/elixir/Kernel.SpecialForms.html#with/1
- https://elixir-lang.org/getting-started/mix-otp/docs-tests-and-with.html#with
- http://learningelixir.joekain.com/learning-elixir-with/
- https://openmymind.net/Elixirs-With-Statement/

## Analyzer

It will be necessary to use the analyzer to check that the `with` form was indeed used, and that no alternatives (e.g. `case`s) were used.

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

ThiagosLima:
Hi, I am still learning Elixir and want to help with some issues in the future. But for now, I have just a comment. 

This Analyzer part that assure students use the correct function could be added in some exercises. Like **Strain** exercise which it is not supposed to use `filter` and `reject`. And the **Accumulate** exercise that it is note supposed to use `collect`, `map` or `fmap`.

I don't know if it is already done in the v3. By the way, nice work with the Elixir track. It is really good!

neenjaw:
@ThiagosLima thanks for the comment!! Glad to have you here! Yes, that is the eventual goal to add this functionality to practice exercises at some point too. :)

[...]

PR #795, opened 2021-06-28
New practice exercise `sgf-parsing`

jiegillet:
New one!

This one might need a bit of discussion because I went a little nuts here. I _love_ using parser combinators in Elm and Haskell, so whenever I see a slightly non-trivial parsing problem, I want to use that. I believe there are such libraries in Elixir (NimbleParsec judging [from this article](https://serokell.io/blog/parser-combinators-in-elixir) that inspired me) but I felt like building my own, so I did. 
It was super fun, but I feel that most people will not use that technique, they would probably use regex or something. What is the go-to parsing method in Elixir? 

In any case, this might be acceptable because it's only an example implementation, but that makes it hard to pick good prerequisites and a difficulty level. What do you think?


angelikatyborska:
> I love using parser combinators in Elm and Haskell, so whenever I see a slightly non-trivial parsing problem, I want to use that.
> (...)
> What is the go-to parsing method in Elixir?

Honestly, this is a bit above my knowledge level 🙈. I'm just a simple web developer who doesn't know fancy algorithms... it's embarrassing but I have successfully avoided solving the `Markdown` exercise so far 🙈 @neenjaw how is your advanced string parsing in Elixir?

> I believe there are such libraries in Elixir but I felt like building my own, so I did.

That's a good call because you couldn't use an external dependency in a solution anyway 😉

> In any case, this might be acceptable because it's only an example implementation

It works, it fulfills the requirements, I don't understand it at all so I can't say more 😅 but it's fine with me

> but that makes it hard to pick good prerequisites and a difficulty level. What do you think?

That probably means we should require all of the usual things :) difficulty level 8 also sounds reasonable to me.

I'm only conflicted about the `"errors"` concept. I think it should be defined as "raising errors" and as such shouldn't be part of this exercise or `circular-buffer`. Sorry that I wasn't decisive about this earlier. But now that I look at the config, only exercises that explicitly `raise` something or have a test assertions about raising "practice" errors. The ok/error tuple pattern is introduces in the "tuples" concept.

Nitpick: there's some wild `.config.json.swo` file 👀

exercises/practice/sgf-parsing/.docs/instructions.md
```diff
@@ -0,0 +1,66 @@
...
+
+This is a tree with three nodes:
+
+- The top level node has three properties: FF\[4\] (key = "FF", value
+  = "4"), C\[root\](key = "C", value = "root") and SZ\[19\] (key =
+  "SZ", value = "19"). (FF indicates the version of SGF, C is a
+  comment and SZ is the size of the board.)
+  - The top level node has a single child which has a single property:
+    B\[aa\].  (Black plays on the point encoded as "aa", which is the
+    1-1 point (which is a stupid place to play)).
```
neenjaw, line 27: Is this really in the problem specs instructions?

```suggestion
    B\[aa\].  (Black plays on the point encoded as "aa", which is the
    1-1 point).
```

angelikatyborska: https://github.com/exercism/problem-specifications/pull/1814

exercises/practice/sgf-parsing/.meta/example.ex
```diff
@@ -0,0 +1,207 @@
...
+      |> map(&Enum.join(&1, ""))
+      |> validate(&(&1 == String.upcase(&1)), "property must be in uppercase"),
+      some(
+        char(?[)
+        |> error("properties without delimiter")
+        |> drop_and(many(escaped(&(&1 != ?]))))
+        |> drop(char(?]))
+        |> map(&Enum.join(&1, ""))
+      )
+    )
```
neenjaw, line 66: I think it would be easier to read/understand this block if it were split apart,

```suggestion
    node_property = 
      some(satisfy(&(&1 not in '[();')))
      |> map(&Enum.join(&1, ""))
      |> validate(&(&1 == String.upcase(&1)), "property must be in uppercase")

    node_value = 
      some(
        char(?[)
        |> error("properties without delimiter")
        |> drop_and(many(escaped(&(&1 != ?]))))
        |> drop(char(?]))
        |> map(&Enum.join(&1, ""))
      )

    lift2(&{&1, &2}, node_property, node_value)
```

exercises/practice/sgf-parsing/.meta/example.ex
```diff
@@ -0,0 +1,207 @@
...
+
+  def validate(parser, p, err) do
+    fn string ->
+      with {:ok, result, rest} <- parser.(string) do
+        if p.(result) do
+          {:ok, result, rest}
+        else
+          {:error, err, rest}
+        end
+      end
```
neenjaw, line 204: When there is a single expression in the `with`, is there a benefit to using `with` over

```suggestion
      {:ok, result, rest} = parser.(string)
      
      if p.(result) do
        {:ok, result, rest}
      else
        {:error, err, rest}
      end
```

neenjaw (approved):
Cool stuff, I've seen and used NimbleCSV and NimbleParsec before never implemented it myself.  Left a couple comments just what I thought of the example while reading it.  Given it is a proof of solvability, all non-blocking for me.


Great work

neenjaw:
The `*.swo` file is a vim swap file, accidental inclusion I would think

neenjaw:
> What is the go-to parsing method in Elixir?

For beginner/intermediate, which is what we are likely to see on exercism, I would guess that most people will end up using a combination of string lib functions and regex.

exercises/practice/sgf-parsing/.docs/instructions.md
jiegillet: I personally find it pretty amusing :)

exercises/practice/sgf-parsing/.meta/example.ex
jiegillet: The `with` here is important, because I need the error to be passed further if `parser.(string)` fails. If I use ` {:ok, result, rest} = parser.(string)` and the parser fails, I will get an exception.

exercises/practice/sgf-parsing/.docs/instructions.md
neenjaw: Ya, but we've made a conscious effort to remove this type of comment/test case from problems.

jiegillet:
> Nitpick: there's some wild `.config.json.swo` file 👀

> The `*.swo` file is a vim swap file, accidental inclusion I would think

Yes, I'm ashamed 😳 

> For beginner/intermediate, which is what we are likely to see on exercism, I would guess that most people will end up using a combination of string lib functions and regex.

OK, so leaving `regular-expressions` in the prerequisites in the right call.

>  it's embarrassing but I have successfully avoided solving the `Markdown` exercise so far 🙈 

Give it a try, it's not that hard, because you just need to refactor existing code :) 

> I'm only conflicted about the `"errors"` concept. I think it should be defined as "raising errors" and as such shouldn't be part of this exercise or `circular-buffer`. Sorry that I wasn't decisive about this earlier. But now that I look at the config, only exercises that explicitly `raise` something or have a test assertions about raising "practice" errors. The ok/error tuple pattern is introduces in the "tuples" concept.

OK, I will remove errors from the list. But for this particular exercise, handling errors was actually tricky, and I would call that proper handling practice, so I wish we had a concept for that.

exercises/practice/sgf-parsing/.docs/instructions.md
jiegillet: Yes, that's probably the right decision. What do I do for now? Remove it here and you do a PR? Or do we wait for your PR?
Is there a reason why the specifications are duplicated in the tracks repositories? I would think that a central one might work best, with appendices added by the tracks as necessary, which is kind of what's happening now anyway.

neenjaw: Problem specs will take time to approve, and afaik it doesn't sync yet.  I think there is a copy for the same reason as right now it doesn't pull the instructions centrally.  I think this is future work to be done, just hasn't been prioritized ahead of other product work for v3.

neenjaw (approved):

exercises/practice/sgf-parsing/.meta/example.ex
neenjaw: :+1:

angelikatyborska:
>  I would call that proper handling practice, so I wish we had a concept for that.

Maybe it's a topic we could explore more as part of the `with` concept? (Not yet implemented)

Speaking of which, if (when? 😇) we ever add that `with` concept, should that exercise require it and practice it? If yes, I will edit the description of https://github.com/exercism/elixir/issues/560 to mention that

jiegillet:
I think we should touch on it with the `with` concept. Now I'm having too much fun doing the exercises, but when they (soon) run out, I'm planning on taking a look at concepts. I would also like to have a `reduce` concepts, it's so powerful and so underused (looking at other students' code). 

However, I'm not sure this exercise is the best to link with the `with` concept. It's crucial to my implementation, but I would like to see other solutions first.

angelikatyborska:
It would be great to have more help with concepts and concept exercises 🙂 

`reduce` is part of the `enum` concept:
- introduction (available before solving a concept exercise): https://github.com/exercism/elixir/blob/9774308bb6b898df6ecbde64c5c6e48c651870ae/concepts/enum/introduction.md#L14-L20
- about (avaialbe after solving a concept exercise): https://github.com/exercism/elixir/blob/9774308bb6b898df6ecbde64c5c6e48c651870ae/concepts/enum/about.md#L33-L61

angelikatyborska (approved):

angelikatyborska:
It looks to me like @neenjaw's feedback has been all addressed, so I'm merging 🎉

PR #797, opened 2021-06-29
New practice exercise `square-root`

jiegillet:
I did this one in 20 minutes (after thinking researching the algorithm), pretty good feeling :)


exercises/practice/square-root/.docs/instructions.md
```diff
@@ -0,0 +1,9 @@
+# Description
+
+Given a natural radicand, return its square root.
+
+Note that the term "radicand" refers to the number for which the root is to be determined. That is, it is the number under the root symbol.
```
papey, line 5: We may have to add that this is a simplified exercise where only perfect squares are tested.

This way, we may avoid questions where the student asks what happens if functions takes, for example, '3'.

We may also add an anylizer ensuring that `:math.sqrt(x)` is not used. Note that maybe just a hint is ok here, what do you think @angelikatyborska

angelikatyborska: Yes, we definitely need to tell students not to use `:math.sqrt()`. I would recommend:
1. Adding an `instructions.append.md` file to tell students that they need to implement an algorithm on their own and not use `:math.sqrt()` or `Float.pow`
2. Adding an issue to the analyzer repo so that somebody works on it at some point, an issue similar to https://github.com/exercism/elixir-analyzer/issues/42 for example

exercises/practice/square-root/test/square_root_test.exs
```diff
@@ -0,0 +1,57 @@
+defmodule SquareRootTest do
+  use ExUnit.Case
+
+  # @tag :pending
+  test "root of 1" do
+    radicand = 1
+    output = SquareRoot.square_root(radicand)
```
angelikatyborska, line 7: "square root square root" :D sounds a bit silly. Maybe we can deviate from problem specs in this case and do, let's say `SquareRoot.calculate`?

exercises/practice/square-root/.meta/config.json
```diff
@@ -0,0 +1,18 @@
...
+    "solution": [
+      "lib/square_root.ex"
+    ],
+    "test": [
+      "test/square_root_test.exs"
+    ]
+  },
+  "blurb": "Given a natural radicand, return its square root.",
+  "source": "wolf99",
+  "source_url": "https://github.com/exercism/problem-specifications/pull/1582"
```
angelikatyborska, line 17: How meta 😂 I suspect this isn't really a "good" value for source URL but meh 🤷 we're just copying problem specifications.

exercises/practice/square-root/test/square_root_test.exs
neenjaw: ☝🏻

exercises/practice/square-root/.meta/config.json
jiegillet: Hahaha I didn't notice that!

exercises/practice/square-root/.docs/instructions.md
jiegillet: > We may have to add that this is a simplified exercise where only perfect squares are tested.

That will be apparent looking at the tests, I think. It's not uncommon for tests to have more information than the description :)

exercises/practice/square-root/test/square_root_test.exs
jiegillet: It might look better in the tests, but it will look weird in the lib file. How about `sqrt`?

angelikatyborska: My problem with `SquareRoot.square_root` and `SquareRoot.sqrt` is that in both cases both the module and the function name are exactly the same noun, shortened or not. I don't really know why, but it feels more correct to me if function name is either an action that gets executed on the "subject" indicated by the module name, or if the function name is a noun that's a specialization of the module name. I wonder if there's an article about this 🤔 I would like to know where my feeling comes from, maybe it's completely unjustified?

Here's more proposals. They all look ok to me 🤷 

```elixir
defmodule SquareRoot do
  @doc """
  Calculate the integer square root of a positive integer
  """
  @spec calculate(radicand :: pos_integer) :: pos_integer
  def calculate(radicand) do
  end
end
```
```elixir
defmodule Root do
  @doc """
  Calculate the integer square root of a positive integer
  """
  @spec square(radicand :: pos_integer) :: pos_integer
  def square(radicand) do
  end
end
```
```elixir
defmodule Math do
  @doc """
  Calculate the integer square root of a positive integer
  """
  @spec sqrt(radicand :: pos_integer) :: pos_integer
  def sqrt(radicand) do
  end
end
```
```elixir
defmodule Calculator do
  @doc """
  Calculate the integer square root of a positive integer
  """
  @spec sqrt(radicand :: pos_integer) :: pos_integer
  def sqrt(radicand) do
  end
end
```
```elixir
defmodule PosInteger do
  @doc """
  Calculate the integer square root of a positive integer
  """
  @spec sqrt(radicand :: pos_integer) :: pos_integer
  def sqrt(radicand) do
  end
end
```

angelikatyborska (approved):
🥳

neenjaw (approved):

PR #798, opened 2021-06-29
New practice exercise `variable-length-quantity`

jiegillet:
New one. 
Nothing too special, except that bit manipulation code is alway hard to read (or at least I can't make it legible).

### DIFF
--- a/exercises/practice/variable-length-quantity/lib/variable_length_quantity.ex
+++ b/exercises/practice/variable-length-quantity/lib/variable_length_quantity.ex
@@ -1,0 +1,15 @@
+defmodule VariableLengthQuantity do
+  @doc """
+  Encode integers into a series of VLQ encoded bytes
+  """
+  @spec encode(integers :: [integer]) :: [integer]
+  def encode(integers) do
+  end
+
+  @doc """
+  Decode a series of VLQ encoded bytes into a series of integers
+  """
+  @spec decode(bytes :: [integer]) :: {:ok, [integer]} | {:error, String.t()}
+  def decode(bytes) do
+  end
+end
--- a/exercises/practice/variable-length-quantity/lib/variable_length_quantity.ex
+++ b/exercises/practice/variable-length-quantity/lib/variable_length_quantity.ex
@@ -1,15 +1,15 @@
 defmodule VariableLengthQuantity do
   @doc """
-  Encode integers into a series of VLQ encoded bytes
+  Encode integers into a bitstring of VLQ encoded bytes
   """
-  @spec encode(integers :: [integer]) :: [integer]
+  @spec encode(integers :: [integer]) :: <<_::8>>
   def encode(integers) do
   end
 
   @doc """
-  Decode a series of VLQ encoded bytes into a series of integers
+  Decode a bitstring of VLQ encoded bytes into a series of integers
   """
-  @spec decode(bytes :: [integer]) :: {:ok, [integer]} | {:error, String.t()}
+  @spec decode(bytes :: <<_::8>>) :: {:ok, [integer]} | {:error, String.t()}
   def decode(bytes) do
   end
 end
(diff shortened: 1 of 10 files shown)


neenjaw:
What if we changed this exercise a bit, rather than working with a list of integers, what if we use bitstrings? So the exercise would be converting integers to VLQ encoded bitstrings, and VLQ encoded bitstrings to integers.  I don't think we have anything like this in our practice list, might be super interesting to promote a binary-based pattern matching exercise.

To encode the integer you can do something like this:

```elixir
integer = 78
<<encoded::7>> = <<integer::7>>
remainder = integer - encoded

if remainder == 0 do
  <<0::1, encoded::7, following::binary>>
else
  encode(remainder, <<1::1, encoded::7, following::binary>>)
end
```

angelikatyborska:
I'm with @neenjaw, bitstrings are the perfect choice for this exercise. We definitely need more practice exercises for bitstrings!

neenjaw (changes requested):
Just adding a request for changes based on: https://github.com/exercism/elixir/pull/798#issuecomment-870707967

jiegillet:
Excellent suggestion. It also made my code more readable :)

Also CI with 1.8 is failing, I'm not sure why. Any idea?

exercises/practice/variable-length-quantity/lib/variable_length_quantity.ex
```diff
@@ -0,0 +1,15 @@
...
+  Encode integers into a bitstring of VLQ encoded bytes
+  """
+  @spec encode(integers :: [integer]) :: <<_::8>>
+  def encode(integers) do
+  end
+
+  @doc """
+  Decode a bitstring of VLQ encoded bytes into a series of integers
+  """
+  @spec decode(bytes :: <<_::8>>) :: {:ok, [integer]} | {:error, String.t()}
```
neenjaw, line 12:
```suggestion
  @spec decode(bytes :: binary) :: {:ok, [integer]} | {:error, String.t()}
```

neenjaw (approved):
Nice solution, looks awesome. ` <<_::8>>` is equivalent to the `binary` type since a `binary` is always divisible by 8 bits.  `bitstrings` can be any number of bytes.

That's an interesting error.  I wonder if there has been a regression somewhere in older versions because I thought that would have worked.  Anyway, this seems to solve it with `use Bitwise` then using the infix operators:

```elixir
defmodule VariableLengthQuantity do
  use Bitwise

  @doc """
  Encode integers into a bitstring of VLQ encoded bytes
  """
  @spec encode(integers :: [integer]) :: <<_::8>>
  def encode(integers) when is_list(integers) do
    integers
    |> Enum.map(&encode/1)
    |> Enum.map(&String.reverse/1)
    |> Enum.join(<<>>)
  end

  def encode(0), do: <<0>>

  def encode(int, leading \\ 0)
  def encode(0, _), do: <<>>

  def encode(int, leading) do
    rest = int >>> 7
    <<leading::1, int::7, encode(rest, 1)::binary>>
  end

  @doc """
  Decode a bitstring of VLQ encoded bytes into a series of integers
  """
  @spec decode(bytes :: <<_::8>>) :: {:ok, [integer]} | {:error, String.t()}
  def decode(bytes, _status \\ :complete, acc \\ 0)
  def decode(<<>>, :complete, _), do: {:ok, []}
  def decode(<<>>, :incomplete, _), do: {:error, "incomplete sequence"}

  def decode(<<lead::1, byte::7, bytes::binary>>, _status, acc) do
    acc = (acc <<< 7) + byte

    if lead == 1 do
      decode(bytes, :incomplete, acc)
    else
      with {:ok, result} <- decode(bytes, :complete),
           do: {:ok, [acc | result]}
    end
  end
end
```

jiegillet:
Nice, the inline operator looks better anyway :)

angelikatyborska (approved):
Perfect 🥳 I'm so happy we finally have a bitstrings practice exercise

PR #800, opened 2021-06-30
New practice exercise `resistor-color-duo`

jiegillet:
New one. Quite simple. 
Big brother `resistor-color-trio` coming up next.

### DIFF
--- a/config.json
+++ b/config.json
@@ -2661,7 +2661,22 @@
           "bitstrings"
         ],
         "difficulty": 6
-      }
+       },
+       {
+        "slug": "resistor-color-duo",
+        "name": "Resistor Color Duo",
+        "uuid": "92387ac7-4c6d-4568-819e-e037b1d86a72",
+        "prerequisites": [
+          "atoms",
+          "lists",
+          "pattern-matching",
+          "maps"
+        ],
+        "practices": [
+          "maps"
+        ],
+        "difficulty": 2
+       }
     ],
     "foregone": [
       "linked-list",
--- a/exercises/practice/resistor-color-duo/.meta/example.ex
+++ b/exercises/practice/resistor-color-duo/.meta/example.ex
@@ -1,0 +1,22 @@
+defmodule ResistorColorDuo do
+  @colors %{
+    black: 0,
+    brown: 1,
+    red: 2,
+    orange: 3,
+    yellow: 4,
+    green: 5,
+    blue: 6,
+    violet: 7,
+    grey: 8,
+    white: 9
+  }
+
+  @doc """
+  Calculate a resistance value from two colors
+  """
+  @spec value(colors :: [atom]) :: integer
+  def value([a, b | _]) do
+    10 * @colors[a] + @colors[b]
+  end
+end
--- a/exercises/practice/resistor-color-duo/.docs/instructions.md
+++ b/exercises/practice/resistor-color-duo/.docs/instructions.md
@@ -1,0 +1,34 @@
+# Description
+
+If you want to build something using a Raspberry Pi, you'll probably use _resistors_.
+For this exercise, you need to know two things about them:
+
+* Each resistor has a resistance value.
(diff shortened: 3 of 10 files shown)


config.json
```diff
@@ -2661,7 +2661,22 @@
...
+        "uuid": "92387ac7-4c6d-4568-819e-e037b1d86a72",
+        "prerequisites": [
+          "atoms",
+          "lists",
+          "pattern-matching",
+          "maps"
+        ],
+        "practices": [
+          "maps"
+        ],
```
angelikatyborska, line 2677: Interesting. With those prerequisites, `resistor-color-duo` would be unlocked before `resistor-color` because `resistor-color` has an implementation with a list and `Enum.find_index(@colors, &(&1 == color))`, so it requires `enum`, and `enum` comes after all those prerequisites.

We cannot edit `resistor-color` to NOT use `enum` because it needs to return all the colors in order, and map keys aren't ordered.

I don't know if that's a problem, but it defies my expectations. I would think that `resistor-color` is easier than `resistor-color-duo`, and that `resistor-color-duo` is easier than `resistor-color-trio` 🤔

exercises/practice/resistor-color-duo/.meta/example.ex
```diff
@@ -0,0 +1,22 @@
...
+    blue: 6,
+    violet: 7,
+    grey: 8,
+    white: 9
+  }
+
+  @doc """
+  Calculate a resistance value from two colors
+  """
+  @spec value(colors :: [atom]) :: integer
```
angelikatyborska, line 18: I like that the colors are atoms. What do you think about editing `resistor-color` to also use atoms for the colors? Now they're strings.

angelikatyborska (approved):

config.json
jiegillet: But this one has `maps`, does that come after `lists` and `enum` or is it a different branch?
I could use a list implementation to make them all consistent. To be honest I didn't notice `resistor-color` existed haha.

exercises/practice/resistor-color-duo/.meta/example.ex
jiegillet: Leave it to me. Maybe we should make the trio all a little bit more consistent on other things as well?

neenjaw: 🤷🏻 [6 of one, half dozen of the other](https://en.wikipedia.org/wiki/Six_of_One#:~:text=The%20idiom%20%22six%20of%20one,matter%20which%20one%20we%20choose.)

config.json
angelikatyborska: `lists` and `maps` are required by `enum`. 

I think using a list for this exercise would make it less idiomatic. I guess we need to accept it like it is :)

neenjaw: > map keys aren't ordered.

100% agree with this statement, it is a common fallacy for students to believe that map keys are ordered just because they appear so in small maps.  

Not sure if this is a problem though.  This is somewhat similar to the "Instruments of Texas" theme that we reuse in the concept exercises as Maud had designed these exercises as teaching exercises. Each one just teaches something different.

neenjaw (approved):

config.json
jiegillet: Someone going through all three exercises in a row is likely to reuse the same implementation though. Either with maps or lists. I probably would have written `resistor-color ` with maps as well since lookup is `O(1)` rather than `O(n)`.

angelikatyborska:
I'll merge.
I'm willing to accept that duo and trio come before `resistor-color` when all have idiomatic solutions 😅 but I'll ask other tracks if they also have this "problem".

angelikatyborska:
```
Derk-Jan Karrenbeld says:
Didactily they must go after resistor-color. If returning the array complicates the exercise too much, I'd remove that.
```

PR #804, opened 2021-07-01
[#793] New practice exercise `word-search`

jiegillet:
As promised :)


jiegillet:
1.7 is failing. Any guess? `Enum.with_index`? My Elixir journey started with 1.9 I think :)

angelikatyborska:
I have all the minor Elixir versions installed with `asdf` just for this case. I'll check :D

angelikatyborska:
The error is not very clear. I have seen it before in some exercise:
```
$ mix test --include pending
Compiling 1 file (.ex)

== Compilation error in file lib/word_search.ex ==
** (exit) an exception was raised:
    ** (UndefinedFunctionError) function :erlang.get_stacktrace/0 is undefined or private
        :erlang.get_stacktrace()
        (elixir) lib/kernel/parallel_compiler.ex:214: anonymous fn/4 in Kernel.ParallelCompiler.spawn_workers/6


17:06:06.021 [error] Process #PID<0.106.0> raised an exception
** (UndefinedFunctionError) function :erlang.get_stacktrace/0 is undefined or private
    :erlang.get_stacktrace()
    (elixir) lib/kernel/parallel_compiler.ex:214: anonymous fn/4 in Kernel.ParallelCompiler.spawn_workers/6
```

angelikatyborska:
This is what I get with Elixir 1.7 and Erlang 23.0. I'm unable to install Erlang 22 or 21 because they're too old for my system version:
```
$ mix test --include pending
Compiling 1 file (.ex)

== Compilation error in file lib/word_search.ex ==
** (MatchError) no match of right hand side value: []

```

angelikatyborska:
I think `:reduce` for list comprehensions was only added in 1.8 but I can't confirm that for sure (source: https://github.com/elixir-lang/elixir/commit/5ce4b619915fa5ad7f9816eb7b5fc07055535f6c)

@neenjaw did we want to drop 1.7 if it starts go cause trouble?

jiegillet:
I'm using it twice. Not a bit deal to pipe the result into `Enum.reduce` I suppose.
I'll wait for a review before making changes though. Thank you as always :)

exercises/practice/word-search/test/word_search_test.exs
```diff
@@ -0,0 +1,463 @@
...
+    alxhpburyi
+    jalaycalmp
+    clojurermt
+    """
+
+    words = ["clojure"]
+    output = WordSearch.search(grid, words)
+
+    expected = %{
+      "clojure" => %Location{from: {10, 1}, to: {10, 7}}
```
angelikatyborska, line 122: I wonder if we should make the coordinates more explicit like this:
```suggestion
      "clojure" => %Location{from: %{row: 10, column: 1}, to: %{row: 10, column: 7}}
```

When I first saw `{10, 1}`, I thought: `{x, y}`, which is basically the opposite of what it really is 😁

neenjaw: ☝🏻 agreed, it isn't intuitive from a pair of numbers to know what they mean if not `{x, y}`

exercises/practice/word-search/lib/word_search.ex
```diff
@@ -0,0 +1,14 @@
+defmodule WordSearch do
+  defmodule Location do
+    defstruct [:from, :to]
+    @type t :: %Location{from: {integer, integer}, to: {integer, integer}}
```
neenjaw, line 4: Secondary to that comment in the test file, we should either standardize this to `{x, y}` or find a way to communicate an alternative meaning to the tuple pair

neenjaw (changes requested):
Just curious for your thoughts how we might be able to ease the expectation of knowing the interpretation of that tuple-pair, otherwise looks great

exercises/practice/word-search/test/word_search_test.exs
jiegillet: That's quite interesting, to me `{row, column}` is by far the most intuitive. I guess it's because I'm from a physics/math background and that's how you manipulate matrices. 
Originally the tests included explicitly row and column, I'll put them back in. Sorry about that!

angelikatyborska (approved):
Looks perfect to me now 🥳 @neenjaw?

neenjaw (approved):

PR #805, opened 2021-07-01
Exercise fix: Change colors to atoms and add @doc in `resistor-color`

jiegillet:
See discussion in #800. I know it wasn't a big deal, but it was also quick to do.


angelikatyborska (approved):
Perfect!

PR #807, opened 2021-07-02
Simplify resistor-color

angelikatyborska:
@SleeplessByte says the intention behind `resistor-color` was that it will be solved before `resistor-color-duo` and `resistor-color-trio`.

This PR simplifies `resistor-color` so that it can get unlocked before or at the same time as its big brothers.

It we don't like that, we could also modify the big brothers to add a `colors` function to each.

### DIFF
--- a/config.json
+++ b/config.json
@@ -861,8 +861,7 @@
           "enum"
         ],
         "practices": [
-          "charlists",
-          "maps"
+          "charlists"
         ],
         "difficulty": 2
       },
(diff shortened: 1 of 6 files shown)


angelikatyborska:
@jiegillet tagging you because GitHub doesn't allow me adding you as a reviewer

jiegillet (approved):
Looks good to me. 

I don't think we need to add the color list to all of them. Practicing map/list lookups makes sense, but for returning a constant you just typed, there is already `hello-world`.

SleeplessByte:
I support this PR

neenjaw (approved):
Sure. :)

PR #808, opened 2021-07-02
[#792] New practice exercise `two-bucket`

jiegillet:
One more.

### DIFF
--- a/config.json
+++ b/config.json
@@ -2708,15 +2708,12 @@
           "pattern-matching",
           "if",
           "cond",
-          "match",
+          "case",
           "structs",
           "enum",
           "recursion"
         ],
-        "practices": [
-          "atoms",
-          "integers"
-        ],
+        "practices": [],
         "difficulty": 7
        }
     ],
--- a/config.json
+++ b/config.json
@@ -2661,7 +2661,22 @@
           "bitstrings"
         ],
         "difficulty": 6
-      }
+       },
+       {
+        "slug": "resistor-color-duo",
+        "name": "Resistor Color Duo",
+        "uuid": "92387ac7-4c6d-4568-819e-e037b1d86a72",
+        "prerequisites": [
+          "atoms",
+          "lists",
+          "pattern-matching",
+          "maps"
+        ],
+        "practices": [
+          "maps"
+        ],
+        "difficulty": 2
+       }
     ],
     "foregone": [
       "linked-list",
--- a/config.json
+++ b/config.json
@@ -2695,6 +2695,29 @@
           "integers"
         ],
         "difficulty": 3
+       },
+       {
+        "slug": "two-bucket",
+        "name": "Two Bucket",
+        "uuid": "07b4679a-3d9f-42bc-ad1e-b9ba272a1c15",
+        "prerequisites": [
+          "atoms",
+          "tuples",
+          "integers",
+          "lists",
+          "pattern-matching",
+          "if",
+          "cond",
+          "match",
+          "structs",
+          "enum",
+          "recursion"
+        ],
+        "practices": [
+          "atoms",
+          "integers"
+        ],
+        "difficulty": 7
        }
     ],
     "foregone": [
--- a/config.json
+++ b/config.json
@@ -2676,6 +2676,25 @@
           "maps"
         ],
         "difficulty": 2
+       },
+       {
+        "slug": "resistor-color-trio",
+        "name": "Resistor Color Trio",
+        "uuid": "f1bad952-302b-4754-a439-a8c60c1d22a4",
+        "prerequisites": [
+          "atoms",
+          "tuples",
+          "integers",
+          "lists",
+          "pattern-matching",
+          "if",
+          "maps"
+        ],
+        "practices": [
+          "atoms",
+          "integers"
+        ],
+        "difficulty": 3
        }
     ],
     "foregone": [
--- a/exercises/practice/two-bucket/.meta/example.ex
+++ b/exercises/practice/two-bucket/.meta/example.ex
@@ -1,6 +1,6 @@
 defmodule TwoBucket do
-  defstruct [:goal, :moves, :other_bucket]
-  @type t :: %TwoBucket{goal: :one | :two, moves: integer, other_bucket: integer}
+  defstruct [:bucket_one, :bucket_two, :moves]
+  @type t :: %TwoBucket{bucket_one: integer, bucket_two: integer, moves: integer}
 
   @doc """
   Find the quickest way to fill a bucket with some amount of water from two buckets of specific sizes.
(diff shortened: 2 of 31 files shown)


config.json
```diff
@@ -2695,6 +2695,29 @@
...
+        "uuid": "07b4679a-3d9f-42bc-ad1e-b9ba272a1c15",
+        "prerequisites": [
+          "atoms",
+          "tuples",
+          "integers",
+          "lists",
+          "pattern-matching",
+          "if",
+          "cond",
+          "match",
```
angelikatyborska, line 2711: This isn't a concept 🤔 what did you mean by "match"?

config.json
```diff
@@ -2695,6 +2695,29 @@
...
+          "structs",
+          "enum",
+          "recursion"
+        ],
+        "practices": [
+          "atoms",
+          "integers"
+        ],
```
angelikatyborska, line 2719: Those are some pretty basic concepts for a rather complex exercise 🤔 and the check is failing for integers... maybe we could leave it completely empty? Or we say this is specifically a tail call recursion exercise and it practices and requires `tail-call-recursion`?

exercises/practice/two-bucket/.meta/example.ex
```diff
@@ -0,0 +1,55 @@
...
+    [
+      # Empty a bucket
+      {0, fill_two},
+      {fill_one, 0},
+      # Fill a bucket
+      {fill_one, size_two},
+      {size_one, fill_two},
+      # Pour one into the other
+      {min(size_one, fill_one + fill_two), max(0, fill_two - size_one + fill_one)},
+      {max(0, fill_one - size_two + fill_two), min(size_two, fill_one + fill_two)}
```
angelikatyborska, line 51: Normally I don't comment on the example solution, but I'm actually struggling to understand this exercise. Could you explain to me why `min` and `max` are necessary? I would think that when you pour from one bucket into the other until the first one is full, there is only one possible outcome.

exercises/practice/two-bucket/.meta/example.ex
```diff
@@ -0,0 +1,55 @@
...
+  def pour([{{other, goal}, moves} | _], _, _, goal),
+    do: {:ok, %TwoBucket{goal: :two, moves: moves, other_bucket: other}}
+
+  def pour([{state, moves} | states], forbidden_states, next_moves, goal) do
+    next =
+      next_moves.(state)
+      |> Enum.reject(&MapSet.member?(forbidden_states, &1))
+      |> Enum.map(&{&1, moves + 1})
+
+    pour(states ++ next, MapSet.put(forbidden_states, state), next_moves, goal)
```
angelikatyborska, line 38: This is just a generic worry, I haven't analyzed the solution in detail, so it might be misguided, but:

Appending to a list in a loop? Are you sure this is the best way?

exercises/practice/two-bucket/.meta/example.ex
```diff
@@ -0,0 +1,55 @@
+defmodule TwoBucket do
+  defstruct [:goal, :moves, :other_bucket]
+  @type t :: %TwoBucket{goal: :one | :two, moves: integer, other_bucket: integer}
```
neenjaw, line 3: Is this the clearest interface?

Would it not be clearer for a student studying the problem to retun their answer using a struct `%TwoBucker{bucket_one: x, bucket_two: y, moves: z}`

config.json
jiegillet: I think I mean `case`, I was pretty tired when I wrote that -_-

config.json
jiegillet: This is not really a tail call recursion, it's more of a breadth first search algorithm. I don't remember why I wrote these two. I think it's best to leave empty.

exercises/practice/two-bucket/.meta/example.ex
jiegillet: No worries, I'm happy to explain. There is only one possible outcome indeed, but you have to be careful how to calculate it  and I'm using min/max as a shortcut not to use an `if` statement. For example, imagine we want to pour the first bucket in the other one (line 51). There are 2 cases:
1) There is enough room in the second bucket for all the water => result is 0 in the first bucket
2) There is not enough room in the second bucket => result is current fill `fill_one` minus the remaining room in the second bucket `size_two - fill two`. 
In general, I can calculate `fill_one - (size_two - fill_two)` and if it's negative, it means that there was extra room in the second bucket but since columns can't be negative, I need to cap that to 0. Hence `max(0, fill_one - (size_two - fill_two))`. If the same as `if fill_one > size_two - fill_two, do: 0, else: fill_one - (size_two - fill_two) `.

It's a similar process for the other bucket, but I use min to prevent over-fullness. 
Does that make sense? Should I use an `if` statement instead?

exercises/practice/two-bucket/.meta/example.ex
jiegillet: Yes, because this is a breadth first search. I need to look at states with a lower number of moves first, and the next states I just created last. Otherwise I would need to generate all paths and check for the lowest after. 
A queue is the best data structure for this, but it's not worth implementing in this solution I think (and also the standard FP queue is made out of two lists and use that infamous `Enum.reverse` that you don't like :p ).

exercises/practice/two-bucket/.meta/example.ex
jiegillet: It's the interface suggested in the tests. But you are right it's not super intuitive, I'll change it. Thank you!

neenjaw: Cool, I think things like this are good reasons to break continuity with the problem specs repo

config.json
jiegillet: Oh I remember now, copy paste :(
Sorry about that!

exercises/practice/two-bucket/.meta/example.ex
angelikatyborska: Agreed! It reads much better now

exercises/practice/two-bucket/.meta/example.ex
angelikatyborska: Makes sense now, thanks! No, there's no need to change it to an `if` :)

exercises/practice/two-bucket/.meta/example.ex
angelikatyborska: All good. My comment about two lists and `Enum.reverse` was purely in the context of double linked lists 👻.

If you need a queue, use Erlang's (http://erlang.org/doc/man/queue.html). It has functions for adding and removing from both ends. And yes, it's two lists 😁

angelikatyborska (approved):
Did I mention how happy I am that you're adding all of those complex exercises so I don't have to solve them myself? 🥳

neenjaw (approved):

jiegillet:
No worries, I'm a problem solver, in this case quite literally :)

PR #811, opened 2021-07-04
[#787] New practice exercise `knapsack`

### LINKED ISSUE #787: New practice exercise `knapsack`
# New practice exercise: `knapsack`

Implement the `knapsack` practice Exercise in the Elixir track as it is defined in: [problem-specifications/exercises/knapsack](https://github.com/exercism/problem-specifications/tree/main/exercises/knapsack)

This exercise will most likely require an implementation with processes, so that the buffer can be stateful, but the state be hidden from the user.

### 1. New mix project

Create a new mix project in `exercises/practice/<exercise-slug>` and configure it in the same way as other exercises (with regard to Elixir version, `.formatter.exs`, `test/test_helper.exs` etc.). The easiest way to achieve this is to copy an existing exercise.

You might also try to use the script `bin/bootstrap_practice_exercise.exs` that you will find in this repository. It's an experimental approach of creating a skeleton of a new exercise that might or might not work for this exercise. It automates steps 2 and 3, and partially steps 4 and 6.

### 2. Instructions

Copy `description.md` from problem specifications into `.docs/instructions.md` of the new exercise. Do not edit it.

[What to do if you need to add Elixir specific instructions](https://github.com/exercism/docs/blob/main/building/tracks/practice-exercises.md#file-docsintroductionappendmd).

### 3. Exercise config

Create `.meta/config.json`. The fields `blurb`, `source`, and `source_url` should be copied from problem specifications, from the file `metadata.yml`. Each field can be skipped if it doesn't exist in `metadata.yml`.

[More details about the exercise config](https://github.com/exercism/docs/blob/main/building/tracks/practice-exercises.md#file-metaconfigjson).

Make sure to give yourself credit for being the author :muscle:.

### 4. Tests

Create a test suite.

- The test suite should implement all test cases as described in `canonical-data.json` in the problem specifications repository for this exercise. Remember that this file describes the exercise for all tracks. You will need to interpret in a way that makes sense for Elixir.
- [See instructions here about how to read `canonial-data.json`](https://github.com/exercism/problem-specifications#test-data-format)
- Avoid omitting any test cases from `canonial-data.json` unless there is a good Elixir-specific reason to do so.
- You can come up with more test cases than `canonial-data.json` describes, especially if there's something Elixir-specific that needs to be tested.
- Create a `.meta/tests.toml` file that documents which tests from `canonial-data.json` were implemented. This can be done automatically with [`configlet sync`](https://github.com/exercism/elixir/blob/adcbf3a31b4eb79a65268c79a40555b08bf621ce/CONTRIBUTING.md#L72-L81).

### 5. Example solution

Write an example solution in `.meta/example.ex`. It needs to pass all tests added in the previous step. It doesn't have to be beautiful, it only has to prove that the Exercise can be solved in all supported Elixir versions.

### 6. Solution stub

Write a solution stub in `lib/<exercise_name>.ex`. It should probably include empty function definitions for all the functions required by the test suite, each with a `@doc` and a `@spec`.

### 7. Optionally, document your decisions

If there is something unusual about this exercise, you can create a `.meta/design.md` file and describe the problem there.

### 8. Add exercise to track config

Open up the file `config.json` at the root of this repository. Find the list under the key `exercises.practice` and add your new exercise at the end of that list. A unique id can be generated with [uuidgenerator.net/version4](https://uuidgenerator.net/version4). You will need to decide how difficult the exercise is on the scale from 1 to 9, but also which concepts must be known by the student to solve this exercise well. Leave this field empty if you're not sure. A track maintainer can help you with that.

[More about this config](https://github.com/exercism/docs/blob/main/building/tracks/practice-exercises.md#metadata).

## Stuck? Questions?

Do not hesitate to tag @angelikatyborska (me) and ask for help.

jiegillet:
closes #787.

I'm very slightly uneasy about that one, the knapsack problem is super famous, I was bracing myself for slow results, having to use memoization... And everything finished instantly on the first try of my naive implementation. Well, I guess I'm overthinking this. 

I also updated a couple of lines in `bin/bootstrap_practice_exercise.exs` to write variable names in snake case, I noticed it wasn't happening with this exercise. I can do a separate PR if it's needed.

### DIFF
--- a/bin/bootstrap_practice_exercise.exs
+++ b/bin/bootstrap_practice_exercise.exs
@@ -26,8 +26,11 @@ defmodule Generate do
         variables:
           if match?(%{}, input) do
             Enum.map(input, fn
-              {var, %{} = val} -> {var, Enum.map(val, fn {v, _} -> v end)}
-              {var, _} -> {var, nil}
+              {var, %{} = val} ->
+                {Macro.underscore(var), Enum.map(val, fn {v, _} -> Macro.underscore(v) end)}
+
+              {var, _} ->
+                {Macro.underscore(var), nil}
             end)
           else
             [{"input", nil}]
@@ -78,7 +81,10 @@ defmodule Generate do
   end
 
   def print_input(%{} = input),
-    do: Enum.map_join(input, "\n", fn {variable, value} -> "#{variable} = #{inspect(value)}" end)
+    do:
+      Enum.map_join(input, "\n", fn {variable, value} ->
+        "#{Macro.underscore(variable)} = #{inspect(value)}"
+      end)
 
   def print_input(input), do: "input = #{inspect(input)}"
 
--- a/config.json
+++ b/config.json
@@ -2735,6 +2735,22 @@
         "difficulty": 7
        },
        {
+        "slug": "knapsack",
+        "name": "Knapsack",
+        "uuid": "1770aab6-2289-4134-b6ac-7d346e421e06",
+        "prerequisites": [
+          "atoms",
+          "integers",
+          "lists",
+          "maps",
+          "pattern-matching",
+          "recursion",
+          "guards"
+        ],
+        "practices": [],
+        "difficulty": 6
+       },
+       {
         "slug": "word-search",
         "name": "Word Search",
         "uuid": "41ab29d0-e094-4910-bb51-b99bdedc7f17",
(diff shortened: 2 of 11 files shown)


neenjaw (approved):

angelikatyborska (approved):
I think the difference between the typical "knapsack problem" and what we have here is that in the typical problem, each item has an infinite number of copies. Your solution is a very nice example of divide and conquer, I'm not worried 🙂. Very nice exercise!

PR #814, opened 2021-07-05
[#788] New practice exercise `house`

jiegillet:
One of @neenjaw's favorite singy-song type. 
Brace yourself, there is [another one coming](https://github.com/exercism/elixir/issues/786) :)


neenjaw:
Nice work!

I would vote that we don't add any more of this type of exercise after this is merged. We already have several of the verse type of exercise and I don't think each offers much over the last.

neenjaw (approved):

jiegillet:
Woops, already submitted the last one. Feel free to refuse.

neenjaw:
🤷🏻 If the work is already done, then it's alright

angelikatyborska:
I think @jiegillet's goal is to add ALL of the exercises 😁 and I think the goal will be achieved after all the current PRs get merged, excluding the exercises without canonical data.

exercises/practice/house/.meta/example.ex
```diff
@@ -0,0 +1,38 @@
...
+
+  def verse(n) do
+    stuff =
+      @protagonists
+      |> Enum.take(n)
+      |> Enum.reverse()
+      |> Enum.join(" ")
+
+    """
+    This is #{stuff}.
```
angelikatyborska, line 35: "This is stuff" 😂

config.json
```diff
@@ -2759,6 +2759,18 @@
...
+        "slug": "house",
+        "name": "House",
+        "uuid": "5ff34497-3a3f-4aaf-b50c-872e7a7b8f51",
+        "prerequisites": [
+          "lists",
+          "enum",
+          "strings"
+        ],
+        "practices": [],
+        "difficulty": 4
```
angelikatyborska, line 2773: Same as beer song IMO
```suggestion
        "difficulty": 3
```

config.json
```diff
@@ -2759,6 +2759,18 @@
...
         "difficulty": 7
+       },
+       {
+        "slug": "house",
+        "name": "House",
+        "uuid": "5ff34497-3a3f-4aaf-b50c-872e7a7b8f51",
+        "prerequisites": [
+          "lists",
+          "enum",
+          "strings"
```
angelikatyborska, line 2770: `start..stop` 🙂 
```suggestion
          "strings",
          "ranges"
```

angelikatyborska (approved):

exercises/practice/house/.meta/example.ex
jiegillet: I usually try to give meaningful variables names, but I couldn't resist this one :D

exercises/practice/house/.meta/config.json
```diff
@@ -0,0 +1,18 @@
+{
+  "authors": ["jiegillet"],
+  "contributors": [],
```
jiegillet, line 3:
```suggestion
  "contributors": [ "angelikatyborska" ],
```

jiegillet:
> I think @jiegillet's goal is to add ALL of the exercises 😁 and I think the goal will be achieved after all the current PRs get merged, excluding the exercises without canonical data.

Actually, there are three more: `react`, `rest-api` and `micro-blog`.

1. I believe you mentioned you don't want `micro-blog` because it's too easy, so please confirm that I shouldn't bother. 
2. I just finished the implementation of `react` (understanding the problem and creating the tests was more challenging). PR coming soon.
3. I have a question about `rest-api`. This is obviously catering to the type of people who learn Elixir so it's quite important to do right, I think. The exercise is based on [passing JSON files](https://github.com/exercism/problem-specifications/blob/main/exercises/rest-api/description.md) to an API. I'm not supposed to use dependencies for the examples solution, but can I expect the students to add `jason` to the project?

angelikatyborska:
1. Correct, I don't want `micro-blog`. There is a comment in canonical data that says
   > "This exercise is probably too easy in languages that use Unicode aware string slicing."
   This applies to Elixir.
2. Alrighty!
3. No, external dependencies are not allowed. We want to avoid them so that we don't have to deal with versions mismatches of the dependency and Elixir. Our test runner cannot handle external dependencies either. AFAIK our test runner runs in a docker container without a network interface and that somehow makes it 10x faster (@neenjaw would know more about this). So we would need a really really good reason to add the first external dependency, and I'm not sure if a single exercise is enough.

The rest-api exercise is a hard one, I'm not sure what to do about it 🤔. Parsing JSONs without a library doesn't sound fun to me. I wonder if we should change it so that it's more about OTP? Or maybe even about using `mnesia` as a database?

jiegillet:
Mmh I see. We could always pretend and send maps instead. And tell students why we did this in an appendix.

neenjaw:
> 3. No, external dependencies are not allowed. We want to avoid them so that we don't have to deal with versions mismatches of the dependency and Elixir. Our test runner cannot handle external dependencies either. 

These are the biggest reasons, any dependency that would be fetched, we would have to inject into the container which I think would be tedious and a chore to maintain. I'm not sure if compiled beam files are loaded based on version or just by module name, so I would imagine the chance of a version mismatch causing mysterious failures here to be high.

> AFAIK our test runner runs in a docker container without a network interface and that somehow makes it 10x faster (@neenjaw would know more about this). 

No network access is a requirement of the container to run untrusted code. The beam VM behaves strangely in containers without a network interface (particularly alpine-linux) and the VM  warmup time is 2-3 seconds slower than than when there is a network interface. So the elixir container is run with a stubbed network interface that does not connect to the internet. 

> So we would need a really really good reason to add the first external dependency, and I'm not sure if a single exercise is enough.

I lean to keep exercism's elixir track about problems and not projects. Requiring dependencies should be an indicator that a problem is outside of this scope. 

> The rest-api exercise is a hard one, I'm not sure what to do about it 🤔. Parsing JSONs without a library doesn't sound fun to me. I wonder if we should change it so that it's more about OTP? Or maybe even about using `mnesia` as a database?

I'm not super familiar with this exercise, but building a rest-api seems to fit in more of the "project" rather than "problem", because likely the things needed to do this canonically in elixir is `cowboy` and `plug`.

jiegillet:
Thank you for the feedback. I will have a closer look at the exercise and we can discuss on Slack, maybe.

Issue #559, opened 2021-01-28
Implement new Concept Exercise: `genserver`


angelikatyborska:
## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md)

## Goal

The student should understand the GenServer behaviour and know how to use it, writing a public module API and implementing the behaviour's callbacks.

## Learning objectives

- Know about the GenServer behaviour and its callbacks.
- Know that the name stands for "generic server".
- Understand the difference between `handle_call`, `handle_cast`, and `handle_info`.
- Understand the possible return values of the callbacks.
- Know it's a good idea to have a catch-all `handle_info` to avoid an ever-growing process inbox.
- Know that GenServers usually expose an easy-to-use public API instead of expecting its users to call `GenServer.call` themselves.
- Know when *not* to use GenServer (if `Agent` is good enough, if there is no need to keep state)

## Out of scope

- process linking, monitoring, `:DOWN` messages
- process registry, process names

## Concepts

- `genserver`

## Prerequisites

- `atoms`
- `tuples`
- `processes`
- `pid`
- `agent` (it will be easier to understand after having learned about agents)
- `use` (as in `use GenServer`)
- `behaviours`

## Practice exercises

Those practice exercises should have `genserver` in their prerequisites:
- `bank-account`
- `react`

## Resources to refer to

- https://elixir-lang.org/getting-started/mix-otp/genserver.html
- https://hexdocs.pm/elixir/master/GenServer.html

## Story

It might be a good idea to either duplicate or come up with a continuation of the stories used for the `processes` exercise (take-a-number machine) or the `agent` exercise (community garden). Ideally the story would force the student to use all three types of callbacks.

## Analyzer

The analyzer should check that indeed a module using the GenServer behaviour was implemented, and not a generic receive loop or an Agent.

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

PR #820, opened 2021-07-06
New practice exercise `react`

jiegillet:
This is one a bit harder than the last two. Even combined :)

I would welcome your opinion especially concerning the API, choice of functions, return types and such. For example, I'm sending back the callbacks in a map, but I'm not sure that is idiomatic. 

I chose to use cell names as strings instead of atoms, because I know that atoms are not garbage collected and this felt like the kind of system where you would programmatically create new cells, even though it doesn't happen here. But atoms are cleaner, so I'm on the fence.

I left some of the test comments in, because they gave me  some insight into the problem, but I can remove them if needed.

Difficulty around 8 or 9 sounds right, but I think I'm mixing the difficulty of designing the tests.

### DIFF
--- a/config.json
+++ b/config.json
@@ -2685,6 +2685,34 @@
         "difficulty": 8
       },
       {
+        "slug": "react",
+        "name": "React",
+        "uuid": "e4462e0c-b396-459c-815a-993a0f04c46a",
+        "prerequisites": [
+          "atoms",
+          "tuples",
+          "lists",
+          "enum",
+          "case",
+          "cond",
+          "if",
+          "multiple-clause-functions",
+          "pattern-matching",
+          "anonymous-functions",
+          "default-arguments",
+          "guards",
+          "maps",
+          "structs",
+          "pids",
+          "processes",
+          "recursion"
+        ],
+        "practices": [
+          "processes"
+        ],
+        "difficulty": 8
+      },
+      {
         "slug": "sgf-parsing",
         "name": "SGF Parsing",
         "uuid": "f87d00fa-99c4-4fda-9d90-d9c4d569a82e",
@@ -2773,34 +2801,6 @@
           "recursion"
         ],
         "difficulty": 9
-      },
-      {
-        "slug": "react",
-        "name": "React",
-        "uuid": "e4462e0c-b396-459c-815a-993a0f04c46a",
-        "prerequisites": [
-          "atoms",
-          "tuples",
-          "lists",
-          "enum",
-          "case",
-          "cond",
-          "if",
-          "multiple-clause-functions",
-          "pattern-matching",
-          "anonymous-functions",
-          "default-arguments",
-          "guards",
-          "maps",
-          "structs",
-          "pids",
-          "processes",
-          "recursion"
-        ],
-        "practices": [
-          "processes"
-        ],
-        "difficulty": 8
       },
       {
         "slug": "zebra-puzzle",
--- a/config.json
+++ b/config.json
@@ -2806,6 +2806,34 @@
           "list-comprehensions"
         ],
         "difficulty": 7
+       },
+       {
+        "slug": "react",
+        "name": "React",
+        "uuid": "e4462e0c-b396-459c-815a-993a0f04c46a",
+        "prerequisites": [
+          "atoms",
+          "tuples",
+          "lists",
+          "enum",
+          "case",
+          "cond",
+          "if",
+          "multiple-clause-functions",
+          "pattern-matching",
+          "anonymous-functions",
+          "default-arguments",
+          "guards",
+          "maps",
+          "structs",
+          "pids",
+          "processes",
+          "recursion"
+        ],
+        "practices": [
+          "processes"
+        ],
+        "difficulty": 8
        }
     ],
     "foregone": [
--- a/exercises/practice/react/.meta/example.ex
+++ b/exercises/practice/react/.meta/example.ex
@@ -120,7 +120,7 @@ defmodule React do
   @impl true
   def handle_cast({:remove_callback, name, callback}, %State{cells: cells} = state) do
     %OutputCell{callbacks: callbacks} = cell = cells[name]
-    callbacks = Map.delete(callbacks, name)
+    callbacks = Map.delete(callbacks, callback)
     {:noreply, %{state | cells: Map.put(cells, name, %{cell | callbacks: callbacks})}}
   end
 
--- a/exercises/practice/react/.meta/example.ex
+++ b/exercises/practice/react/.meta/example.ex
@@ -12,7 +12,7 @@ defmodule React do
     @type t :: %OutputCell{
             name: String.t(),
             inputs: [String.t()],
-            compute: (... -> any),
+            compute: fun(),
             value: any,
             callbacks: %{String.t() => (String.t(), any -> :ok)}
           }
@@ -23,7 +23,7 @@ defmodule React do
   @doc """
   Start a reactive system
   """
-  @spec new(cells :: [InputCell.t() | OutputCell.t()]) :: {:ok, pid}
+  @spec new(cells :: [%{atom => any}]) :: {:ok, pid}
   def new(cells) do
     GenServer.start_link(React, cells)
   end
@@ -73,7 +73,14 @@ defmodule React do
 
   @impl true
   def init(cells) do
-    cells = Map.new(cells, fn cell -> {cell.name, cell} end)
+    cells =
+      Map.new(cells, fn
+        %{type: :input} = cell ->
+          {cell.name, %InputCell{name: cell.name, value: cell.value}}
+
+        %{type: :output} = cell ->
+          {cell.name, %OutputCell{name: cell.name, inputs: cell.inputs, compute: cell.compute}}
+      end)
 
     initialized_cells =
       Map.new(cells, fn {name, cell} -> {name, initialize_value(cell, cells)} end)
--- a/exercises/practice/react/.meta/example.ex
+++ b/exercises/practice/react/.meta/example.ex
@@ -1,0 +1,164 @@
+defmodule React do
+  use GenServer
+
+  defmodule InputCell do
+    defstruct [:name, :value]
+    @type t :: %InputCell{name: String.t(), value: any}
+  end
+
+  defmodule OutputCell do
+    defstruct [:name, :inputs, :compute, :value, callbacks: []]
+
+    @type t :: %OutputCell{
+            name: String.t(),
+            inputs: [String.t()],
+            compute: (... -> any),
+            value: any,
+            callbacks: [String.t()]
+          }
+  end
+
+  # CLIENT SIDE
+
+  @doc """
+  Start a reactive system
+  """
+  @spec new(cells :: [InputCell.t() | OutputCell.t()]) :: {:ok, pid}
+  def new(cells) do
+    GenServer.start_link(React, cells)
+  end
+
+  @doc """
+  Return the value of an input or output cell
+  """
+  @spec get_value(cells :: pid, cell :: String.t()) :: any()
+  def get_value(cells, cell) do
+    GenServer.call(cells, {:get_value, cell})
+  end
+
+  @doc """
+  Set the value of an input cell
+  """
+  @spec set_value(cells :: pid, cell :: String.t(), value :: any) :: %{String.t() => any}
+  def set_value(cells, cell, value) do
+    GenServer.call(cells, {:set_value, cell, value})
+  end
+
+  @doc """
+  Add a callback to an output cell
+  """
+  @spec add_callback(cells :: pid, cell :: String.t(), callback :: String.t()) :: :ok
+  def add_callback(cells, cell, callback) do
+    GenServer.cast(cells, {:add_callback, cell, callback})
+  end
+
+  @doc """
+  Remove a callback from an output cell
+  """
+  @spec remove_callback(cells :: pid, cell :: String.t(), callback :: String.t()) :: :ok
+  def remove_callback(cells, cell, callback) do
+    GenServer.cast(cells, {:remove_callback, cell, callback})
+  end
+
+  # SERVER SIDE
+
+  defmodule State do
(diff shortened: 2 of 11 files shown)


config.json
```diff
@@ -2806,6 +2806,34 @@
...
+          "default-arguments",
+          "guards",
+          "maps",
+          "structs",
+          "pids",
+          "processes",
+          "recursion"
+        ],
+        "practices": [
+          "processes"
```
angelikatyborska, line 2834: I made a note in https://github.com/exercism/elixir/issues/559 that this will be a genserver exercise once we have that concept 🙂

exercises/practice/react/.meta/example.ex
```diff
@@ -0,0 +1,164 @@
...
+  defmodule InputCell do
+    defstruct [:name, :value]
+    @type t :: %InputCell{name: String.t(), value: any}
+  end
+
+  defmodule OutputCell do
+    defstruct [:name, :inputs, :compute, :value, callbacks: []]
+
+    @type t :: %OutputCell{
+            name: String.t(),
```
angelikatyborska, line 13: Do you think cell names should be atoms instead?

angelikatyborska (approved):
That's a really fascinating exercise 🤔 I'm really curious how people will approach it, and I'm very happy we have more genserver!

exercises/practice/react/test/react_test.exs
```diff
@@ -0,0 +1,227 @@
...
+    React.add_callback(cells, "output", "callback1")
+    callbacks = React.set_value(cells, "input", 4)
+    assert callbacks["callback1"] == 10
+  end
+
+  @tag :pending
+  test "callbacks should not be called if dependencies change but output value doesn't change" do
+    # Some incorrect implementations simply mark a compute cell as dirty when a dependency changes,
+    # then call callbacks on all dirty cells.
+    # This is incorrect since the specification indicates only to call callbacks on change.
```
angelikatyborska, line 204: I think those comments are very helpful for understanding the exercise

exercises/practice/react/.meta/example.ex
angelikatyborska: > I chose to use cell names as strings instead of atoms, because I know that atoms are not garbage collected and this felt like the kind of system where you would programmatically create new cells, even though it doesn't happen here. But atoms are cleaner, so I'm on the fence.

I agree with the concern, let's keep strings. Maybe you could describe this in `.meta/design.md`? for future maintainer generations :)

jiegillet: That's a good suggestion, I will do that. Those notes are only meant for maintainers right?

exercises/practice/react/lib/react.ex
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  defmodule OutputCell do
+    defstruct [:name, :inputs, :compute, :value, :callbacks]
+
+    @type t :: %OutputCell{
+            name: String.t(),
+            inputs: [String.t()],
+            compute: (... -> any),
+            value: any,
```
angelikatyborska, line 14: If we can allow cells to have any type, should we add a custom test case that tests string cells for example? or boolean cells? (but no error handling of type mismatch)

Problem specifications say:
> If your language supports generics, you may consider allowing reactors to act on other types.
> Tests for that are not included here.

exercises/practice/react/lib/react.ex
```diff
@@ -0,0 +1,53 @@
...
+
+  @doc """
+  Return the value of an input or output cell
+  """
+  @spec get_value(cells :: pid, cell :: String.t()) :: any()
+  def get_value(cells, cell) do
+  end
+
+  @doc """
+  Set the value of an input cell
```
angelikatyborska, line 34: Maybe this needs a hint about what the function returns?

exercises/practice/react/.meta/example.ex
angelikatyborska: Yes

exercises/practice/react/lib/react.ex
jiegillet: Ah, sure, why not! 
Any particular procedure for adding tests?

angelikatyborska: Since those are our own, we don't need to add anything to the toml file. Maybe mention that we did that in `design.md` but otherwise nothing special.

exercises/practice/react/lib/react.ex
jiegillet: Ah, the callbacks. Yes, good idea.
Also I have a question about this. It is specified that you can only have specific callback once, so I picked a map as the data structure since the uniqueness of the keys would always enforce that. But that seems like a decision that has to be made by the student. Of course, it's also important for the tests so... I'm not sure.

angelikatyborska: After some thought, I find it a bit weird that there are no actual functions involved in the callbacks 🤔. 
I'm thinking the API should rather be like this:

```
React.add_callback(cells, "always_two", fn value -> send(self(), {:callback1, value}))
:ok = React.set_value(cells, "input", 2)
refute_received {:callback, _}, "Expected callback not to be called" # maybe the failure message can be improved
```

jiegillet: Ah nice! 
I actually don't have a good intuition of what callbacks are, so this is very helpful. This feels closer to what it should be!

jiegillet: I had a go at something that I found sensible, please check :)

neenjaw (commented):
I think this is supposed to be a difficult problem, and I'm okay with it being difficult.  I think though, we are perhaps prescribing a very specific approach by defining the structs and using them in the test file.

For me, I'd like to see the interface be a little more agnostic to the solution's data model. A more generic input and allow the student to flex some creativity in their own solution. (To be clear, there is certainly nothing wrong with the approach @jiegillet has taken with their solution to use structs for this)

E.g.

```elixir
# input format
{:input, "input", 1}
# output format
{:output, "output", ["input"], fn a -> a + 1 end}

# Then the interface for the test file becomes:
test "compute cells calculate initial value" do
  {:ok, cells} =
    React.new([
      {:input, "input", 1},
      {:output, "output", ["input"], fn a -> a + 1 end}
    ])

  assert React.get_value(cells, "output") == 2
end
```
The format could be implemented as a provided type.

Interested in your thoughts

jiegillet:
That makes sense. I use Haskell and Elm a lot, so I'm very type-driven and that's how I structure code, but I should let students make their own design choices. The change you propose is pretty straightforward too, so no problem.

exercises/practice/react/lib/react.ex
```diff
@@ -0,0 +1,41 @@
...
+  Add a callback to an output cell
+  """
+  @spec add_callback(
+          cells :: pid,
+          cell :: String.t(),
+          callback :: String.t(),
+          send :: (String.t(), any -> :ok)
+        ) :: :ok
+  def add_callback(cells, cell, callback, send) do
+  end
```
angelikatyborska, line 33: I think we could improve argument names in the whole stub file E.g. I'm not entirely sure about the name "send". It's specific to message sending, but the callback can do anything. In practice it only makes sense if it does side-effect (because there's nobody that cares about the return value) but the side effects could be not only message sending, but also doing network requests or something 🤷 

I would propose:
`cell` -> `cell_name` (because this is the string, not the map `%{type: :input, value: 1}` and definitely not a single instance of the first argument, `cells`)
`callback` -> `callback_name` (because "callback" usually means it's a function)
`send` -> `callback` or `function` or `callback_fun`?

exercises/practice/react/lib/react.ex
```diff
@@ -0,0 +1,41 @@
...
+
+  @doc """
+  Add a callback to an output cell
+  """
+  @spec add_callback(
+          cells :: pid,
+          cell :: String.t(),
+          callback :: String.t(),
+          send :: (String.t(), any -> :ok)
```
angelikatyborska, line 30: Does it really matter what the callback returns?

exercises/practice/react/lib/react.ex
angelikatyborska: Looks much better 👍

exercises/practice/react/lib/react.ex
```diff
@@ -0,0 +1,41 @@
+defmodule React do
+  @doc """
+  Start a reactive system
+  """
+  @spec new(cells :: [%{atom => any}]) :: {:ok, pid}
```
angelikatyborska, line 5: The map `%{atom => any}` will be a very specific map, right? One of:
```elixir
%{type: :input, name: String.t(), value: any()}
%{type: :output, name: String.t(), inputs: [String.t()], compute: (...) -> any()}
```
Should we write that in the typespec instead of the generic `%{atom => any}`?

exercises/practice/react/test/react_test.exs
```diff
@@ -0,0 +1,264 @@
...
+    React.set_value(cells, "input", 20)
+    assert React.get_value(cells, "input") == 20
+  end
+
+  @tag :pending
+  test "compute cells calculate initial value" do
+    {:ok, cells} =
+      React.new([
+        %{type: :input, name: "input", value: 1},
+        %{type: :output, name: "output", inputs: ["input"], compute: fn a -> a + 1 end}
```
neenjaw, line 26: I understand why you are using a map in your data model, to enforce a single entry per key, but I think when sending messages it is much more common to use a tuple for this.

```suggestion
        {:input, "input", 1},
        {:output, "output", ["input"], fn a -> a + 1 end}
```

neenjaw (commented):
I think this is a good iteration, but I still think a tuple is preferable to a map/struct for the arguments

exercises/practice/react/test/react_test.exs
jiegillet: Actually, this is not why I use a tuple, here I use a map because I want to have named fields, otherwise it's unnecessarily complicated for the student to figure out which value is what... This is just following the data from `canonical-data.json`. 

If there was only two fields, it would be ok, but if I have sometimes 2, sometimes 3, I would always go for clarity. 
If you insist I will change to tuples, though.

exercises/practice/react/lib/react.ex
jiegillet: Well the callback must send a message, which is of type `:ok`, otherwise the tests can't be passed.

exercises/practice/react/test/react_test.exs
jiegillet: Well, maybe with the `@spec` specifying which is what it would be OK.

angelikatyborska: I would rather we keep the maps precisely to have named fields. Tuples will be more confusing IMO, because there will be two different ones of different lengths... 

> but I think when sending messages 

This is not sending messages 🤔

exercises/practice/react/lib/react.ex
angelikatyborska: Must send a message? The callback we use in tests sends a message so that we have an easy way of verifying that it was called, but in general the callback can do anything. It's even in your implementation:

```
Enum.each(cell.callbacks, fn {name, send} -> send.(name, value) end)
```

`Enum.each` -> return value gets discarded, only side effects matter

exercises/practice/react/test/react_test.exs
neenjaw: I mean "send a message" in a more philospohical sense ([archived article](https://web.archive.org/web/20170520194237/https://www.integralist.co.uk/posts/message-passing.html) since my link is dead).  Calling a function is "sending a message". Eg. when invoking the function `new` I am sending a message to create a system with these constraints.

neenjaw: Tuples for this sort of thing is still super common, and we see tuples of varying lengths for various meanings all the time: `{:ok, value1, value2}` vs `{:error, "message"}`

neenjaw: But that's why I suggested having a defined type for the tuple so that you can give meaning to the tuple

exercises/practice/react/lib/react.ex
jiegillet: Right, I get it. 

I think what I was doing here is matching the type of the actually callbacks , which are all predefined in the tests as `send_callback` and are indeed all `send`, but I get what you are saying, in a real API, you wouldn't restrict the callback type. I will change it :)

exercises/practice/react/test/react_test.exs
jiegillet: How about we compromise: 
```elixir
%{type: :input, name: "input", value: 1},
{:output, "output", ["input"], fn a -> a + 1 end}
```
(🤣 I'm very sorry 😆)

I will reflect on it for a bit. Thank you both for the detailed review.

angelikatyborska: Hmm, I see. Ok, I won't insist on maps, if @neenjaw is sure tuples will be better in this context, they will work too if we write a nice spec.

angelikatyborska:
I think it's good to merge from my side 🎉 let's see what @neenjaw says

exercises/practice/react/lib/react.ex
```diff
@@ -0,0 +1,42 @@
+defmodule React do
+  @doc """
+  Start a reactive system
+  """
+  @spec new(cells :: [{:input, String.t(), any} | {:output, String.t(), [String.t()], fun()}]) ::
+          {:ok, pid}
+  def new(cells) do
+  end
```
neenjaw, line 8: You could create a type here to break down this complex type
```suggestion
  @type input :: {:input, name :: String.t(), value :: any}
  @type output ::
          {:output, name :: String.t(), dependencies :: list(String.t()), callback :: function()}
  @type cell_definition :: input() | output()

  @opaque cells :: pid

  @doc """
  Start a reactive system
  """
  @spec new(cell_definitions :: [cell_definition()]) :: {:ok, cells()}
  def new(cell_definitions) do
  end
```

neenjaw (approved):
Looks great! 🎉

Issue #819, opened 2021-07-06
Concept exercise stubs raise errors but errors are taught later on 


angelikatyborska:
Concept exercise stubs raise errors but errors are taught later on. Students might get confused about the `raise` keyword and how can you raise a string :) but that's not the point of the exercises they're solving.

A partial example of a current solution stub for `recursion` (comes before `errors`):
```elixir
defmodule BirdCount do
  def today(list) do
    raise "Please implement the today/1 function"
  end

  def increment_day_count(list) do
    raise "Please implement the increment_day_count/1 function"
  end
end
```

What should we do instead?

- comments?
    ```elixir
    defmodule BirdCount do
      def today(list) do
        # Please implement the today/1 function
      end

      def increment_day_count(list) do
        # Please implement the increment_day_count/1 function
      end
    end
    ```
- nothing?
    ```elixir
    defmodule BirdCount do
      def today(list) do
      end

      def increment_day_count(list) do
      end
    end
    ```

jiegillet:
My vote goes to "nothing". When solving exercises it always feels like extra work to me to delete the comments. Of course I know I'm supposed to implement the function!

angelikatyborska:
Maybe I misunderstood Jeremy's worry and it was only about the `hello-world` exercise because Ruby stubs raise a `NotImplemented` error and he wants it like that 🤔 
https://github.com/exercism/ruby/blob/main/exercises/concept/lasagna/lasagna.rb#L2-L4

Python stubs are empty functions:
https://github.com/exercism/python/blob/main/exercises/concept/log-levels/enums.py#L8-L9

JavaScript stubs raise errors:
https://github.com/exercism/javascript/blob/main/exercises/concept/freelancer-rates/freelancer-rates.js#L29

C# stubs raise errors:
https://github.com/exercism/csharp/blob/main/exercises/concept/elons-toys/ElonsToys.cs#L7

I think the question is what is clearer to the student
1. raising an error
   - a bit harsh
   - maybe confusing because errors aren't known yet
   - unidiomatic in Elixir
   - there's no specific `NotImplemented` error like Ruby and C# have
   - but provides a clear message what to do
2. empty function
   - friendlier test failure
   - maybe confusing at first if the student didn't internalize yet that there's always an implicit return
   - but that's idiomatic
   - no clear instruction what to do (if we don't include comments)
       - Jeremy says that if we have empty functions, we should add comments "please implement..." but I totally agree that it's annoying having to remove them everywhere 😅 

[...]

PR #823, opened 2021-07-07
Use comments instead of RuntimeErrors in conccept exercise stubs

angelikatyborska:
Closes https://github.com/exercism/elixir/issues/819

If nobody has any strong feelings against, I think I prefer empty functions and comments with instructions instead of RuntimeErrors because they're more idiomatic.

I promise I'll work on more important things soon :)


neenjaw (approved):
No strong feelings, 👍🏻

PR #824, opened 2021-07-07
🤖 Simplify `hello-world` stub

ErikSchierboom:
Simplify the `hello-world` exercise's stub. The goal of this is to make things easier for students to get started and to introduce as little syntax as possible.

See https://github.com/exercism/v3-launch/issues/46

### DIFF
--- a/exercises/practice/hello-world/lib/hello_world.ex
+++ b/exercises/practice/hello-world/lib/hello_world.ex
@@ -2,6 +2,7 @@ defmodule HelloWorld do
   @doc """
   Simply returns "Hello, World!"
   """
+  @spec hello :: String.t()
   def hello do
     "Goodbye, Mars!"
   end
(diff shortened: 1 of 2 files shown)


exercises/practice/hello-world/lib/hello_world.ex
```diff
@@ -2,8 +2,7 @@ defmodule HelloWorld do
   @doc """
   Simply returns "Hello, World!"
   """
-  @spec hello :: String.t()
```
ErikSchierboom, line 5: I used the example implementation as the basis for this change and that did no include the `@spec` part. I'm trying to make this file as simple as possible, so I've omitted it. Is that okay?

angelikatyborska: No, we should add it to the example implementation too. All practice exercises stubs have `@spec`s and `@doc`s.

Unfortunately our example implementations get out of date with the stub with regard to documentation. The example says:
```
  Greets the user by name, or by saying "Hello, World!"
  if no name is given
```

Which I guess used to be the case a while ago.

ErikSchierboom: I've updated

angelikatyborska (approved):

angelikatyborska:
Thanks!

Issue #821, opened 2021-07-07
Reorder practice exercises


angelikatyborska:
The order of exercises in `config.json` matters. It's used to display them on the website.

If you're completely new to the track, you will first see all concept exercises in their order in `config.json`, and then all practice exercises. If you started working on the track already, this ordering still applies (type, difficulty), but within the same status (in progress, completed, unlocked, locked).

Our current order of exercises is a bit... historical. First, we have former core exercises in their order from v2:
- rna transcription
- word count
- roman numerals
- bob
- beer song
- robot simulator
- list ops
- markdown
- bank account
- zipper
- bowling
- forth
- 
From very easy to very hard. And then you randomly go back to very easy (resistor color, two fer), and then I'm not sure what's the order... probably they're ordered by which core exercise unlocked them in v2. At the very end, we have a bunch of new exercises added after we stopped synchronizing with v2.

I would like to propose that we clean up the order so that all practice exercises are always ordered by their difficulty, and within the same difficulty either alphabetically or by how many prerequisites they have? Or hand-picked order however we want 🙂 ?

Alphabetically would be easier. This would produce the order:

- hello-world (exception to the rune, always first)
- resistor color
- simple linked list
- two fer
- accumulate
- acronym
- all your base
- anagram
- armstrong numbers
- ...
- forth

Upsides:
- easy to find most challenging exercises
- guiding students to the easiest ones first
- better than current random-ish order

Downsides:
- resistor color family would be separated 😢 
- the exact number for difficulty becomes more important, but maybe we didn't choose it so carefully?
- exercise position will change if it ever gets easier or more difficult

If we do this, an automated check on the CI for this ordering rule would be great to have.

neenjaw:
Could write a bash script to just sort the config by name and difficulty using `jq`: https://phpfog.com/sorting-json-by-value-with-jq/

angelikatyborska:
Yeah, it would be easy to automate the sorting. The most difficult part would probably be getting it nicely formatted in the end.

PR #827, opened 2021-07-08
Reorder practice exercises

angelikatyborska:
Approach:
- For the automated check that exercises are ordered, I'm using Elixir because writing Elixir scripts is much easier for me than bash, also probably for every other maintainer.
- For the automated reordering, I'm using `jq` because Elixir JSON parsers all parse objects into maps, losing key order information, and thus producing very bad git diffs when encoding the file again.
- To ensure that hello-world is always first, I added a custom `order` field to that exercise only. It will be used before `slug`

Updated questionable difficulties:
- simple-linked-list: 1 -> 4
- markdown: 5 -> 6
- bank-account: 7 -> 5

Unrelated change: mix format on `exs` files in `bin`

Closes https://github.com/exercism/elixir/issues/821

### DIFF
--- a/bin/check_practice_exercise_order.exs
+++ b/bin/check_practice_exercise_order.exs
@@ -9,13 +9,15 @@ Mix.install([
     aliases: [w: :write]
   )
 
+max_custom_order = 999
+
 config =
   "./config.json"
   |> File.read!()
   |> Jason.decode!()
 
 sort_by = fn exercise ->
-  {exercise["difficulty"], exercise["name"]}
+  {exercise["difficulty"], exercise["order"] || max_custom_order, exercise["slug"]}
 end
 
 reordered_config =
@@ -36,13 +38,31 @@ actual_exercise_order =
   |> Enum.map(& &1["slug"])
 
 if opts[:write] do
+  # set order to a value if missing because otherwise it's null and null comes before 1
+  {file_with_order_field, 0} =
+    System.cmd("jq", [
+      ".exercises.practice[] |= if .order==null then . + {order: #{max_custom_order + 1}} else . + {order: .order} end",
+      "config.json"
+    ])
+
+  File.write!("config.json", file_with_order_field)
+
+  # reorder exercises based on difficulty, custom order overwrites, and slug
   {reordered_file, 0} =
-    System.cmd("jq", [".exercises.practice|=sort_by(.difficulty, .name)", "config.json"])
+    System.cmd("jq", [".exercises.practice |= sort_by(.difficulty, .order, .slug)", "config.json"])
 
   File.write!("config.json", reordered_file)
+
+  # remove fields added in the first step
+  {reordered_without_order_field_file, 0} =
+    System.cmd("jq", [
+      ".exercises.practice[] |= if .order > #{max_custom_order} then del(.order) else . end",
+      "config.json"
+    ])
+
+  File.write!("config.json", reordered_without_order_field_file)
 else
-  if desired_exercise_order === actual_exercise_order &&
-       Enum.at(desired_exercise_order, 0) == "hello-world" do
+  if desired_exercise_order === actual_exercise_order do
     IO.write("Practice exercises are ordered #{IO.ANSI.green()}✔#{IO.ANSI.reset()}\n")
   else
     IO.write("Practice exercises are not ordered #{IO.ANSI.red()}⨯#{IO.ANSI.reset()}\n")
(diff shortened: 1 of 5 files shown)


config.json
```diff
@@ -16,7 +16,7 @@
     "highlightjs_language": "elixir"
   },
   "test_runner": {
-    "average_run_time": 4.0
+    "average_run_time": 4
```
angelikatyborska, line 19: Fuck! 😂 jq did this, but integers aren't allowed here

bin/check_practice_exercise_order.exs
```diff
@@ -0,0 +1,52 @@
...
+  |> Enum.map(& &1["slug"])
+
+actual_exercise_order =
+  config
+  |> get_in(["exercises", "practice"])
+  |> Enum.map(& &1["slug"])
+
+if opts[:write] do
+  {reordered_file, 0} =
+    System.cmd("jq", [".exercises.practice|=sort_by(.difficulty, .name)", "config.json"])
```
angelikatyborska, line 40: This sorting logic will starting failing us once there is a difficulty-1 exercise that comes alphabetically before hello-world 😞 not sure what to do about that.

config.json
```diff
@@ -588,280 +588,252 @@
...
+          "maps"
         ],
-        "difficulty": 2
+        "difficulty": 1
       },
       {
-        "slug": "word-count",
-        "name": "Word Count",
-        "uuid": "4a24ba2f-ae92-4095-be53-64bc881422ea",
+        "slug": "simple-linked-list",
```
jiegillet, line 605: `simple-linked-list` is difficulty 1??
No way, that exercise is very abstract, a solid 5 in my book, you need to know what a linked list even is. I checked community solutions on the site and a solid proportion of answers are all wrong, using `Tuple.insert_at` or even `List` functions. A difficulty 1 should have 99% correct answers on the first submission.

bin/check_practice_exercise_order.exs
jiegillet: Why does `hello-world` need to come first? 
We Ould add out own optional field `"hardcoded_order" : 1` to something.

config.json
angelikatyborska: Totally agree with you, there's no way it's a 1, which would mean "it's as easy as `two-fer` and `resistor-color`, but easier than `resisitor-color-duo`" 😂. Maybe the original author thought it should be solved with the built in linked lists.

That's my main worry about this ordering, that we got the difficulties wrong. `markdown` is a 5... I think I would give `simple-linked-list` a 4 and `markdown` a 6 or 7?

Another very badly scored exercise is `bank-account`. It has a 7, but it's a very straight-forward use of `Agent` or `GenServer`, should be a 3-4?

bin/check_practice_exercise_order.exs
angelikatyborska: `hello-world` is a special exercise in that it isn't allowed to have any prerequisites and it has the purpose of making the student run the simplest test suite and do the simplest possible code change in a new language. Every other exercise, including concept exercises, are all blocked until you do `hello-world`.

A custom key on that one exercise should work 🤔 currently there are no linter rules disallowing it.

config.json
jiegillet: Yes that seems better. 
I have solved a bunch, but not all, so it's hard to judge them all :)

Can we have stats on how many have submitted solutions for each exercise? That could possibly give us a general idea for difficulties.

angelikatyborska: We could ask Jeremy to give us those stats, but I fear they would be skewed by the current unlocking order and core exercise. E.g. everyone will have solved "rna transcription" because you need it do unlock anything else, but few people might attempt "bowling" because you need to do "markdown" and "zipper" first. If you're in mentored mode, not interdependent mode.

jiegillet:
Yes, it will be skewed data, but it might still be valuable for non-track exercises. Or at least interesting :)

config.json
jiegillet: Of course it will be skewed, but it might be useful for non-track exercises. Or at least interesting.

angelikatyborska: Let's talk about that on Slack 👍

bin/check_practice_exercise_order.exs
angelikatyborska: I managed to come up with some `jq` comments that work with a custom `"order"` field that is optional on an exercise and can be used to override the default alphabetical order. This way we will be able to control order of other exercises too, if that ever matters within the same difficulty group.

neenjaw (approved):
Nice work with jq, this will serve us well for now. :)

PR #833, opened 2021-07-10
Rename lib and test files to match slug

jiegillet:
I started looking in `elixir-analyzer` to see if I could help and I made it halfway through the README when I saw

> This assumes the solution has the file of the proper name and also a test unit by the proper name

I though "how hard could it be to rename the files?" In typical dev fashion, probably 3 or 4 times as much as I thought. 

I know this is a completely unprompted PR, and possibly pointless or worse, but I would feel better knowing that all naming is consistent now :)


angelikatyborska:
@jiegillet I'm sorry, but I don't really want this... file names in Elixir should match the module names defined in them, and I want us to choose module names that make logical sense with the code and aren't forced by the exercise name. E.g. `Queens` instead of `QueenAttack`, `Year` instead of `Leap` etc.

Unless @neenjaw is really happy with this change, I would want to close the PR.

Btw. Elixir is not a good tool to regenerate JSONs that were written and maintained by humans :( key order gets changes.

neenjaw:
I think I would have just fixed the readme. The slug is named after the exercise, but in elixir the filenames (and file location) are generally named after the module name. And I think the module's name should relate directly to the task, which isn't always the name of the exercise.

neenjaw:
> Unless @neenjaw is really happy with this change, I would want to close the PR.

Also move to close, sorry for the work you've already committed to this @jiegillet

jiegillet:
OK, no worries, I knew this could happen. I probably should have asked before hand :)

neenjaw:
💙

PR #836, opened 2021-07-21
Correct example in log-level instructions

martinsvalin:
The log-level instructions contain an example where the debug log level is treated as unsupported in legacy apps, but according to instructions (and tests), it is supported. Change the example to use the trace log level, which is unsupported in legacy.


angelikatyborska:
You're right! Thank you 🙂

angelikatyborska (approved):

PR #837, opened 2021-07-22
commented pending tag typo fix

Cohen-Carlisle:
I noticed this typo while doing the exercise and trying to just run one test when I commented the pending tag back in and it went boom :rofl:


neenjaw (approved):
🎉 nice catch!

Issue #817, opened 2021-07-05
Should we disable the test runner for some practice exercises?


angelikatyborska:
From Slack:
> Jeremy Walker  1:20 PM
> We mentioned on the call that you can disable the test-runner for specific exercises that you know it won't work for (e.g. exercises that take ages to run). The flag is here: https://github.com/exercism/docs/pull/169

> If there is an exercise where many solutions will time out, but the optimal ones will pass in time (e.g. pythag triplets), it is worth mentioning this in the instructions:

> > Note: your solution must complete in under 10s for the tests to pass on the online editor and exercism website.

> Remember, student's won't be able to submit in the editor until the tests pass, so you're effectively blocking students from submitting suboptimal solutions if the test_runner is enabled for these exercises.

Here's the test runtime for example solutions on my machine (3 year old MBP):

| exercise                  | example solution test duration |
| ---                       | ---                            |
| accumulate                | 0.06 seconds                   |
| acronym                   | 0.06 seconds                   |
| affine-cipher             | 0.07 seconds                   |
| all-your-base             | 0.06 seconds                   |
| allergies                 | 0.1 seconds                    |
| alphametics               | 10.5 seconds                   |
| anagram                   | 0.07 seconds                   |
| armstrong-numbers         | 0.06 seconds                   |
| atbash-cipher             | 0.07 seconds                   |
| bank-account              | 0.07 seconds                   |
| beer-song                 | 0.04 seconds                   |
| binary                    | 0.05 seconds                   |
| binary-search             | 0.06 seconds                   |
| binary-search-tree        | 0.1 seconds                    |
| bob                       | 0.08 seconds                   |
| book-store                | 0.06 seconds                   |
| bowling                   | 0.1 seconds                    |
| change                    | 0.07 seconds                   |
| circular-buffer           | 0.09 seconds                   |
| clock                     | 0.1 seconds                    |
| collatz-conjecture        | 0.04 seconds                   |
| complex-numbers           | 0.08 seconds                   |
| connect                   | 0.04 seconds                   |
| crypto-square             | 0.08 seconds                   |
| custom-set                | 0.1 seconds                    |
| darts                     | 0.04 seconds                   |
| diamond                   | 0.03 seconds                   |
| difference-of-squares     | 0.03 seconds                   |
| diffie-hellman            | 0.08 seconds                   |
| dnd-character             | 0.06 seconds                   |
| dominoes                  | 0.07 seconds                   |
[...]

PR #841, opened 2021-07-25
[#817] Add `:slow` tags to exclude tests

jiegillet:
Solves #817.

The "slow" test for `prime-factors` is really silly, there are only two factors and the first one is 5. Of course 5 will get found quickly, even by looking at the number you can tell. A real test would have two large factors. Anyway.

Still WIP, because I would like to add a `instructions.append.md` file to each of these and I want to get the wording right before I add the files in case I have to change it. How about this:

> One or several of the tests of this exercise have been tagged as `:slow`, which means that they will get skipped by the automated test runner, since some solutions may take cause a time out. Download the exercise locally to run all the tests on your machine if you want to test the efficiency of your solution.

Should we add that some optimized solutions exist for some specific ones? `pythagorean-triplet` already has a hint, and for `prime-factors`, the more naive the implementation the more likely it is to pass. The others take a while to pass for the example implementations. Maybe `list-ops` could use a mention of `foldr` to avoid using `reverse` (although I have not benchmarked that).


angelikatyborska:
> Still WIP, because I would like to add a instructions.append.md file to each of these and I want to get the wording right before I add the files in case I have to change it.

Great idea!

About the content of that comment: it sounds good but it needs to make sense both for students that read it on the platform and that already downloaded the exercise. I would also rephrase it to first say why those tests are "slow" and only then how that affects the student. Here's my proposal:

> One or several of the tests of this exercise have been tagged as `:slow`. Those tests might take a long time to finish if your solution is not optimal. For this reason, they will not be run on the platform by the automated test runner. If you are solving this exercise directly on the platform in the web editor, you might want to consider downloading this exercise to your machine instead. This will allow you to run all the tests and check the efficiency of your solution.

> Should we add that some optimized solutions exist for some specific ones?

Only if you are sure of what to add there and that it would be correct. I wouldn't know 🤷 

>  Maybe list-ops could use a mention of foldr to avoid using reverse (although I have not benchmarked that).

`foldr` is one of the functions that the student must implement now after the exercise got updated to match problem specs. No students attempted that yet because the updated exercise wasn't deployed. For this reason, I would hold off from adding any hints to this one for now.

jiegillet:
> Those tests might take a long time to finish if your solution is not optimal.

I stayed away from that wording because it implies that there is an optimal solution. There is one for some, like  `pythagorean-triplet `, but we don't have a solution for `palindrome-products` that finishes in 10 seconds. 

How about

> Those tests involve large input values and may take a long time to finish especially for non-optimal solutions.

angelikatyborska:
> I stayed away from that wording because it implies that there is an optimal solution.

right 🤔 then let's just write:

> Those tests might take a long time to finish.

Without mentioning anything about the possibility of making them pass fast or not. It's probably something they would need to ask the mentor about anyway.

I don't like the phrase "large inputs" because it might suggest that the slowness comes from the input itself taking up a lot of space in memory, which is not true for most of those exercises.

angelikatyborska (approved):
Great! Thank you

angelikatyborska:
(I haven't thought this through yet, just I'm writing down my observation)

We will need to fix the test runner because it reports skipped tests as failed.

I also think the append files could use a heading, but I'm not sure ATM how heading levels work with appends files. I will need to ask or read the website source.

Screen Shot 2021-07-30 at 19 01 29

Screen Shot 2021-07-30 at 19 01 44

angelikatyborska:
I opened PRs for both problems

PR #842, opened 2021-07-27
Prefer Integer.pow over :math.pow |> round

frerich:
This tweaks the hints for the diffie-hellman exercise such that instead of suggesting to use :math.pow |> round, it points to Integer.pow.

Integer.pow has only been introduced in Elixir v1.12 but is convenient since it produces integers and works with very large powers, too.

The thoughts on runtime speed still apply though, Integer.pow is not fast enough to deal with the large powers used in the test cases.

### DIFF
--- a/exercises/practice/diffie-hellman/.docs/hints.md
+++ b/exercises/practice/diffie-hellman/.docs/hints.md
@@ -2,7 +2,7 @@
 
 - For generating random numbers, Erlang's `:rand.uniform` or `Enum.random` are
 good places to start.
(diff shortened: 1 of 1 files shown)


frerich:
Hmm. seeing that the CI builds also test with Elixir 1.11 and older, maybe this PR should be closed after all.

angelikatyborska:
Thank you! I think it makes sense to use the newest features of Elixir in instructions and docs. For students that will use the web editor, they will have access to Elixir 1.12. Yes, the CI runs older versions of Elixir, but that only matters for the example solutions.

angelikatyborska (approved):

PR #843, opened 2021-07-31
Add headings to instructions.append.md

angelikatyborska:
- Each md file should start with an h1. It will be removed by the platform so its content doesn't matter.
- The new sections about slow test blend into the instructions on the platform, so I thought they deserve their own heading @jiegillet 

Screen Shot 2021-07-30 at 19 01 44


jiegillet:
Thank you for that. Looks good to me :)

angelikatyborska:
@exercism/maintainers-admin Please approve, it was already reviewed by @jiegillet

iHiD (approved):

PR #847, opened 2021-08-09
Grouped tests to increase `date-parser` test-runner performance

jiegillet:
I went from 142 tests running in 8s with `test-runner` locally on my machine to 51 tests running in 4.8s. 
It's not that different, but it might be enough.

### DIFF
--- a/exercises/concept/date-parser/test/date_parser_test.exs
+++ b/exercises/concept/date-parser/test/date_parser_test.exs
@@ -76,7 +76,7 @@ defmodule DateParserTest do
       assert("9" =~ Regex.compile!(DateParser.month()))
       assert("10" =~ Regex.compile!(DateParser.month()))
       assert("11" =~ Regex.compile!(DateParser.month()))
-      assert("11" =~ Regex.compile!(DateParser.month()))
+      assert("12" =~ Regex.compile!(DateParser.month()))
     end
 
     @tag task_id: 1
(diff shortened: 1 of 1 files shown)


exercises/concept/date-parser/test/date_parser_test.exs
```diff
@@ -98,48 +64,33 @@ defmodule DateParserTest do
...
+      assert("3" =~ Regex.compile!(DateParser.month()))
+      assert("4" =~ Regex.compile!(DateParser.month()))
+      assert("5" =~ Regex.compile!(DateParser.month()))
+      assert("6" =~ Regex.compile!(DateParser.month()))
+      assert("7" =~ Regex.compile!(DateParser.month()))
+      assert("8" =~ Regex.compile!(DateParser.month()))
+      assert("9" =~ Regex.compile!(DateParser.month()))
+      assert("10" =~ Regex.compile!(DateParser.month()))
+      assert("11" =~ Regex.compile!(DateParser.month()))
+      assert("11" =~ Regex.compile!(DateParser.month()))
```
angelikatyborska, line 79: Error carried over from the previous code 😂 
```suggestion
      assert("12" =~ Regex.compile!(DateParser.month()))
```

angelikatyborska (approved):

angelikatyborska:
Perfect! Thank you. Let's see it in action after it deploys in a few minutes.

exercises/concept/date-parser/test/date_parser_test.exs
jiegillet: Oh I missed that.
I was super tempted to write a loop at some point ^^

PR #848, opened 2021-08-12
Add more test cases for bread-and-potions

angelikatyborska:
Adding an extra test case for each edible item to check that it doesn't affect the other property (mana vs health). If you accidentally overwrite one of those in the protocol implementation, only the last test will fail which might not be that straight-forward to debug.


jiegillet (approved):
Looks good to me. Nice improvement.

PR #855, opened 2021-08-29
Adding test for `newsletter`

jiegillet:
Add a test to check if repeated used of `send_newsletter` overwrites log file, to prevent use of `:append`.

Also adjusts `design.md` according to [this discussion](https://github.com/exercism/elixir-analyzer/pull/129).


angelikatyborska (approved):
Perfect 🎉

PR #857, opened 2021-09-01
City office: accept both named and not named argument specs

michallepicki:


angelikatyborska:
@michallepicki the names are required on purpose, to force the student to practice how to use names arguments in a spec at least once. Can you explain why you would want to remove that requirement?

michallepicki:
@angelikatyborska currently the tests require names not to be there for function arguments. It doesn't require them, it rejects them.

angelikatyborska:
Ah, I see! Sorry, it has been a long day 😅 sounds good, let me review properly.

michallepicki:
@angelikatyborska I bet! But great job, the Elixir track looks to be in great shape compared to others.

I can change it to always require names for the arguments. Not sure if that's better.

angelikatyborska:
This is perfect, thank you 💜 I'll merge as soon as the CI passes.

I don't want to force people to add the name everywhere, especially in places where it doesn't add much, like `address :: address()`, so let's leave it optional. But it's great to have it as a possibility for each step.

angelikatyborska (approved):

angelikatyborska:
Damn, I was so tired I forgot to add you as a contributor to the exercise! I opened a PR with that change: https://github.com/exercism/elixir/pull/864

PR #858, opened 2021-09-01
Freelancer rates: Round discount result before comparing floats

### LINKED ISSUE #834: The 0.300000004 problem in freelancer rates
I've been solving freelancer rates again and came up with a solution that the test suite wasn't ready for.

I wrote
```
((100 - discount) * before_discount) / 100
```
instead of
```
(1 - discount/100) * before_discount
```

The test failure in this case is fascinating. I get a result that _looks like_ it needs to be rounded up but the test case says "test it doesn't round" 😂. I suspect we need a `assert_in_delta` but probably just for this test case, since the others are supposed to ceil/floor/trunc?

Screen Shot 2021-07-13 at 18 15 36

michallepicki:
closes #834

### DIFF
--- a/exercises/concept/freelancer-rates/test/freelancer_rates_test.exs
+++ b/exercises/concept/freelancer-rates/test/freelancer_rates_test.exs
@@ -31,7 +31,9 @@ defmodule FreelancerRatesTest do
 
     @tag task_id: 2
     test "it doesn't round" do
-      assert FreelancerRates.apply_discount(111.11, 13.5) == 96.11015
+      # We round result 5th decimal place to accept results that vary marginally
+      # because of slightly different order of operations and float imprecision
+      assert Float.round(FreelancerRates.apply_discount(111.11, 13.5), 5) == 96.11015
     end
   end
 


angelikatyborska:
Thanks 💜 

This solution to the problem is fine by me but what do you think @jiegillet @neenjaw? Would you still prefer to switch the whole exercise to integers like the GitHub issue says? I think I would not, even though it's more correct for money, but it's less... educational 🙂

jiegillet:
Using integers remains my personal preference, but I don't mind keeping floats if you think it's more educational :)
Also, this fixes an issue we have right now, so I'm in favor of merging.

exercises/concept/freelancer-rates/test/freelancer_rates_test.exs
```diff
@@ -31,7 +31,9 @@ defmodule FreelancerRatesTest do
 
     @tag task_id: 2
     test "it doesn't round" do
-      assert FreelancerRates.apply_discount(111.11, 13.5) == 96.11015
+      # We round result 5th decimal place to accept results that vary marginally
```
michallepicki, line 34:
```suggestion
      # We round result to 5th decimal place to accept results that vary marginally
```

exercises/concept/freelancer-rates/test/freelancer_rates_test.exs
```diff
@@ -31,7 +31,9 @@ defmodule FreelancerRatesTest do
 
     @tag task_id: 2
     test "it doesn't round" do
-      assert FreelancerRates.apply_discount(111.11, 13.5) == 96.11015
+      # We round result to 5th decimal place to accept results that vary marginally
+      # because of slightly different order of operations and float imprecision
+      assert Float.round(FreelancerRates.apply_discount(111.11, 13.5), 5) == 96.11015
```
jiegillet, line 36: I think we should use `assert_in_delta` here.

jiegillet: And if we do, maybe we don't need a comment, but I don't really mind either way...

angelikatyborska (approved):

PR #866, opened 2021-09-02
Update darts_test.exs

bpannes:
update to task_ids


angelikatyborska:
Same as https://github.com/exercism/elixir/pull/867

PR #867, opened 2021-09-02
Update year_test.exs

bpannes:
update from :pending to task_ids


angelikatyborska:
@bpannes Thank you, but this is not correct. Leap is a practice exercise. It doesn't have tasks, so it shouldn't have `:task_id` tags. All practice exercises are supposed to have all tests tagged `:pending` except the first one.

PR #869, opened 2021-09-02
Rational numbers: Update pow_real spec to integer to align with tests

### LINKED ISSUE #868: Rational numbers: specs say input is float but is integer in tests
For the `pow_real` function, I wanted to use `Float.pow/2` but it failed in tests because the input is an integer (and that function guards against that), despite the included spec saying it will be float. I worked around this by using `:erlang.float/1` to cast an integer to float (one could also multiply by 1.0).

I didn't send a PR because I see this is one of the exercises ported to Elixir from [problem-specifications](https://github.com/exercism/problem-specifications/tree/main/exercises/rational-numbers). In there the input indeed specifies integers, but description says this function (as name suggests) takes a real number.

Either the spec could be updated to say `number` (which is `integer | float`), and possibly there could be a hint in description about handling both, or maybe Elixir test cases should use floats as input?

michallepicki:
closes #868

### DIFF
--- a/exercises/practice/rational-numbers/lib/rational_numbers.ex
+++ b/exercises/practice/rational-numbers/lib/rational_numbers.ex
@@ -46,7 +46,7 @@ defmodule RationalNumbers do
   @doc """
   Exponentiation of a real number by a rational number
   """
-  @spec pow_real(x :: float, n :: rational) :: float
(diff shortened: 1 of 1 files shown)


angelikatyborska (approved):
Perfect 👏

PR #870, opened 2021-09-02
Update hints.md

ejuten:
Update the function composition URL to point to the computer science function composition as opposed to the mathematical function composition


angelikatyborska (approved):
Thanks!

Issue #686, opened 2021-03-08
Create `introduction.md.tpl` files for concept exercises


angelikatyborska:
According to https://github.com/exercism/docs/blob/main/building/configlet/generating-documents.md#document-concept-exercises-introductionmd-file, it will be possible to use configlet to generate concept exercise `introduction.md` from its concepts' `introduction.md` by creating a `<concept-exercise>/.docs/introduction.md.tpl` file.

All of Elixir concept exercise introductions are exact copies of their concepts' introductions, so we should leverage that template file instead of maintaining two identical copies.

Let's do this once configlet supports this feature.

Note there might be a few exceptions where we do want a slightly different introductions, e.g. for `freelancer-rates` (`integers` and `floating-point-numbers`)

angelikatyborska:
Configlet currently supports templates, but not changing heading levels when merging templates.

jiegillet:
We don't need that feature, do we? Our introductions are identical.

angelikatyborska:
We do. Our exercise introductions have headings one level lower than concept introductions because we add concept names as h2. For example:
```
<!-- bitstrings introduction.md >
# Introduction

Bitstrings are ...

## Pattern matching

Pattern matching on bitstrings ...
```
and
```
<!-- tail call recursion introduction.md >
# Introduction

When recursing ...
```
becomes
```
<!-- file sniffer introduction.md >
# Introduction

## Bitstrings

Bitstrings are ...

### Pattern matching

Pattern matching on bitstrings ...

## Tail Call Recursion

When recursing ...
```

angelikatyborska:
To be clear, we're waiting on: https://github.com/exercism/configlet/issues/328

PR #871, opened 2021-09-04
Fix typos in the HighScore exercise introduction

sevenseacat:
`Example.example_value/0` definitely shouldn't return the value `2`!

Screen Shot 2021-09-04 at 11 32 38 am

### DIFF
--- a/concepts/maps/introduction.md
+++ b/concepts/maps/introduction.md
@@ -23,4 +23,4 @@ While there is no canonical format, choose a consistent way to represent the key
 
 ## Map module functions
 
-Elixir provides many functions for working with maps in the _Map module_. Some _Map module_ functions require an _anonymous_ function_ to be passed into the function to assist with the map transformation.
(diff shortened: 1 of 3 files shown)


jiegillet (approved):
I couldn't agree more. Thank you :)

angelikatyborska:
The same content appears in the introduction files for those concepts. Can you fix that too?

https://github.com/exercism/elixir/blob/main/concepts/module-attributes-as-constants/introduction.md
https://github.com/exercism/elixir/blob/main/concepts/maps/introduction.md

Almost all concept exercise introductions are copies of concept introductions. We're waiting on https://github.com/exercism/elixir/issues/686 to be possible to just use a template instead.

sevenseacat:
will do!

angelikatyborska:
Thank you 💜 

@jiegillet you responded here first, do the merging honors 🎉

PR #873, opened 2021-09-04
Community garden: Clarify expected ids format

michallepicki:
Tests do not accept other IDs than `1, 2, 3` etc, so it should be now clearer that e.g. IDs generated with `System.unique_integer` are not correct.


jiegillet:
I don't think this is correct. I modified the canonical solution to start ids from 100 and go 10 by 10 and I passed all the tests. I even used strings and it worked.
The tests don't even try to register more than one plot, so the iteration is never really tested.

michallepicki:
You're right!

PR #877, opened 2021-09-05
End two-fer test sentences with dot

mpas97:
Be consistent with the instructions.

grafik

### DIFF
--- a/exercises/practice/two-fer/.meta/example.ex
+++ b/exercises/practice/two-fer/.meta/example.ex
@@ -3,5 +3,5 @@ defmodule TwoFer do
   Two-fer or 2-fer is short for two for one. One for you and one for me.
   """
   @spec two_fer(String.t()) :: String.t()
-  def two_fer(name \\ "you") when is_binary(name), do: "One for #{name}, one for me"
+  def two_fer(name \\ "you") when is_binary(name), do: "One for #{name}, one for me."
 end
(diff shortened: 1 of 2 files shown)


jiegillet:
Looks good to me, this is actually aligned with the [problem-specification tests](https://github.com/exercism/problem-specifications/blob/main/exercises/two-fer/canonical-data.json), however, could you also update the example implementation in `.meta/example.ex` so that the test will pass with it?

mpas97:
@jiegillet Sure, already done!

jiegillet (approved):
LGTM, thank you.
I'll let @angelikatyborska have another look and merge, because somehow `two-fer` feels like a legendary and sacred exercise to me 😆

angelikatyborska:
Haha, it is a "legendary" exercise because it's the easiest one, it has been used as the first test case for writing analyzers by a lot of tracks. 

I took a look at the analyzer and it seems like it will still work the same after adding the dot, but we need to bring it up to date with this change too. Dots would need to be added everywhere in the strings in this file: https://github.com/exercism/elixir-analyzer/blob/main/test/elixir_analyzer/test_suite/two_fer_test.exs and in those 4 copies of the exercise https://github.com/exercism/elixir-analyzer/tree/main/test_data/two_fer (note the 3 of them are wrong on purpose because they're test fixtures for the analyzer).

Otherwise I'm fine with this change. The problem specifications say there needs to be a dot.

jiegillet:
And that is exactly why we should import those via a git submodule... 
@mpas97 would you fell up to making the same change in the `elixir-analyzer` repo? The price of meddling with a legendary exercise :p

mpas97:
@jiegillet Nice price 😁 @angelikatyborska Hopefully I got all files exercism/elixir-analyzer#141 ;)

PR #878, opened 2021-09-05
Remove duplicate `@type acc` definition in ListOps

sevenseacat:
This stops the tests from running on a fresh exercise start:

```
== Compilation error in file lib/list_ops.ex ==
** (CompileError) lib/list_ops.ex:29: type acc/0 is already defined in lib/list_ops.ex:24
```


angelikatyborska:
Thank you!

angelikatyborska (approved):

PR #879, opened 2021-09-05
Update appended instructions for ListOps exercise

sevenseacat:
This is to note that the shared documentation for `foldl` uses a function that takes (accumulator, item) whereas Elixir's `reduce` uses a function that takes (item, accumulator).


angelikatyborska:
Thank you 💜

angelikatyborska (approved):

Issue #880, opened 2021-09-05
Atom ambiguity (plural forms) in kitchen-calculator exercise


lauramcastro:
The kitchen-calculator exercise uses a series of atoms to represent baking units, namely (as quoted in the exercise description),  `:cup`, `:fluid_ounce`, `:teaspoon`, `:tablespoon`, and `:milliliter`.

However, the last section of the exercise, describing the `convert/2` function, features an example using a different atom for `:teaspoon`, using the plural `:teaspoons`. This should be fixed to avoid confusing students.

On further exploration, the test file kitchen_calculator_test.exs features also an occurrence of `:teaspoons` in the tests for the `get_volume/1` function, but since this is likely to be implemented ignoring the unit atom, it might have gone unnoticed so far due to this reason. There is also an occurrence of the plural `:tablespoons` in another test for the same `get_volume/1` function.

In summary,

1. The kitchen-calculator exercise features three uses of two plural forms for the unit atoms which are not specified in the exercise description, namely:
    - A `:teaspoons` in the example of use for the `convert/2` function. Should be corrected to `:teaspoon`.
    - A `:teaspoons` in one of the tests for the `get_volume/1` function. Should be corrected to `:teaspoon`.
    - A `:tablespoons` in one of the tests for the `get_volume/1` function. Should be corrected to `:tablespoon`.

I am happy to fork the repo and submit a PR!

angelikatyborska:
@lauramcastro You're totally right. This needs to be corrected to be always singular. A PR from you would be much appreciated, thanks!

jiegillet:
Fixed in #881. Thank you again for the PR.

PR #881, opened 2021-09-05
Fix ambiguity (plural forms) in kitchen-calculator exercise

lauramcastro:
This addresses issue #880 as proposed: all three instances of plural forms of atoms that had slipped through are removed (replaced by their singular counterparts, as specified in the exercise description).


jiegillet (approved):
Very nice, please accept these 5 `:teaspoon`s of appreciation: 🥄🥄🥄🥄🥄

PR #883, opened 2021-09-06
Fix update_score function name in instructions.md

Tom-Wyszkowski:
Incorrect function name and arity "update_player/2" mentioned in instructions. Change to "update_score/3"


jiegillet (approved):
Thank you very much, this is great!

Issue #884, opened 2021-09-06
Make instructions more precise on what they expect


mpas97:
Exercise instructions are often very unclear / unspecific. For `Acronym` this is the whole thing we get:
![grafik](https://user-images.githubusercontent.com/9994218/132216039-c3c17eb5-f1a2-4e5e-8908-3c68a7768fd9.png)

So I would expect that something like this would do it:
`string |> String.split |> Enum.map(&(String.first(&1))) |> Enum.join |> String.upcase`
but instead I always have to do dozens of "try and error" as the tests expect:
- `Metal-Oxide` -> `MO`
- `Something - I` -> `SI`
- `The Road _Not_ Taken` -> `TRNT`
or even things like respecting camel case:
- `HyperText` -> `HT`

So in this exercise, for instance, it would make much sense to explain what characters are expected to be ignored and which should act as spliters. And we should mention that a case change should also trigger the split. We shoudn't have to reverse engineer from the tests to understand an exercise.

verdammelt:
Not an elixir track person but this description is the 'standard' one from the [problem specifications repo](https://github.com/exercism/problem-specifications) which many tracks use as the genesis of these exercises. If that is the case with elixir then we should probably actually move this issue to the problem specifiations repository as the change would need to be made there.

jiegillet:
That's right, we try to follow the specifications closely. If the specifications change, we will follow.

On personal note, I agree with you, I don't like this exercise much either, it's edge cases after edge cases, and all kind of arbitrary. That's what real life problems are for ;)

mpas97:
@verdammelt Thanks! The description is indeed very short there too, but eg. the mixed-case test has already been removed 4 years ago: https://github.com/exercism/problem-specifications/commit/cae7ae184caf9adb08525077d5b2a8e60f844417

jiegillet:
Well, I said we would follow, we should follow then :)
@mpas97, care to do the honors?

angelikatyborska:
> We shoudn't have to reverse engineer from the tests to understand an exercise.

I'm afraid that is exactly what you need to do for almost every practice exercise on Exercism, in every track. The descriptions are almost all very general and shared for all tracks so they cannot include track-specific details, and it's always the tests that specify corner cases. But if it's necessary, we can append track-specific content to the instructions via an extra `instructions.append.md` file.

> but eg. the mixed-case test has already been removed 4 years ago: exercism/problem-specifications@cae7ae1

[...]

PR #885, opened 2021-09-06
Remove additional test case from Acronym

mpas97:
See #884


jiegillet:
I'm slightly leaning towards merging, but not strongly enough to do it unilaterally. I would like to hear the opinion of other maintainers.

angelikatyborska (approved):

angelikatyborska:
I'm for it, and Tim said "shrug emoji" on Slack 😬

PR #889, opened 2021-09-06
Remove prerequisites from deprecated Practice Exercises

ee7:
From [the spec](https://github.com/exercism/docs/blob/14de83e5305b/building/configlet/lint.md#rule-configjson-file-is-valid):

> - The `"exercises.practice[].prerequisites"` value must be an empty
    array if `"exercises.practice[].status"` is equal to `deprecated`

---

This PR resolves an otherwise-upcoming CI failure - `configlet lint` will enforce this rule in the future.

The `mensch-aergere-dich-nicht` concept exercise is also deprecated, but already had an empty `prerequisites` array.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1154,18 +1154,7 @@
         "slug": "binary",
         "name": "Binary",
         "uuid": "29c3c9ff-c7f3-4646-9612-7cd768d237ab",
-        "prerequisites": [
-          "cond",
-          "case",
-          "if",
-          "multiple-clause-functions",
-          "pattern-matching",
-          "binaries",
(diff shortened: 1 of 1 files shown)


angelikatyborska:
Alrighty! But "practices" doesn't need to be empty?

angelikatyborska (approved):

ee7:
> Alrighty! But "practices" doesn't need to be empty?

It does. I just forgot it somehow, even though it's right in the upcoming `configlet lint` output (see https://github.com/exercism/configlet/pull/415 for context).

PR #894, opened 2021-09-08
Fixed example format in German Sysadmin exercise

lauramcastro:
The last example in the German Sysadmin exercise description has a different format from the rest that students find thus far.

IMO, the format of this example would be preferable, since it is slightly confusing to find the `# =>`  prefix before the expected outputs in examples, given that both `#`  and especially `=>` are meaningful symbols in Elixir. Thus, a better alternative (but this might be opinionated, and this opinion based on my lack of knowledge about Exercism restrictions or the like) would be to eliminate said prefix from all other examples in the track. Given that this seems rather extreme, at least this PR makes this example layout coherent with all the rest.

### DIFF
--- a/exercises/concept/german-sysadmin/.docs/instructions.md
+++ b/exercises/concept/german-sysadmin/.docs/instructions.md
@@ -39,5 +39,5 @@ Extend the `sanitize/1` function. It should substitute German characters accordi
 
 ```elixir
 Username.sanitize('cäcilie_weiß')
-'caecilie_weiss'
+# => 'caecilie_weiss'
 ```


angelikatyborska:
We definitely want to keep the `# =>` prefix. The `#` is necessary to make a comment. We need the comment to mark that this code is not part of the example code that you need to execute. The following `=>` is just a convention to differentiate an unstructured comment from a "this is the return value" comment. I am not sure right now where I have seen this convention before. It might come from some Ruby documentation?

I know that the way Elixir docs do it is to mark the executed code with `iex>` and leave the returned value:
```elixir
iex> 1 + 2
3
```

We decided against that convention for two reasons:
- It exposes the student to the concept of interactive Elixir from the very beginning, and we didn't want to force that upon students. If somebody uses the web editor to start learning Elixir on Exercism, they won't even have access to `iex`.
- The syntax coloring wouldn't be correct since `iex>` is not recognized as a special symbol but rather as "some variable greater than...".

> Given that this seems rather extreme, at least this PR makes this example layout coherent with all the rest.

That was a good conclusion 💜 thank you for the fix.

angelikatyborska (approved):

lauramcastro:
Thanks for the extended clarification. I knew there must be some rationale behind it, this documents it perfectly. Very much appreciated!

PR #897, opened 2021-09-08
Fixed name of function in Basketball Website exercise.

lauramcastro:
The Basketball Website uses the name `extract_from/2`, but the code skeleton (and tests) uses `extract_from_path/2`.

The latter is consistent with the `get_in_path/2`, which cannot be named just `get_in/2` because of a collision with the `Kernel` function the student is supposed to use.

Hence, the coherent way of solving this issue is sticking with the `extract_from_path/2` name, and using it in the exercise description. This PR solves the issue in this way.

### DIFF
--- a/exercises/concept/basketball-website/.docs/instructions.md
+++ b/exercises/concept/basketball-website/.docs/instructions.md
@@ -4,7 +4,7 @@ You are working with a web development team to maintain a website for a local ba
 
 ## 1. Extract data from a nested map structure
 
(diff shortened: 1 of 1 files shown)


angelikatyborska (approved):
> Hence, the coherent way of solving this issue is sticking with the extract_from_path/2 name, and using it in the exercise description. This PR solves the issue in this way.

Good conclusion, thank you 💜

Issue #882, opened 2021-09-05
Overwrite `debug.md` for exercises whose tests capture IO


angelikatyborska:
At some point it will be possible to overwrite exercise shared files, see: [docs/building/tracks/shared-files.md#overwriting](https://github.com/exercism/docs/blob/main/building/tracks/shared-files.md#overwriting)

We need to confirm that it's not possible to use `IO.inspect` to debug exercises that capture IO in the tests, and if that's the case, add a custom `debug.md` file for them.

Those exercises are probably only `need-for-speed` and `rpg-character-sheet`.

Suggested `debug.md` content for those exercises but feel free to tweak it:

```
# Debugging

This exercise involves handling IO, and as such, its tests must capture IO. For this reason, there is no way to debug this specific exercise with `IO.inspect`. 
```

angelikatyborska:
Related issue for the platform to actually support this: https://github.com/exercism/exercism/issues/5792

angelikatyborska:
Closed in https://github.com/exercism/elixir/pull/898

PR #898, opened 2021-09-08
Overwrite debug.md for exercises whose tests capture IO

kszambelanczyk:
For issue #882

### DIFF
--- a/exercises/concept/rpn-calculator-output/.docs/debug.md
+++ b/exercises/concept/rpn-calculator-output/.docs/debug.md
@@ -1,0 +1,3 @@
+# Debugging
+
+This exercise involves handling IO, and as such, its tests must capture IO. For this reason, there is no way to debug this specific exercise with `IO.inspect`. 
--- a/exercises/concept/rpn-calculator-output/.docs/debug.md
+++ b/exercises/concept/rpn-calculator-output/.docs/debug.md
@@ -1,3 +1,0 @@
-# Debugging
-
-This exercise involves handling IO, and as such, its tests must capture IO. For this reason, there is no way to debug this specific exercise with `IO.inspect`.
--- a/exercises/concept/rpn-calculator-output/.docs/debug.md
+++ b/exercises/concept/rpn-calculator-output/.docs/debug.md
@@ -1,3 +1,3 @@
 # Debugging
 
-This exercise involves handling IO, and as such, its tests must capture IO. For this reason, there is no way to debug this specific exercise with `IO.inspect`. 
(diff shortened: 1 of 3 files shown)


exercises/concept/rpn-calculator-output/.docs/debug.md
```diff
@@ -0,0 +1,3 @@
+# Debugging
+
+This exercise involves handling IO, and as such, its tests must capture IO. For this reason, there is no way to debug this specific exercise with `IO.inspect`.
```
angelikatyborska, line 3: This one surprised me! I had to do a double-check but that's correct. There is a test that captures IO in this exercise. But I wonder if that's really necessary? The exercise deals with writing to _any_ IO device, not necessarily stdio. Could we rewrite the test to use a file instead and assert the content of the file instead of capturing stdio? (cc @jiegillet @neenjaw)

If we chose to change the test, that would be in another PR, and the only way this PR would be affected would be not adding this custom file. Let's see what people think.

jiegillet: It's easy to do indeed, here it is:

```diff
     send(self(), {:open, filename})
 
     case filename do
-      "filename" -> {:ok, :stdio}
+      "filename" -> File.open(filename, [:write])
       "bad_filename" -> {:ok, spawn(fn -> nil end)}
     end
   end
 
-  def close(_) do
+  def close(file) do
     send(self(), :close)
-    :ok
+    File.close(file)
   end
 
   describe "write/3" do
@@ -50,10 +50,9 @@ defmodule RPNCalculator.OutputTest do
       resource = __MODULE__
       filename = "filename"
       equation = "1 1 +"
+      {:ok, ^equation} = RPNCalculator.Output.write(resource, filename, equation)
 
-      assert capture_io(fn -> RPNCalculator.Output.write(resource, filename, equation) end) ==
-               "1 1 +",
-             @use_write_error_message
+      assert File.read!(filename) == "1 1 +", @use_write_error_message
     end

```

jiegillet: I committed the change [in this branch](https://github.com/jiegillet/elixir/tree/jie-rpn-calculator-output-file), ready to PR as needed. Or we could also include it here?

angelikatyborska: @jiegillet let's open a separate PR

@kszambelanczyk given this change, can you remove the overwritten debug.md for the rpn-calculator-output exercise? Then this PR will be ready to merge 🙂

angelikatyborska (approved):

angelikatyborska:
Thanks 👍

PR #899, opened 2021-09-09
Changing `rpn_calculator` test to use write on File instead of `:stdio`

jiegillet:


exercises/concept/rpn-calculator-output/test/rpn_calculator/output_test.exs
```diff
@@ -50,10 +50,9 @@ defmodule RPNCalculator.OutputTest do
       resource = __MODULE__
       filename = "filename"
       equation = "1 1 +"
+      {:ok, ^equation} = RPNCalculator.Output.write(resource, filename, equation)
 
-      assert capture_io(fn -> RPNCalculator.Output.write(resource, filename, equation) end) ==
-               "1 1 +",
-             @use_write_error_message
+      assert File.read!(filename) == "1 1 +", @use_write_error_message
```
angelikatyborska, line 55: Could you add a cleanup after this step? Right now, running the test suite leaves you with a file called `filename` in the exercise's root directory.

jiegillet: Actually I was wondering, what happens when you create a file in this way? Does it remain on the website with the other files?

angelikatyborska: No, not at all. The website will only ever display files that are defined in the `config.json` of the exercise or files that the student explicitly submitted. But now that I think about it... I wonder if this is allowed by the test runners. The test runners runs in a Docker container with a write-only file system and you can only write to predefined locations that are mounted as volumes. I think the whole exercise directory is mounted as a volume, so it should be safe to write there like this. Alternatively we could use a temp dir (https://til.hashrocket.com/posts/8rkl48vqmc-get-a-writable-temporary-directory-in-elixir) because `/tmp` is definitely writable in the test runner container.

Anyway, whatever we do, this exercise is not yet live because it's waiting for a better `dynamic-dispatch` exercise so we won't feel the consequences immediately 😁

jiegillet: I see, we may need to revisit it then.

jiegillet:
Darn, I used `after_suite` that was only added in 1.8...

angelikatyborska:
Try `on_exit`: https://hexdocs.pm/ex_unit/ExUnit.Callbacks.html#on_exit/2

angelikatyborska (approved):
Perfect!

Issue #900, opened 2021-09-09
Simple Linked List - Test at line 100


czrpb:
Hi!

I believe this test is slightly wrong:

https://github.com/exercism/elixir/blob/67824de8209ff1b6ed2f736deedfb5bd815130ca/exercises/practice/simple-linked-list/test/linked_list_test.exs#L100

I believe the expected `pop`s are in the wrong order. You can also reference other tests where the expectations are LIFO; such as tests on lines: 53 & 81.

I have the code locally and can do a PR, but it will be my first and thus want to adhere to community norms/mores.

Thoughts? Suggestions?

Thanks!!

  << q

czrpb:
Hi again!

I have forked and can do a PR if appropriate and whenever! :)

thx!! << q

jiegillet:
Hi @czrpb,
Thank you for raising this issue.
Yes, please, open a PR and we will gladly review it!

czrpb:
Hi!

Ok, I will:

1 Close the current PR (which is changing the test)
2 Open a PR to change the documentation for practice exercise

Hope I am not wasting your time!??

thx!! << q

jiegillet:
No time wasted whatsoever, thank you for your help!
But there is no need change the documentation, just the name of the test will be enough.

czrpb:
Hi!

New PR #908 that:
* Updates 2 test names reminding that order should be kept/preserved
* 1 new test that verifies that `from_list/1` is successive `push/2` but reversed

Thx!

jiegillet:
Closed via #908.

PR #901, opened 2021-09-10
change order of expectation/pops

czrpb:
The linked list expectation is LIFO and thus this test has it backwards, other tests do not.

See issue #900 for a bit more detail.

Thanks!

### DIFF
--- a/exercises/practice/simple-linked-list/test/linked_list_test.exs
+++ b/exercises/practice/simple-linked-list/test/linked_list_test.exs
@@ -100,8 +100,8 @@ defmodule LinkedListTest do
   test "from_list/1 of 2 element list" do
     list = LinkedList.from_list([:a, :b])
     assert LinkedList.length(list) == 2
-    assert {:ok, :a, list} = LinkedList.pop(list)
     assert {:ok, :b, list} = LinkedList.pop(list)
+    assert {:ok, :a, list} = LinkedList.pop(list)
     assert {:error, :empty_list} = LinkedList.pop(list)
   end
 


czrpb:
re: label for reputation, not sure how to do that, but this would be x:size/tiny.

jiegillet:
Sorry, I should have caught that in the issue that you opened, but I don't think this needs to change. 
Basically this test checks in which order you should transfer a list.

Lists in Elixir are also linked list, so we have
```elixir
[:a, :b, :c] == [:a | [ :b | [ :c | []]]]
```
and `:a` is the head of the list. We want to preserve this property in the `LinkedList` too, so `[:a, :b, :c]` would become (depending on your implementation) something like
```elixir
{:a, {:b, {:c, {}}}}
```
where `:a` is also the head and gets popped first. In other words, we want the equivalent `hd` (for elixir lists) and `pop` (for `LinkedList`) to return the same things in the sane order if the lists are equivalent.

I do see your point though, if elements are given one by one to be pushed (first `:a`, then `:b`), then yes, `:b` would be popped first. But I think creating lists that have equivalent properties is a better interpretation for `from_list` and `to_list`.

What do you think?

jiegillet:
Maybe it would be nice to update the description of the test instead, maybe something like "from_list/1 of 2 element list preserves the positions of the list"?

czrpb:
Hi!

I think I get what you are saying. I will admit that I think understanding the differences between this test and say the ones at lines 67 & 81 is pretty subtle.

So, that said, yep, I do think a bit of change to the test's name will help.

I shall close this then yes?

thx!! << q

jiegillet:
In 67 and 81, the test is about is pushing numbers one at a time and popping them. This test is about transforming an Elixir list into a LinkedList. I agree that the difference is subtle :)

You could close, or you could change your PR to update the test with a better description, so you can help people with the same understanding as yours, that would be really appreciated!

czrpb:
Closing and hope to open a doc change soon/today!

PR #908, opened 2021-09-11
update test names & new test showing order semantic differences between from_list/1 and push/2

czrpb:
I initially wrote `from_list/1` as successive `push/2` but that failed the `from_list/1` & `to_list/1` tests.

This is not the intended semantics of `from_list/1`.

This PR updates 2 test names and adds a new test making this (hopefully!) clearer and at least explicit.

### DIFF
--- a/exercises/practice/simple-linked-list/test/linked_list_test.exs
+++ b/exercises/practice/simple-linked-list/test/linked_list_test.exs
@@ -130,7 +130,8 @@ defmodule LinkedListTest do
     ll_via_from_list = LinkedList.from_list(list)
     ll_via_pushed = Enum.reduce(list, LinkedList.new(), &LinkedList.push(&2, &1))
 
-    assert LinkedList.to_list(ll_via_from_list) == LinkedList.to_list(ll_via_pushed) |> Enum.reverse
+    assert LinkedList.to_list(ll_via_from_list) ==
+             LinkedList.to_list(ll_via_pushed) |> Enum.reverse()
   end
 
   @tag :pending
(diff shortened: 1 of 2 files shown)


SleeplessByte:
For the elixir maintainer that checks this: also check problem-spec to see if the same naming issue lives there. If so, it should be updated there too.

jiegillet:
@SleeplessByte Part of the problem here is that there are no tests for `simple-linked-list` in problem-spec. Currently the only exercises not implemented in Elixir (except for some foregone ones) are ones that don't have tests, precisely for this kind of issues I guess :)

exercises/practice/simple-linked-list/test/linked_list_test.exs
```diff
@@ -118,11 +118,22 @@ defmodule LinkedListTest do
...
     list = LinkedList.from_list([:mon, :tues])
     assert LinkedList.to_list(list) == [:mon, :tues]
   end
 
+  @tag :pending
+  test "from_list/1 and successive push/2 of a list result in reversed order" do
+    list = [:mon, :tues]
+
+    ll_via_from_list = LinkedList.from_list(list)
+    ll_via_pushed = Enum.reduce(list, LinkedList.new(), &LinkedList.push(&2, &1))
```
jiegillet, line 131: Variable names with `l` are hard to read because they look like `1`. 
Maybe better names could be `linked_list` and `pushed_list`?

jiegillet (commented):
Yes, I think this new test would address the ambiguity. Thank you.

exercises/practice/simple-linked-list/test/linked_list_test.exs
czrpb: Done! Thx!

jiegillet (approved):
Cool, that works for me. Thank you!

jiegillet:
Oh, one more thing, you added a test, so if you would like to add your name into the list of contributors in`.meta/config.json`, go ahead.

PR #909, opened 2021-09-12
rpg-character-sheet: improve testing on IO.inspect's label

Reunina:
the label message was not taken int account. ( see screenshot below )
 

Capture d’écran 2021-09-11 à 19 58 45 (1)
Capture d’écran 2021-09-11 à 19 58 23 (1)
Capture d’écran 2021-09-11 à 19 57 56 (1)


jiegillet (approved):
Nice catch, `inspect` doesn't actually have a `:label` option, but no error is raised if you use it anyway.

Thank you for the contribution!

angelikatyborska (approved):

PR #911, opened 2021-09-12
Remove Agent memoization from `book-store`

jiegillet:
Indeterminate behavior of the Agent memoization could (rarely) fail the test suite and make the whole CI fail.
The memoization isn't necessary for the size of the test values, so it's better to get rid of it.


angelikatyborska (approved):
Would explicitly calling `Agent.stop` in `total` also solve this issue?

jiegillet:
I'm not sure. I actually haven't been able to reproduce the failure, much less debug it.

angelikatyborska:
🤷 then feel free to merge

Issue #910, opened 2021-09-12
Extend `remote-control-car` tests


angelikatyborska:
There are those existing checks for functions that accept the struct:
```elixir
  test "display distance raises error when not given struct" do
    fake_car = %{
      battery_percentage: 100,
      distance_driven_in_meters: 0,
      nickname: "Fake"
    }

    assert_raise(FunctionClauseError, fn ->
      RemoteControlCar.display_distance(fake_car)
    end)
  end
```

Those checks guard against plain maps, but not against other structs. It's possible to do a pattern match on `%{__struct__: _}` that will allow *any* struct.

To prevent this, we could add extra tests (`"... raises error when given an unexpected struct"`?) that attempt to pass a locally defined `FakeRemoteControlCar` struct and check that `FunctionClauseError` is raised.

Stqs:
I can take it @angelikatyborska 
Do you allow creating of a new branches in this repo?

angelikatyborska:
@Stqs Thank you for the help 🙏. Unfortunately creating a branch in this repository is a privilege that requires write access, which is reserved for maintainers. The standard procedure in all open source repos on GitHub is that contributors must create their own fork first (like you have done already).

PR #912, opened 2021-09-12
[#910] Add new tests to remote-control-car

Stqs:
Fixes https://github.com/exercism/elixir/issues/910


angelikatyborska:
@Stqs Awesome! This is a very good direction, but I meant for those to be additional tests, not replacements for the current tests with fake map cars. So first there should be a test `"(function) raises error when not given struct"` with a map, and then a second one `"(function) raises error when given struct the wrong struct"` with the local fake car struct.

This will allow student to gradually discover the difference between a map and a struct, and then between different structs.

Stqs:
Updated

angelikatyborska (approved):
Now it's perfect! Thank you a lot for finding this corner case with your solution and by adding more tests 💜

Issue #907, opened 2021-09-11
Date Parser instructions don't match the tests


Eiermann:
Task 1 instructions in Date Parser state in part: "Do not worry about error checking. You can assume you will always be passed a valid numeric component." The tests, however, check that the code does error check the input and don't always pass a valid numeric component.

jiegillet:
I think you are correct. Would you like to open a PR to correct the instructions? That would be greatly appreciated!

Eiermann:
Sure, will do.

Eiermann:
I tried to push a branch and was denied permission. I can connect to github using ssh (`ssh -vT git@github.com` works fine for me and I cloned the repo using ssh). I google'd a bit for a fix but didn't find anything that worked. If you have any tips, I'll give them a shot, if not, someone with correct access will need to resolve this issue.

angelikatyborska:
@Eiermann You need to create your own fork and push there, and then open a cross-fork PR. You don't have write access to this repository.

Eiermann:
Thanks for your patience. I *believe* I've managed to fork (well, I already had a fork from a long time ago and have managed to update it) and submit the cross fork PR. If not, I may not be smart enough for Github (very possible). No worries either way.

angelikatyborska:
This was fixed in https://github.com/exercism/elixir/pull/913

PR #913, opened 2021-09-12
#907 - Date Parser - Removed a line from instructions to better match the tests

Eiermann:
The instructions say it can be assumed that the code will always be passed a valid numeric component. The tests, however, do not always pass a numeric component nor a valid numeric component, so I removed that line from the instructions.


jiegillet (approved):
Thank you, this is very much appreciated.

Issue #905, opened 2021-09-11
Problem with the german sysadmin test cases


spapas:
Hello, I am trying the german sysadmin exercise: https://exercism.org/tracks/elixir/exercises/german-sysadmin/ using the online code editor ... I am using the following code as a starting point:

```
defmodule Username do
  def sanitize(username) do
    Enum.map(username, fn x -> x end)
  end
end
```

It does not run the tests at all and displays a strange error:

> An error occurred while running your tests. This might mean that there was an issue in our infrastructure, or it might mean that you have something in your code that's causing our systems to break. Please check your code, and if nothing seems to be wrong, try running the tests again.

After discussing the problem in the slack channel, I understood that the problem was that the tests was timing out because one test case uses the input 0..0x10FFFF, and since the function doesn't modify them, the test suite tries to print it out to display the error.

I think that this particular test case should be removed or modified somehow so it wouldn't time out. Thank you

angelikatyborska:
Thank you for reporting! I agree this is a problem. I would prefer to keep this test case but I think there are options of modifying it in a way that doesn't make the test runner time out.

Quoting @neenjaw from Slack:

> one test case uses the input 0..0x10FFFF to generate a list of integer codepoints (a charlist), and since your function doesn't modify them, the test suite tries to print it out to display the error.  I think while printing it out on the test-runner process, it times out because printing to std out a list that long is very slow (relatively)

I haven't had the chance to verify this yet, but if it's true, it should be possible to fix the problem by splitting the test into many different test with different ranges (but explicitly, without meta programming, for the sake of the test runner being able to extract the test metadata correctly). If that would produce too many tests, we could only test a part of the ranges, with more focus around the latin character range and the few special characters from the exercise and only check two or three codepoints from other ranges.

Alternatively, we could modify the test to do smarter assertions, something like this (I haven't tested if this works):
```elixir
input = Enum.to_list(0..0x10FFFF) -- [?_, ?ß, ?ä, ?ö, ?ü]
expected = 'abcdefghijklmnopqrstuvwxyz'
actual = Username.sanitize(input)

assert Enum.take(actual, Enum.count(expected)) == expected
assert Enum.count(actual) == Enum.count(expected)
```

spapas:
Hello I just sumbled upon the prime factors exercise that mentions marking tests with `:slow` to avoid running them on the web interface. Maybe also marking the offending test as slow in this exercise would be helpful.

angelikatyborska:
[...]

PR #915, opened 2021-09-13
German sysadmin - fix for long running test

kszambelanczyk:
for issue #905


angelikatyborska:
👍

Locally, a failing solution (that doesn't filter out anything) before the changes:
```
$ time mix test --include pending > /dev/null
mix test --include pending > /dev/null  3.08s user 1.15s system 121% cpu 3.491 total
```

After the changes
```
$ time mix test --include pending > /dev/null
mix test --include pending > /dev/null  1.75s user 0.77s system 131% cpu 1.908 total
```
That cut the runtime in half.

angelikatyborska (approved):

PR #896, opened 2021-09-08
Fixed extra comman in Boutique Inventory exercise section 2 example

lauramcastro:
The example in section 2 of the Boutique Inventory exercise has an extra comma. This is of course harmless, but fixing it is convenient for students to copy & paste while solving the exercise.

This PR fixes said extra comma.

PR #917, opened 2021-09-17
Renamed null to nil in Flatten Array code example

lauramcastro:
Similar to what happened in PR #896 , the code example in the Flatten Array exercise uses `null` instead of `nil`, which prevents convenient copy & pasting.

This PR also fixes a mid-sentence upcasing.

### DIFF
--- a/exercises/practice/flatten-array/.docs/instructions.md
+++ b/exercises/practice/flatten-array/.docs/instructions.md
@@ -4,8 +4,8 @@ Take a nested list and return a single flattened list with all values except nil
 
 The challenge is to write a function that accepts an arbitrarily-deep nested list-like structure and returns a flattened structure without any nil/null values.
 
-For Example
+For example:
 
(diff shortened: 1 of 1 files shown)


angelikatyborska:
Hi! The instructions are copied from https://github.com/exercism/problem-specifications which is a repository that applies to all programming languages on Exercism. That's why the instruction sometimes don't make sense for Elixir.

If you have an idea for how to make those instrutions language-agnostic, it would be great it you opened a PR in https://github.com/exercism/problem-specifications

Here, even if we update them, they will get overwritten the next time we update the exercise. For this reason I cannot accept this change. I'll mention this in CONTRIBUTING.MD.

lauramcastro:
It makes sense, I totally understand! I thought that might be the reason for keeping "null/nil" in the text, but didn't know the whole content was automatically imported from somewhere else.

Thanks, once more, for taking the time to explain and filling in my lacking context <3

PR #921, opened 2021-09-22
Add more tests in `rpn-calculator` concept

cjmaxik:
This PR adds more tests in `rpn-calculator` concept to make the test more in line with the README exposition.

Namely, it says the following, but it is not checked during the exercise:
> When doing more research you notice that many functions _**use atoms and tuples to indicate their success/failure**_.

### DIFF
--- a/exercises/concept/rpn-calculator/.meta/config.json
+++ b/exercises/concept/rpn-calculator/.meta/config.json
@@ -5,7 +5,8 @@
     "neenjaw"
   ],
   "contributors": [
-    "angelikatyborska"
+    "angelikatyborska",
+    "cjmaxik"
   ],
   "files": {
     "solution": [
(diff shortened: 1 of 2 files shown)


angelikatyborska (commented):
Awesome! Thank you.

Two small details:
- Feel free to add your GitHub handle to the `"contributors"` array in `exercises/concept/rpn-calculator/.meta/config.json` 🚀 
- On the platform, concept exercises do not show the full test file. For this reason, all test blocks (`test "..." do end`) need to be self-explanatory and understandable without looking outside of the block. This means we cannot use module attributes to DRY up the tests. Could you remove them?

cjmaxik:
No problem, all done.

angelikatyborska (approved):
Thanks again 🚀

Issue #925, opened 2021-09-23
Error in test case for Bird Count


tradfursten:
Isn't the final test in the bird count exercise wrong.

The function should return the index of the largest element but the test asserts the following:
```
assert BirdCount.busy_days([0, 6, 10, 4, 4, 5, 0]) == 3
```
The third element in the list has index 2.

angelikatyborska:
The function doesn't return indices. It returns how many days have at least 5 birds visiting. In this case 3 - `6`, `10`, and `5`. The test is correct.

tradfursten:
You should read the whole instructions before you say anything :)

PR #926, opened 2021-09-23
Fixed broken test in Bird Count

tradfursten:
Solving issue https://github.com/exercism/elixir/issues/925


angelikatyborska:
Thanks, but I replied on the issue why this change is not correct. I'm going to close the PR.

PR #930, opened 2021-09-24
New concept exercise `top-secret` (concept: `ast`)

angelikatyborska:
@jiegillet I loved your idea so much that I immediately started working on it :) this is just the example solution, I will work on the test suite next. Not necessary to review yet, but I just wanted to signal I'm working on it.


jiegillet:
Glad my idea inspired you so much :)
Maybe I can come up with a couple more test cases ^^

angelikatyborska:
@jiegillet the instructions, tests, stubs, and exemplar should all be "finished". It would be great if you could try to solve the exercise by following the instructions without looking at the exemplar first and let me know if they make sense, maybe even suggest hints based on your own experience 🙂 

I will work on the introduction, hints, and everything else that needs to be finished soon.

exercises/concept/top-secret/.docs/instructions.md
```diff
@@ -0,0 +1,73 @@
+# Instructions
+
+You're part of a task force fighting against corporate espionage. You have a secret informer at Shady Company X, which you suspect of stealing secrets from its competitors.
```
angelikatyborska, line 3: I wanted to avoid any implicit political associations so I decided to mention explicitly that this is about corporate espionage. And to avoid suggesting that the student is supposed to be a criminal here 😁 so I went with _fighting_ espionage. I hope this still makes sense.

exercises/concept/top-secret/test/top_secret_test.exs
```diff
@@ -0,0 +1,235 @@
...
+      string = "def adjust(a, b), do: :scale"
+      ast = TopSecret.to_ast(string)
+      acc = ["re"]
+
+      assert TopSecret.decode_secret_message_part(ast, acc) == {ast, ["ad", "re"]}
+    end
+
+    @tag task_id: 3
+    test "function arity 0 results in empty string" do
+      string = "def adjust(), do: :scale"
```
jiegillet, line 132: If you want to be mean, also have a test with `"def adjust, do: :scale"`

jiegillet (commented):
Awesome exercise. Here is my solution without looking at yours:

```elixir
defmodule TopSecret do
  def to_ast(string) do
    Code.string_to_quoted!(string)
  end 

  @def_ops [:def, :defp]
  def decode_secret_message_part({op, _, [{name, _, args} | _]} = ast, acc)
      when op in @def_ops and is_list(args) do
    acc = add_part(name, length(args), acc)
    {ast, acc}
  end 

  def decode_secret_message_part({op, _, [{:when, _, [{name, _, args} | _]} | _]} = ast, acc)
      when op in @def_ops and is_list(args) do
    acc = add_part(name, length(args), acc)
    {ast, acc}
  end 

  def decode_secret_message_part(ast, acc) do
    {ast, acc}
  end 

  def decode_secret_message(string) do
    {_, message} =
      string
      |> to_ast()
      |> Macro.prewalk([], &decode_secret_message_part/2)

    message
    |> Enum.reverse()
    |> Enum.join()
  end 

  defp add_part(name, arity, acc) do
    name
    |> to_string()
    |> String.slice(0, arity)
    |> then(&[&1 | acc])
  end 
end
```

exercises/concept/top-secret/.meta/exemplar.ex
```diff
@@ -0,0 +1,32 @@
+defmodule TopSecret do
+  def to_ast(string) do
+    Code.string_to_quoted!(string)
+  end
+
+  def decode_secret_message_part({keyword, _, children} = ast, acc)
+      when keyword in [:def, :defp] do
+    [{function_name, _, arguments} | _] = children
+
+    arity = length(arguments)
```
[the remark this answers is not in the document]
angelikatyborska: Great point! I'll add a test for this too

angelikatyborska:
Hmm. Actually I like the idea of including a case for `when` as an extra task. It would increase the amount of actual AST work in the exercise, and together with the zero-arity-no-parentheses case it would drive the point of how non-trivial working with ASTs is when it needs to work with all kinds of inputs 🤔

concepts/ast/about.md
```diff
@@ -0,0 +1,40 @@
+# About
```
angelikatyborska, line 1: I gave up on writing a custom `about.md` and just copy-pasted `introduction.md` because I have doubts about how many students actually read those and I'm low on time.

jiegillet: I think we will need to circle back to it once we have quote/unquote and macros concepts anyway.

concepts/ast/about.md
```diff
@@ -0,0 +1,40 @@
...
+
+## AST
+
+The Abstract Syntax Tree (AST), also called a _quoted expression_, is a way to represent code as data.
```
jiegillet, line 5: Maybe add a link to the Elixir getting started page?

config.json
```diff
@@ -596,6 +596,23 @@
           "dates-and-time"
         ],
         "status": "beta"
+      },
+      {
+        "slug": "top-secret",
```
jiegillet, line 601: We should add the prerequisite to `dot-dsl` too.

exercises/concept/top-secret/.docs/hints.md
```diff
@@ -0,0 +1,52 @@
...
+- The first element on that list is the function's name, and the second element is the function's body.
+- The exact pattern to match an AST node that defines a function and get its name is: `{op, _, [{name, _, _} | _]} when op in [:def, :defp]`.
+
+## 3. Decode the secret message part from function definition
+
+- Inspect the output of [`quote`][doc-quote] to familiarize yourself with how ASTs look like for specific code snippets.
+- The AST node that contains the function's name also contains the function's argument list as the third element.
+- The exact pattern to match an AST node that defines a function and get its name and arguments is: `{op, _, [{name, _, arguments} | _]} when op in [:def, :defp]`.
+- The arity of a function is the length of its argument list.
+- There is a [built-in function in the `String` module][string-slice] that can get the first `n` characters from a string.
```
jiegillet, line 30: I would add one more hint:

```suggestion
- There is a [built-in function in the `String` module][string-slice] that can get the first `n` characters from a string.
- A function without arguments written without parentheses will not have a list as argument but an atom.
```

exercises/concept/top-secret/.docs/hints.md
```diff
@@ -0,0 +1,52 @@
...
+
+- Inspect the output of [`quote`][doc-quote] to familiarize yourself with how ASTs look like for specific code snippets.
+- When a function has a guard, the third element in the tuple for the `:def/:defp` operation is a bit different.
+- That third element is a list with two elements, the first one is the tuple for the `:when` operation, and the second one is the function's body.
+- The `:when` operation's arguments are a two-element list, where the first argument is the function's name, and the second is the guard expression.
+- The exact pattern to match an AST node that defines a function with a guard and get its name and arguments is: `{op, _, [{:when, _, [{name, _, arguments} | _]} | _]} when op in [:def, :defp]`.
+
+## 5. Decode the full secret message
+
+- There is a [built-in function][macro-prewalk] that can visit each node in an AST with an accumulator.
```
jiegillet, line 42: One more:

```suggestion
- There is a [built-in function][macro-prewalk] that can visit each node in an AST with an accumulator.
- Use the function `to_ast/1` that you implemented in the first task to create the AST.
```

jiegillet (approved):
Looks great!
I left a couple of comments for small things.

exercises/concept/top-secret/.docs/hints.md
```diff
@@ -0,0 +1,52 @@
...
+
+## 2. Parse a single AST node
+
+- Inspect the output of [`quote`][doc-quote] to familiarize yourself with how ASTs look like for specific code snippets.
+- The operations that define a function are `:def` and `:defp`.
+- The operation is the first element in a three-element AST node tuple.
+- You can ignore the second element in the tuple in this exercise completely.
+- The third element in the tuple is the argument list of the operation that defines the function.
+- The first element on that list is the function's name, and the second element is the function's body.
+- The exact pattern to match an AST node that defines a function and get its name is: `{op, _, [{name, _, _} | _]} when op in [:def, :defp]`.
```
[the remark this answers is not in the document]
angelikatyborska: Yeah, maybe, I wasn't sure.. I'm a bit worried that some people might not figure it at at all on their own, but you're right. I'll remove it.

The only other exercise where we have a very revealing hint like that is for Captain's Log, for the formatting string for `:iolib.format`, and that's because Erlang documentation for that topic is... written in a way that I don't understand it at all 🙂

angelikatyborska:
I'll merge this in an hour or so 🤞 it feels much more stressful to add new exercises when v3 is already live.

PR #932, opened 2021-09-24
Update HELLO.md

angelikatyborska:
- Fixes outdated links
- Stops talking about v3 and using the future tense, the future is now!
- Removes the task to port an exercise from problem specifications because those have been depleted 😬


jiegillet (approved):
> Removes the task to port an exercise from problem specifications because those have been depleted 😬

Sorry, not sorry ;)

PR #935, opened 2021-09-25
Dont shadow kernel length

Cohen-Carlisle:
At least one student has reported that defining length/1 warns that doing so conflicts with the automatically imported function Kernel.length/1, and it seems generally prudent not to shadow Kernel functions.

### DIFF
--- a/exercises/practice/list-ops/.docs/instructions.append.md
+++ b/exercises/practice/list-ops/.docs/instructions.append.md
@@ -6,6 +6,8 @@ The above description of this exercise is shared between all Exercism tracks.
 
 To be consistent with Elixir's standard library, the functions used by `foldl` and `foldr` should take the item as the first argument, and the accumulator as the second.
 
+Additionally, in order not to conflict with the automatically imported function `Kernel.length/1`, the function counting the length will be named `count` instead.
+
 ## Slow tests
 
 One or several of the tests of this exercise have been tagged as `:slow`, because they might take a long time to finish. For this reason, they will not be run on the platform by the automated test runner. If you are solving this exercise directly on the platform in the web editor, you might want to consider downloading this exercise to your machine instead. This will allow you to run all the tests and check the efficiency of your solution.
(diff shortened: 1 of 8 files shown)


jiegillet (approved):
I think this is an excellent change, thank you!

angelikatyborska (changes requested):
I... disagree. We can't change the instructions. They will be overwritten the next time we autoupdate exercises from problem specifications. Couldn't this also be solved by adding `import Kernel, except: [length: 1]` to the solution stub?

jiegillet:
Oh, my bad! Somehow I though this was a concept exercise. Sorry!

This could be solved by adding `import Kernel, except: [length: 1]`, but I feel that the name change is a cleaner solution.

Cohen-Carlisle:
Small update: This only causes a compilation error if you invoke the local function in your module. For example, I can define:
```
iex(1)> defmodule A do
...(1)> def length(_), do: 1
...(1)> end
{:module, A,
 <<70, 79, 82, 49, 0, 0, 4, 180, 66, 69, 65, 77, 65, 116, 85, 56, 0, 0, 0, 134,
   0, 0, 0, 14, 8, 69, 108, 105, 120, 105, 114, 46, 65, 8, 95, 95, 105, 110,
   102, 111, 95, 95, 10, 97, 116, 116, 114, ...>>, {:length, 1}}
```
just fine. But If I do this:
```
iex(2)> defmodule B do
...(2)> def length(_), do: 1
...(2)> def test(b), do: length(b)
...(2)> end
** (CompileError) iex:3: imported Kernel.length/1 conflicts with local function
    (elixir 1.12.2) src/elixir_locals.erl:94: :elixir_locals."-ensure_no_import_conflict/3-lc$^0/1-0-"/2
    (elixir 1.12.2) src/elixir_locals.erl:95: anonymous fn/3 in :elixir_locals.ensure_no_import_conflict/3 
```
then I get the CompileError.

angelikatyborska (approved):

PR #936, opened 2021-09-28
fix test description

volkerholloh:
Hi there,

while working through the awesome examples I ran into a test having a not fitting description. Thanks a ton for all the work you are doing on Exercism.

Kind regards,
Volker


angelikatyborska (approved):
Ah, that's a classic copy-paste mistake from an earlier test :). Thank you!

Issue #937, opened 2021-09-29
Instructions in list-ops doesn't match the expected function signature


kahgoh:
While working on the list-ops exercise, I noticed the some of the methods in the instructions doesn't match what is expected by the tests. For example:
- The instructions list a `concatenate` function, but the tests expect `concat`.
- The instructions lists the `list` as the second argument to the functions, but the tests expects it as the first.

jiegillet:
Well the instructions also say before the list of functions "The precise number and names of the operations to be implemented will be track dependent to avoid conflicts with existing names" and that is exactly what is happening here.

The names of functions and order of the arguments have been chosen to match the Elixir built-in function (but sometimes renamed to avoid name clashes).

jiegillet:
Closing the issue. See more details on #938.

PR #938, opened 2021-09-29
Align list-ops instructions with test expectations (#937)

kahgoh:
* Tests were expecting the `concat` method, not `concatenation`
* Reorder arguments to match the order expected by the tests

### DIFF
--- a/exercises/practice/list-ops/.docs/instructions.md
+++ b/exercises/practice/list-ops/.docs/instructions.md
@@ -11,10 +11,10 @@ track dependent to avoid conflicts with existing names, but the general
 operations you will implement include:
 
 * `append` (*given two lists, add all items in the second list to the end of the first list*);
-* `concatenate` (*given a series of lists, combine all items in all lists into one flattened list*);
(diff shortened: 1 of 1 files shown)


jiegillet:
Thank you for the PR. 
As I mentioned on the issue, this is made clear in the instructions already.
Additionally, the exercise instructions must remain identical to the original implementations in [problems-specifications](https://github.com/exercism/problem-specifications/tree/main/exercises/list-ops), so we cannot change it. We could change `instructions.append.md` but we are already justifying some changes over there, and I believe justifying every little change would be overkill.

neenjaw:
What is the recommended course of action here @jiegillet? It appears by your response that it would be that this PR, while clarifies it for elixir, takes it out of sync with the problem specifications repository so it should be closed. If we want to make the instructions more general we should open an issue there so that the changes flow down stream. Correct?

jiegillet:
That's right, sorry for being ambiguous. I'm closing the PR.

> If we want to make the instructions more general we should open an issue there so that the changes flow down stream. 

That's correct, but the goal of the PR (as I understand it) is to make the explanations less general and more precise for Elixir, so I don't think a change upstream could achieve that.

PR #940, opened 2021-09-30
Updates to test execution documentation

### LINKED ISSUE #906: Mention other potentially useful test runner options in `exercises/shared/.docs/tests.md`
Extend `exercises/shared/.docs/tests.md` with the mention of those useful flags:

- To run tests in order: `--seed 0`
- To run tests until the first failure: `--max-failures 1`
- To run a single specific test: `test/some/particular/file_test.exs:12`
- anything else?

Also we need to fix the text about pending test to explain that only practice exercises have pending test.

czrpb:
Hi!

Here is my documentation style. Looking forward to many critiques so I can improve! :smile: 

Closes #906

### DIFF
--- a/exercises/shared/.docs/tests.md
+++ b/exercises/shared/.docs/tests.md
@@ -1,6 +1,6 @@
-# Running tests in exercism
+# Running Elixir tests in exercism
 
-Tests are written in test files and placed in the `test` directory and implemented
+Elixir tests are written in `.exs` test files and placed in the `test` directory and implemented
 as:
 
 ```elixir
@@ -17,8 +17,8 @@ $ mix test
 
 Useful top-level references include:
 
-* [`ExUnit`: Elixir's unit test library](https://hexdocs.pm/ex_unit/ExUnit.html)
-* [`mix test`: Elixir's test execution tool](https://hexdocs.pm/mix/Mix.Tasks.Test.html)
+* [`ExUnit` - Elixir's unit test library](https://hexdocs.pm/ex_unit/ExUnit.html)
+* [`mix test` - Elixir's test execution tool](https://hexdocs.pm/mix/Mix.Tasks.Test.html)
 
 ### Pending tagged tests
 
@@ -147,13 +147,13 @@ of methods to control what tests are executed.
 A number of useful command-line options exist that control which tests are executed
 based on past or current execution, such as:
 
-* `--failed`: runs only tests that failed the last time they ran
-* `--max-failures`: the suite stops evaluating tests when this number of test failures
-is reached.
-* `--seed`: seeds the random number generator used to randomize the order of tests;
+* `--failed` - runs only tests that failed the last time they ran
+* `--max-failures` - the suite stops evaluating tests when this number of test failures
+is reached
+* `--seed` - seeds the random number generator used to randomize the order of tests;
 `--seed 0` disables randomization so the tests in a single file will always be ran
 in the same order they were defined in
-* `--stale`: runs only tests which reference modules that changed since the last
+* `--stale` - runs only tests which reference modules that changed since the last
 time tests were ran with `--stale`
 
 Documentation:
--- a/exercises/shared/.docs/tests.md
+++ b/exercises/shared/.docs/tests.md
@@ -1,12 +1,13 @@
-# Running tests
+# Running tests in exercism
 
 Tests are written in test files and placed in the `test` directory and implemented
 as:
 
 ```elixir
 test "short test description" do
-  # ... test implementation ...
+  # test implementation
 end
+```
 
 All tests may be executed with:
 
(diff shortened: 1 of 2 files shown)


exercises/shared/.docs/tests.md
```diff
@@ -1,14 +1,109 @@
...
 
-Execute the tests with:
+Tests are written in test files and placed in the `test` directory and implemented
+as:
+
+```elixir
+test "short test description" do
+  ... test implementation ...
+end
+```
```
angelikatyborska, line 10: Let's make this a snippet without syntax errors so that the syntax coloring works:
```suggestion
```elixir
test "short test description" do
  # ... test implementation ...
end
```
```

File: exercises/shared/.docs/tests.md [hunk 2]
```diff
@@ -1,14 +1,109 @@
...
+
+Below is just a summary of some very useful methods; full documentation is here:
+https://hexdocs.pm/mix/Mix.Tasks.Test.html
+
+##### Running tests in a specific file
+
+All tests in a single file may be executed via `mix` by specifying the file:
+
+```bash
+$ mix test test/<FILE>.ex
```
angelikatyborska, line 39: Test files are always `.exs` files (also applies to a few other code blocks below)`

exercises/shared/.docs/tests.md
```diff
@@ -1,14 +1,109 @@
...
+7:      end
+8:
+9:      test "Test 2" do
+10-21:    . . .
+22:     end
+23:
+24:     test "Test 3" do
+25-35:    . . .
+36:     end
+```
```
angelikatyborska, line 68: Could this also be somehow reformatted to make it correct Elixir code? Maybe the line numbers can come as comments at the end, like this?

```elixir
test "Test 1" do    # 1
  # . . .           # 2-6
end                 # 7
```

angelikatyborska (commented):
The additions are great! Good job. It reads very well, the wording fits the style of other documents.

Here are 

I would just want a small change to the information order. In case somebody is impatient, I want them to see at the very top: here's how you run all tests (`mix test`) and here's how you unskip tests, which are the two basic things you need to know to solve the exercise. Everything else that's "nice to know" but not essential should come afterwards.

This would roughly mean that I would like the beginning of the document to stay as it was, and then after the `## Pending tests` section we can start explaining how a test block looks like, which testing framework we're using and what other extra options are there.

As a rule, we try to write all code snippets in a way that they are actual correct code snippets without syntax errors. They don't have to compile, they just need to be semantically correct so the the syntax highlighting will manage to color them correctly. So, to mark a line that doesn't do anything, we would write `# ...` instead of `...` (*), or to mark a string whose content doesn't matter, we would write `"..."` instead of `...`.

Another general thing that could be improved is how links are written. We should make use of markdown's link syntax and provide nice link text instead of relying on the platform turning raw URLs into links, so for example, instead of:

```markdown
Elixir's test execution tool: https://hexdocs.pm/mix/Mix.Tasks.Test.html
```

Write:

```markdown
[Elixir's test execution tool](https://hexdocs.pm/mix/Mix.Tasks.Test.html)
```

exercises/shared/.docs/tests.md
czrpb: will do!

exercises/shared/.docs/tests.md
czrpb: also, will do!

exercises/shared/.docs/tests.md
czrpb: pushed

czrpb:
all suggestions/recommendations made!

thx!!

angelikatyborska:
I'm assuming this is ready for review even though the PR is a draft because I can't see anything missing and I will make it not-a-draft :)

angelikatyborska:
I reviewed how the file `exercises/shared/.docs/tests.md` is used in practice and I realized it has to be relatively short. Its only usage is to be included in the `HELP.md` file that students download with each exercise. `HELP.md` starts with the content of `exercises/shared/.docs/tests.md`, but then it also includes instructions on how to submit the solution and where to look for help. If `exercises/shared/.docs/tests.md` would be very long, the students would have a hard time finding instructions on how to submit the solution, which would be pretty bad...

I took all your additions and put them in `docs/TESTS.md` instead. That file can be as long and detailed as we want. It's rendered here: https://exercism.org/docs/tracks/elixir/tests

angelikatyborska (approved):

angelikatyborska:
Thank you a lot for your contribution 💜 I'll merge it now. I had to do some other minor changes, I decided to do them myself not to bother you anymore. If I made a mistake there, let me know.

czrpb:
of course and i hope i was not more effort than value! :wink: 

and thank you!

PR #942, opened 2021-10-01
produce better test failures for date-parser

Cohen-Carlisle:
this change allows those implementing a function and then running the
tests (TDD style) to see more useful info on test failure to help drive
their solutions. before this change, all empty functions would result
in a function clause error, which obscured the information necessary
to TDD the code. after this change, because the right type is returned,
the error messages are more useful as the actual assertion has failed.


jiegillet:
I have to say I like this change. This exercise is approaching regexes with an unusual angle, that I really enjoyed learning, but it took me a while to understand the structure. This provide a nice hint of what is expected. Specs would work as well, but not at this stage of the syllabus.

angelikatyborska:
I'm all for making this exercise more friendly and produce better test failures, but I'm not 100% if modifying the stub is the correct way to go.

Sure, we can do it if that's the only option, but I was thinking we should rather modify the tests to produce more useful failures, especially when run on the stub when every function returns `nil`. Modifying the test also improves the experience for students that modified the stub and are still struggling.

I tried it out with the first test, "un-padded days". I changed every call of `Regex.compile!(DateParser.day()))` to `Regex.compile!("#{DateParser.day()}"))` and... the whole tests passes. For an empty implementation... that also means it passes when we don't change the test, but change the stub to return an empty string. In fact with the new stub only 26 out of 51 tests fail at the beginning. I find that rather weird 🤔.

jiegillet:
I agree that there is some work to be done on the tests, but really the stub is missing something, because right now you start the exercise and see 
```elixir
  def day() do
     # Please implement the day/0 function
   end
```
And it's quite unclear what to do. Of course, knowing a little about regex, I tried things like `~r/\d/` and it took me a while to figure out I was only expected to return a string. Things that would help:
* Adding a `@spec`
* Adding a `@doc`
* Replacing `# Please implement the day/0 function` by something like 
* Adding `""`/`~r//` stubs

Exercism has always been very TDD, but let's not go overboard :)

angelikatyborska:
>  it took me a while to figure out I was only expected to return a string

To be fair, the instructions for each step do include the word "string":

> Implement day/0, month/0, and year/0 to return a string pattern which...


Another idea would be for the first test for each function to only assert on the return type, which would make you realize immediately where a string was expected and where a regex.

> Adding a @spec

The regex exercise doesn't have any dependent learning exercises, so we could just freely make it depend on the typespec learning exercise. But that goes against the pattern that all learning exercises follow, which is not using typespecs, but providing a detailed description of the expected return value in the instructions, so I'm not crazy about that idea. Modifying the return value in the stub or changing tests does sound better to me.

> Exercism has always been very TDD, but let's not go overboard :)

What do you mean by going overboard?

jiegillet:
> To be fair, the instructions for each step do include the word "string":

That's true. My fault for trying to solve the exercise without reading the instructions. I've always been that way, I would never read the building instruction in my kinder egg gizmos for the extra challenge :)
 
> Another idea would be for the first test for each function to only assert on the return type, which would make you realize immediately where a string was expected and where a regex.

 I didn't think of that , I like it!

> The regex exercise doesn't have any dependent learning exercises, so we could just freely make it depend on the typespec learning exercise. But that goes against the pattern that all learning exercises follow, which is not using typespecs, but providing a detailed description of the expected return value in the instructions, so I'm not crazy about that idea. Modifying the return value in the stub or changing tests does sound better to me.

Yeah, I agree that it sticks out from the others. Not my favorite either but I had to mention it.

> > Exercism has always been very TDD, but let's not go overboard :)
> 
> What do you mean by going overboard?

Well I mean that usually, you can at least figure out figure out the basics of what to do from the function name and arguments.  Here the function is `def day`, which is very little. You really have to dig in the instructions and text. Which is completely fair for a concept exercise. 
To be clear, I like this exercise, it taught me a lot, I had never used `Regex.compile!`. It was just a bit harder to get started than most, but maybe that's why I learned more. 

I really like your suggestion of adding tests to assert the return type. That will be the first thing you run into.

Cohen-Carlisle:
I like the idea of having it explicitly test for the type. I'll change this PR to do that.

angelikatyborska (approved):
Great! Feel free to add yourself in `.meta/config.json` as this exercise's contributor 🙂

PR #945, opened 2021-10-03
Fix wording in library-fees instructions

wiserfirst:
In the instructions for task 4, it looks like to me the words `planned` and `actual` should swapped.


jiegillet (approved):
Indeed! 
Thank you.

PR #946, opened 2021-10-03
Only wrap id in square brackets

linusdm:
The badge template is a bit confusing and should only expect the employee id to be wrapped in square brackets.


angelikatyborska (approved):
Right! Thank you 💜

PR #947, opened 2021-10-03
Fix small error in remote-control-car instructions

jlirochon:


jiegillet (approved):
Thank you very much for fixing this!

PR #949, opened 2021-10-04
Fix key name for inventory items

wiserfirst:
The key for inventory items should be `quantity_by_size` instead of `quantity_per_size`


jiegillet (approved):
This is an interesting mistake! Thanks for finding it!

Issue #948, opened 2021-10-04
Redirecting to Page Not Found in Elixir School


ThiagoPMaceda:
While I was reading through the basics on Elixir(https://exercism.org/tracks/elixir/concepts/basics) I noticed that clicking in the pattern match symbol "=" link was redirecting to a page not found in Elixir School, that is happening because the link was changed (probably with the recent update of Elixir School). Below is the link that I was clicking.
![image](https://user-images.githubusercontent.com/30845442/135859833-3f7c6741-8a78-4ab2-8803-11bc20ee807c.png)

jiegillet:
Looks like the `pattern-matching` in the URL should be `pattern_matching`.

jiegillet:
Closed via #953.

PR #953, opened 2021-10-08
fix: pattern matching exercism link

george124816:
small fix on elixirschool link 

solves issue #948


angelikatyborska:
Thanks for the list @jiegillet 😁 
That one also needs fixing:
```
concepts/basics/about.md:[match]: https://elixirschool.com/en/lessons/basics/pattern-matching/
```

The rest still works.

@george124816 could you also fix the link in `concepts/basics/about.md`?

angelikatyborska:
Thanks :rocket

PR #954, opened 2021-10-10
New concept exercise: `new-passport` (concept: `with`)

jiegillet:
Daft for #560.
I had a fun idea so I put a draft together. I only added the empty files, `design.md` and the exemplar for now. 

The idea is about dealing with the frustrations of getting a new passport from the city office. First, you have to get the time right (they are basically never open), you have to find the right counter to go to (the information desk takes long coffee breaks), the counter you should go to depends on your age, and you better have brought the form of the right color!

With so many potential point of failures, you better use `with`. 
Let me know what you think of the idea and I'll work on it more.

### DIFF
--- a/concepts/with/introduction.md
+++ b/concepts/with/introduction.md
@@ -4,21 +4,9 @@ The [special form with][with] provides a way to focus on the "happy path" of a s
 
 ```elixir
 with {:ok, id} <- get_id(username),
-     {:ok, avatar} when is_bitstring(avatar) <- fetch_avatar(id),
-     avatar_size = bit_size(avatar),
+     {:ok, avatar} <- fetch_avatar(id),
      {:ok, image_type} <- check_valid_image_type(avatar) do
-  {:ok, image_type, avatar_size, avatar}
-end
-```
-
-At each step, if a clause matches, the chain will continue until the `do` block is executed. If one match fails, the chain stops and the non-matching clause is returned. You have the option of using an `else` block to catch failed matches and modify the return value.
-
-```elixir
-with {:ok, id} <- get_id(username),
-     {:ok, avatar} when is_bitstring(avatar) <- fetch_avatar(id),
-     avatar_size = bit_size(avatar),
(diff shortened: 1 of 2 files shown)


jiegillet:
Added tests.
The CI fails, maybe because `DateTime.add` was added in 1.8? I'll fix it later

jiegillet:
Ok, I think it's ready for review.

Some things I'm unsure about:
* My hints or maybe the difficulty balance between tasks? I realized that only task 2 has any meat. Everywhere else the hints are stating the obvious. Not sure if it's a problem.
* Which exercise to add as practicing `with`? 
    * To me `phone-number` is an [obvious example](https://exercism.org/tracks/elixir/exercises/phone-number/solutions/jiegillet) of using `with`, but I didn't find any other people using it. 
    * I used it a lot in `sgf-parsing` but @neenjaw (only other solution so far) never touched it
    * I grepped all exemploid (™️) solutions that use `with`, only `sgf-parsing` and `variable-length-quantity` but the latter hardly qualifies as a prerequisite even less a practice.

I still like `with` even if it's not super popular though, probably because it is a mini-monad 🐱

jiegillet:
That is THE single largest most detailed review I've ever got. Thank you for taking the time and leaving all those actionable suggestions :)

jiegillet:
> I found myself not changing the return value of the function until the last task because none of the tests cared. I also noticed that the step `counter = birthday_to_counter.(birthday)`, even though listed as under task 2, doesn't need to be added to make task 2 tests pass.
> 
> One thing that is a bit unusual about this exercise is that the instructions have no return value examples. That probably cannot be added if only the last step makes the function return the correct "happy path" value? Or maybe the error paths can be shown instead?

That is quite unusual indeed, but I couldn't really find ways to get around it that would make the firsts tests still pass after modifying the function. I suppose we could build 1 function per task but that would feel clunky.

Maybe I could add a step `go_to_counter(now, birthday, counter)` that checks that it is correct? It wouldn't add so much to the exercise, but it also doesn't hurt.

PR #968, opened 2021-10-16
Replace "TODO:" with "please" in lasagna's boilerplate

angelikatyborska:
I would like to slightly reword the comments in this boilerplate file. Removing "TODO" from them will make it easier to implement the analyzer check https://github.com/exercism/elixir-analyzer/issues/195 that needs to tell apart our own boilerplate comments from user's own TODO/FIXME comments.

### DIFF
--- a/exercises/concept/lasagna/lib/lasagna.ex
+++ b/exercises/concept/lasagna/lib/lasagna.ex
@@ -1,11 +1,11 @@
 defmodule Lasagna do
-  # TODO: define the 'expected_minutes_in_oven/0' function
+  # Please define the 'expected_minutes_in_oven/0' function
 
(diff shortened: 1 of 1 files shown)


shansiddiqui94:
Lasagna exercise makes no sense, the hints are not helpful, and there is no solution or follow up video

iHiD:
@shansiddiqui94 Sorry you struggled on this. Your comment isn't particularly constructive as it doesn't say *what* is wrong or *why*. Tens of thousands of people have completed Lasagna across multiple languages so lots of people find it ok as an exercise, but we'd love to improve it if you can give some specific details on which bit of it you found too hard for you. 

If you could open a **new issue** for that, rather than commenting on this PR, that will allow our team to respond to you more easily. Thank you.

PR #983, opened 2021-10-29
Update introduction.md

ryanwolhuter:
Tiny grammar fix.


angelikatyborska:
Thank you! There is one more file with exactly the same content and the same typo. Could you fix it there too?

jiegillet (approved):
Thank you for finding the typo.

PR #989, opened 2021-11-01
Bring complex-numbers up to date with canonical-data

jiegillet:
Little bit of a breaking change, from
```elixir
@spec div(a :: complex, b :: complex) :: complex
```
to
```elixir
@spec div(a :: complex | float, b :: complex | float) :: complex
```

### DIFF
--- a/exercises/practice/complex-numbers/lib/complex_numbers.ex
+++ b/exercises/practice/complex-numbers/lib/complex_numbers.ex
@@ -1,6 +1,6 @@
 defmodule ComplexNumbers do
   @typedoc """
-  In this module, complex numbers are represented as a tuple-pair containing the real and
+  In this module, complex numbers, or a real and a complex number are represented as a tuple-pair containing the real and
   imaginary parts.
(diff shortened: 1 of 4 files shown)


exercises/practice/complex-numbers/lib/complex_numbers.ex
```diff
@@ -1,6 +1,6 @@
 defmodule ComplexNumbers do
   @typedoc """
-  In this module, complex numbers are represented as a tuple-pair containing the real and
+  In this module, complex numbers, or a real and a complex number are represented as a tuple-pair containing the real and
```
angelikatyborska, line 3: This addition doesn't sound right. Real numbers aren't represented as tuples

angelikatyborska (approved):
LGTM overall, but take a look at my one comment before merging 🙂

exercises/practice/complex-numbers/lib/complex_numbers.ex
jiegillet: Oh, this is a search and replace accident. Sorry about that.
```suggestion
  In this module, complex numbers are represented as a tuple-pair containing the real and
```

PR #990, opened 2021-11-02
Fix [head | tail] list notation example

sdoering01:
Before

```elixir
iex(1)> [1 | [2 | 3 | []]]
** (CompileError) iex:1: misplaced operator |/2

The | operator is typically used between brackets as the cons operator:

    [head | tail]

where head is a single element and the tail is the remaining of a list.
It is also used to update maps and structs, via the %{map | key: value} notation,
and in typespecs, such as @type and @spec, to express the union of two types
    (stdlib 3.16.1) lists.erl:1358: :lists.mapfoldl/3
    (stdlib 3.16.1) lists.erl:1359: :lists.mapfoldl/3
    (stdlib 3.16.1) lists.erl:1358: :lists.mapfoldl/3
```

After

```elixir
iex(1)> [1 | [2 | [3 | []]]]
[1, 2, 3]
```

### DIFF
--- a/concepts/lists/introduction.md
+++ b/concepts/lists/introduction.md
@@ -16,7 +16,7 @@ Elixir implements lists as a linked list, where each node stores the reference t
 [1 | []]
 
 # [1, 2, 3] represented in [head | tail] notation
-[1 | [2 | 3 | []]]
+[1 | [2 | [3 | []]]]
 ```
 
 We can use _`[head | tail]`_ notation to prepend elements to a list:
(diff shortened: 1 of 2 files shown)


jiegillet:
Good catch, thank you!
This typo is guaranteed to be in at least one other place. Could you find it and fix it there too?

sdoering01:
Ah yes, I didn't check for that. That should be the other place -- I grep'd on the whole repo.

jiegillet (approved):
Awesome! Thank a million!

PR #991, opened 2021-11-02
Remove public functions to match the stubs

jiegillet:
Related to [this analyzer check](https://github.com/exercism/elixir-analyzer/pull/207).

I have compared all the exemploids to the stubs and listed all the functions that were public in the exemploid but not in the stubs. Then I basically added a whole bunch of `p` until the problem went away. Some notes:

- Concept exercises: 3 functions have empty stubs, nothing we can do there except change the analyzer comment
- `circular-buffer`, `bank-account` and `react` have some GenServer `@impl true` functions that are not in the stubs but cannot be private. This is fine because students are expected to write those and their solution will be compared to the examples that have them too.
- `clock` implements its own `to_string` that is not in the stub, but required by the tests. Once again, this is fine for the same reason than above.
- `parallel-letter-frequency` is strange, I think it used to use `:rpc.pmap` but then it was commented and I don't think it's doing anything in parallel anymore. [My solution does](https://exercism.org/tracks/elixir/exercises/parallel-letter-frequency/solutions/jiegillet) if we want to replace it.
- I couldn't help but change some implementations a tiny little bit, sometimes using the notation `do_thing` because otherwise `thing` ends up with several arities. I also had to remove some unused functions.

### DIFF
--- a/exercises/practice/parallel-letter-frequency/.meta/example.ex
+++ b/exercises/practice/parallel-letter-frequency/.meta/example.ex
@@ -7,7 +7,7 @@ defmodule Frequency do
   def frequency(texts, workers) do
     groups = Enum.map(0..(workers - 1), &stripe(&1, texts, workers))
     # :rpc.pmap({Frequency, :count_texts}, [], groups)
-    Enum.map(groups, &Frequency.count_texts/1)
+    Enum.map(groups, &count_texts/1)
     |> merge_freqs()
   end
 
@@ -16,8 +16,8 @@ defmodule Frequency do
   end
 
   # Needs to be public because of how it's invoked by `:rpc.pmap/4`.
-  @doc false
-  def count_texts(texts) do
+  # @doc false
+  defp count_texts(texts) do
     Enum.map(texts, &count_text/1)
     |> merge_freqs()
   end
--- a/exercises/practice/pov/.meta/example.ex
+++ b/exercises/practice/pov/.meta/example.ex
@@ -46,81 +46,53 @@ defmodule Pov do
     end
   end
 
-  def search(%Zipper{focus: {node, _children}} = zipper, node), do: {:ok, zipper}
+  defp search(%Zipper{focus: {node, _children}} = zipper, node), do: {:ok, zipper}
 
-  def search(%Zipper{} = zipper, node) do
+  defp search(%Zipper{} = zipper, node) do
     case zipper |> down |> search(node) do
       {:ok, z} -> {:ok, z}
       _ -> zipper |> right |> search(node)
     end
   end
 
-  def search(nil, _node), do: nil
+  defp search(nil, _node), do: nil
 
-  def reparent(%Zipper{focus: tree, genealogy: []}), do: tree
+  defp reparent(%Zipper{focus: tree, genealogy: []}), do: tree
 
-  def reparent(%Zipper{
-        focus: {node, children},
-        genealogy: [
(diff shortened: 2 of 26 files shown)


exercises/practice/pythagorean-triplet/.meta/example.ex
```diff
@@ -49,11 +49,9 @@ defmodule Triplet do
     multiples ++ generate(next ++ triplets, sum)
   end
 
-  @doc """
-  Given a primitive Pythagorean triple, generate three more.
-  [All primitive Pythagorean triples can be generated this way without duplication][https://en.wikipedia.org/wiki/Tree_of_primitive_Pythagorean_triples].
-  """
-  def next_triplets([a, b, c]),
```
jiegillet, line 56: Private functions cannot have docs, I just learned :)

jiegillet (commented):
Here are some extra comments.

exercises/practice/parallel-letter-frequency/.meta/example.ex
```diff
@@ -7,7 +7,7 @@ defmodule Frequency do
   def frequency(texts, workers) do
     groups = Enum.map(0..(workers - 1), &stripe(&1, texts, workers))
     # :rpc.pmap({Frequency, :count_texts}, [], groups)
-    Enum.map(groups, &Frequency.count_texts/1)
```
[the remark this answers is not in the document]
angelikatyborska: Yup, this solution is not "parallel" in any way. Feel free to replace it with your own in a separate PR.

exercises/practice/pov/.meta/example.ex
```diff
@@ -46,81 +46,53 @@ defmodule Pov do
...
-      }) do
-    %Zipper{
-      focus: left,
-      genealogy: [
-        %Crumb{crumb | left_siblings: lefties, right_siblings: [tree | right]} | grandparents
-      ]
-    }
-  end
-
-  def left(_zipper), do: nil
```
angelikatyborska, line 116: Help me out here - did we have dead code in the example?

exercises/practice/sgf-parsing/.meta/example.ex
```diff
@@ -175,19 +168,15 @@ defmodule SgfParsing do
...
-  def drop_and(p1, p2) do
+  defp drop_and(p1, p2) do
     fn string ->
       with {:ok, _, rest} <- p1.(string) do
         p2.(rest)
       end
     end
   end
 
-  def inject(a) do
```
angelikatyborska, line 186: `string`, `inject` - also dead code?

angelikatyborska (approved):

exercises/practice/pov/.meta/example.ex
jiegillet: Yes we did, that's my fault. Basically for this one and `sgf-parsing` I built a library (in this one, moving thought a tree with zipper, in the other a parser combinator) and the dead code is code that would definitely belong in the the library that I thought I would use but in the end I didn't.

exercises/practice/sgf-parsing/.meta/example.ex
jiegillet: Yup. I was very surprised to realize I didn't need them in the end, but I didn't have the heart to delete them. Bad Jie!

angelikatyborska: Sounds like somebody didn't pay too much attention to the example solutions when doing the code review for the new exercises 😁

PR #996, opened 2021-11-06
Added an example using strings as keys.

chriseyre2000:


angelikatyborska:
On Slack, you wrote that the rationale behind this addition is to make it easier to solve the exercise, right? In that case, the change needs to also happen in those files:
- `exercises/concept/basketball-website/.docs/introduction.md`
- `concepts/access-behaviour/introduction.md`

The file `concepts/access-behaviour/about.md` is only visible to students *after* solving the exercise.

neenjaw:
The `about.md` is only shown after the exercise is complete.  If you think that it also requires an example before the exercise is complete, I think also needs to be added to the `introduction.md` in the same directory.

angelikatyborska (approved):
Looks good to me! Thanks 🙂

PR #997, opened 2021-11-06
Explain the return of the get_counter function

chriseyre2000:
The exercise implies that the get_counter function returns {:ok, counter} when it actually returns counter.
This corrects the text to match the code.

### DIFF
--- a/exercises/concept/new-passport/.docs/instructions.md
+++ b/exercises/concept/new-passport/.docs/instructions.md
@@ -18,7 +18,7 @@ If you get an `:error` tuple, use the `else` block to return it.
 
 The information desk is notorious for taking long coffee breaks. If you are lucky enough to find someone there, they will give you an instruction manual which will explain which counter you need to go to depending on your birth date.
 
-Call the function `find_counter_information/1` with the current time. You will get either a tuple with `:ok` and a manual, represented by an anonymous function, or a tuple with `:coffee_break` and more instructions. In your happy path where you receive the manual, apply it to you birthday (second argument of `get_new_passport/3`) and learn which counter to go to. Return an `:ok` tuple with the counter.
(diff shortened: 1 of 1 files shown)


exercises/concept/new-passport/.docs/instructions.md
```diff
@@ -18,7 +18,7 @@ If you get an `:error` tuple, use the `else` block to return it.
 
 The information desk is notorious for taking long coffee breaks. If you are lucky enough to find someone there, they will give you an instruction manual which will explain which counter you need to go to depending on your birth date.
 
-Call the function `find_counter_information/1` with the current time. You will get either a tuple with `:ok` and a manual, represented by an anonymous function, or a tuple with `:coffee_break` and more instructions. In your happy path where you receive the manual, apply it to you birthday (second argument of `get_new_passport/3`) and learn which counter to go to. Return an `:ok` tuple with the counter.
+Call the function `find_counter_information/1` with the current time. You will get either a tuple with `:ok` and a manual, represented by an anonymous function, or a tuple with `:coffee_break` and more instructions. In your happy path where you receive the manual, apply it to you birthday (second argument of `get_new_passport/3`) and learn which counter to go to (function simply returns the counter number). Return the counter.
```
jiegillet: I'm good with Angelika's suggestions

angelikatyborska, line 21: Hmm, I'm not so sure here. In the previous version, there was this sentence:

> Return an `:ok` tuple with the counter.

This was a technically correct sentence. It was describing what you are supposed to return from `get_new_passport`. But I can totally see how it could mislead people to believe that the "manual'' function also returns a tuple, so I agree this needs to be rephrased. However, the new version :

> Return the counter.

Is not correct, you're not supposed to return *just* the counter form `get_new_passport`, but `{:ok, counter}`.

I would suggest something like this:

```suggestion
Call the function `find_counter_information/1` with the current time. You will get either a tuple with `:ok` and a manual, represented by an anonymous function, or a tuple with `:coffee_break` and more instructions. In your happy path where you receive the manual, apply it to you birthday (second argument of `get_new_passport/3`). It will return the number of the counter where you need to go. Return an `:ok` tuple with that counter number.
```

angelikatyborska (approved):
Perfect, thank you 🎉

PR #999, opened 2021-11-08
Change BoutiqueSuggestions.get_combinations output

JuliaMathias:
When I was doing this exercise, I noticed that on the examples, the output would only show the item name. So when I was solving the exercise, initially I did it in a way where the function would only output the name of each item. But then that made tests fail. So I think it would be clearer to show the actual output expected on the examples, even if the output gets a little long.

### DIFF
--- a/exercises/concept/boutique-suggestions/.docs/instructions.md
+++ b/exercises/concept/boutique-suggestions/.docs/instructions.md
@@ -27,9 +27,9 @@ bottoms = [
 ]
 BoutiqueSuggestions.get_combinations(tops, bottoms)
 # => [
-#      {%{item_name: "Dress shirt"}, %{item_name: "Jeans"}}
-#      {%{item_name: "Dress shirt"}, %{item_name: "Dress trousers"}}
-#      {%{item_name: "Casual shirt"}, %{item_name: "Jeans"}}
+#      {%{item_name: "Dress shirt"}, %{item_name: "Jeans"}},
+#      {%{item_name: "Dress shirt"}, %{item_name: "Dress trousers"}},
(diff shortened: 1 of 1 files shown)


exercises/concept/boutique-suggestions/.docs/instructions.md
```diff
@@ -49,8 +49,10 @@ bottoms = [
 ]
 BoutiqueSuggestions.get_combinations(tops, bottoms)
 # => [
-#      {%{item_name: "Dress shirt"}, %{item_name: "Dress trousers"}}
-#      {%{item_name: "Casual shirt"}, %{item_name: "Jeans"}}
+#      {%{item_name: "Dress shirt", base_color: "blue"},
+#       %{item_name: "Casual shirt", base_color: "black"}},
+#      {%{item_name: "Jeans", base_color: "blue"},
+#       %{item_name: "Dress trousers", base_color: "black"}}
```
angelikatyborska, line 55: Something's wrong here, tops don't go with tops 🤔 
```suggestion
#      {%{item_name: "Dress shirt", base_color: "blue"},
#       %{item_name: "Dress trousers", base_color: "black"}},
#      {%{item_name: "Casual shirt", base_color: "black"},
#       %{item_name: "Jeans", base_color: "blue"}}
```

angelikatyborska (commented):
Good idea 👍 but there is a small mistake in one of the examples

JuliaMathias:
> Good idea +1 but there is a small mistake in one of the examples

yes, sorry! Everything should be ok now 😊

angelikatyborska (approved):

PR #1000, opened 2021-11-09
Refactor anonymous functions into captured functions

jiegillet:
See [this issue](https://github.com/exercism/elixir-analyzer/issues/218).

This PR changes exemploid files that use anonymous functions instead of captures (and some other very minor refactors that I couldn't stop myself from doing).


angelikatyborska (approved):
Nice

PR #1001, opened 2021-11-10
fix typos in rpg-character-sheet introduction

jaimeiniesta:


jiegillet:
Thank you! 
This typo is guaranteed to be at least another file, could you do a global search and find it?

Issue #1005, opened 2021-11-11
Grammer error in Bird Count -> Calculate the number of busy days


moviedo:
Some days are busier` that` others. A busy day is one where five or more birds have visited your garden.
Should be  ` than`.

PR #1006, opened 2021-11-11
fix(bird-count .docs): grammer issue question 5 GH-1005

moviedo:
Closes https://github.com/exercism/elixir/issues/1005


jiegillet (approved):
Thanks a lot!

PR #1007, opened 2021-11-16
Update instructions for resistor-color

jaimeiniesta:
This exercise previously had a test to get the list of all colors, but it was removed later to simplify the exercise. It's still referenced in the instructions, so better remove it.

### DIFF
--- a/exercises/practice/resistor-color/.docs/instructions.append.md
+++ b/exercises/practice/resistor-color/.docs/instructions.append.md
@@ -1,0 +1,3 @@
+# Instructions append
+
+For this exercise, returning the list of colors is not required.


jiegillet:
Thank you, unfortunately we can't change the instructions because they need to stay in sync with the [problem-specifications](https://github.com/exercism/problem-specifications/tree/main/exercises/resistor-color) repo.

The best we can do is add an `instructions.append.md` [like here](https://github.com/exercism/elixir/tree/main/exercises/practice/square-root/.docs) with extra explanations, but I'm not sure what's the best move between leaving things as they are and bringing attention to something that's not there.

jaimeiniesta:
Thanks, I understand. Can we then update https://github.com/exercism/problem-specifications/blob/main/exercises/resistor-color/description.md instead?

Or maybe we can bring back the test to ask for the list of colors?

jiegillet:
We can't do that, because the instructions are shared among all tracks, and I'm sure other tracks do implement the list of colors.

We dropped the list of color test because we wanted the exercise to get unlocked before resistor color duo and trio, but this one required `Enum` to produce the list idiomatically, so it would have been unlocked last. See #807.

jaimeiniesta:
Then I'm not sure what to do here. Some students are confused by the instructions asking the list of colors and the tests not requiring that.

I think that the best option is adding an append to explain that for this first exercise, we don't need to implement the list of colors. Shall I proceed with that?

angelikatyborska:
> I think that the best option is adding an append to explain that for this first exercise, we don't need to implement the list of colors. Shall I proceed with that?

@jaimeiniesta Yes, please!

jaimeiniesta:
@angelikatyborska done, I've reverted the changes to `instructions.md` and added a new `instructions.append.md`. I think the commits can be squashed when merging the PRs, right? I edited this via Github editor and couldn't rebase easily.

jaimeiniesta:
PR squashed, there's only 1 commit now to add the new append file.

jiegillet (approved):
That's a simple change that will help some people, thank you for doing it!

PR #1008, opened 2021-11-17
Ensure value is integer in tests for resistor-color-trio

jaimeiniesta:
I've seen some students use float division instead of integer division when converting to kiloohms, so we better use `===` for the assertion so we can catch this.

### DIFF
--- a/exercises/practice/resistor-color-trio/.meta/example.ex
+++ b/exercises/practice/resistor-color-trio/.meta/example.ex
@@ -15,7 +15,7 @@ defmodule ResistorColorTrio do
   @doc """
   Calculate the resistance value in ohm or kiloohm from resistor colors
   """
-  @spec label(colors :: [atom]) :: {integer, atom}
+  @spec label(colors :: [atom]) :: {number, :ohms | :kiloohms}
   def label([a, b, c]) do
     value = (10 * @colors[a] + @colors[b]) * round(:math.pow(10, @colors[c]))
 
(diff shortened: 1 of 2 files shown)


jiegillet (commented):
It's true that the typespecs expect an integer, so it's a good thing to check.
However, I fear that people who will fail the `===` test won't easily understand what the problem is. What would you think about adding an explicit test about the type?
```elixir
assert is_integer(output)
assert output === expected
```

This of course won't guarantee that they are using `div/2`, maybe they will using `&//2` and then `floor/1`. If we want to check for that, we can do an automated analyzer check, but that's a bit more involved.

jaimeiniesta:
You're right that an explicit test is better, I added it and left the others as they were.

jaimeiniesta:
Thanks, @angelikatyborska I think you're right, I was going to say that resistor color codes could only express integer values but then I learnt more about them and it turns out they can also express floats using the colors `:gold` and `:silver`. Those are not included in this exercise, but it makes sense then to allow a `number` in the typespec.

jiegillet:
Either choice is fine by me. 
Of course resistance is a real number, but in this exercise (without `:gold` and `:silver`) the _declared values_ of the resistance as defined by the colors can only be integers, so requiring integers make sense as well.

jaimeiniesta:
@jiegillet yes that was my reasoning but I'm OK with what @angelikatyborska feels like it's the best choice.

angelikatyborska:
Great, than let's just change the typespec 🙂 @jaimeiniesta can you do that?

jaimeiniesta:
@angelikatyborska sure thing, done! 🙂

jiegillet:
Thank you! Could you change it in the example solution too?

jaimeiniesta:
Done!

angelikatyborska (approved):
Perfect, thank you 💜

PR #1009, opened 2021-11-18
Add www to links to erlang doc

jraregris:
The erlang.org site seems to have changed their url scheme recently, and
left all our links dead. Adding www to the urls fixes this.


jiegillet (approved):
Thanks a lot that's very helpful! And it looks like you got them all too.

PR #1019, opened 2021-12-05
Support 1.13; drop 1.7

angelikatyborska:
There were two problems in example solutions:
- in `yacht`, `score/2` is a public function with a defined contract (the first argument should be an atom) but the same function was defined to also accept an integer.
- in `variable-length-quantity`, reversing a bitstring no longer works with `String.reverse` if the bitstring is not a valid string.


jiegillet (approved):
Looks good, everything seems to run fine!

PR #1023, opened 2022-01-07
Updated linked_list_test.exs. line #61

gerald-annan:
Tailing the original list produces a new list with the tail of the previous list. Test to peek list on #61 should result to {:ok, :b}


angelikatyborska:
This change doesn't look quite right to me. `peek` should return the head of the list. If a head of a list is first equal to `:b`, and then I "remove it" by calling `tail` (which returns the list without the head), `:b` is no longer the head, so the second call to `peek` cannot return` :b`. @jiegillet can you double-check this one?

jiegillet:
Yes, definitely not right, I'll close the PR.

PR #1026, opened 2022-01-10
Sync metadata and docs with configlet

angelikatyborska:
All the changes in this PR were done automatically by configlet.


angelikatyborska:
Formatting issues in problem-specs that tripped up our CI will be fixed in: https://github.com/exercism/problem-specifications/pull/1895

jiegillet (approved):
Some things do not seem consistent, `*` becoming `\*` in some files but ignored in others, sometimes both on the _same line_ (`prime-factor`), some `-` bullet points turned into `*` in `resistor-color-duo` and the exact opposite in `resistor-color-trio`...

Of course, I understand this is an issue for `problem-specifications`...

angelikatyborska:
I have no idea about the logic for escaping `*` (it does seem weird to me too), but for the `*` vs `-` lists, the linter just uses the marker that was used first in the same document for all the lists in that document. It would be nicer if it always used the same marker everywhere :|

PR #1029, opened 2022-01-11
bsd/gnu sed compatibility hotfix; resolves #1028

### LINKED ISSUE #1028: Fix test_exercises.sh to work with both GNU and BSD sed versions
See:
https://github.com/exercism/elixir/blob/d00ca743340fcc328c70ee351274f91b57fd4c8d/bin/test_exercises.sh#L96-L97

Ideally this would detect which sed version is installed, or at least allow the user to choose by passing a flag.

justrp:


angelikatyborska (approved):
Works on my machine (macOS), works on CI (linux) - perfect 🙂

PR #1032, opened 2022-01-14
Deprecate accumulate

angelikatyborska:
Following: https://github.com/exercism/problem-specifications/pull/1918


jiegillet (approved):
What exactly happens now? It won't appear on the tree since it doesn't;t practice anything, but it will still be in the full list?

angelikatyborska:
It will disappear completely from anywhere except for students that already started it, they will see it on the full list. E.g. both "binary" and "hexadecimal" are deprecated. I can find "binary" on my exercise list because I solved it, but not "hexadecimal".

PR #1035, opened 2022-01-15
allow await_many as well as await when expecting timeout message

madlep:
Update to `assert_receive` expectation on timeout in "RPN Calculator Inspection" exercise to allow a timeout message with `:await_many` as the MFA tuple function (as well as allowing current `:await`)

My solution for that exercise that causes the test to incorrectly fail was:

```elixir
  def correctness_check(calculator, inputs) do
    inputs
    |> Enum.map(& fn -> calculator.(&1) end)
    |> Enum.map(&Task.async/1)
    |> Task.await_many(100)
  end
```

Where as the exemplar is:

```elixir
  def correctness_check(calculator, inputs) do
    inputs
    |> Enum.map(fn input -> Task.async(fn -> calculator.(input) end) end)
    |> Enum.map(fn task -> Task.await(task, 100) end)
  end
```

The changes in the test allow both to pass.

### DIFF
--- a/exercises/concept/rpn-calculator-inspection/test/rpn_calculator_inspection_test.exs
+++ b/exercises/concept/rpn-calculator-inspection/test/rpn_calculator_inspection_test.exs
@@ -343,7 +343,11 @@ defmodule RPNCalculatorInspectionTest do
 
       Process.flag(:trap_exit, true)
       pid = spawn_link(fn -> RPNCalculatorInspection.correctness_check(calculator, inputs) end)
-      assert_receive {:EXIT, ^pid, {:timeout, {Task, task_fn, [_task, 100]}}} when task_fn in [:await, :await_many], 150
+
+      assert_receive {:EXIT, ^pid, {:timeout, {Task, task_fn, [_task, 100]}}}
+                     when task_fn in [:await, :await_many],
+                     150
+
       Process.flag(:trap_exit, false)
     end
   end
--- a/exercises/concept/rpn-calculator-inspection/test/rpn_calculator_inspection_test.exs
+++ b/exercises/concept/rpn-calculator-inspection/test/rpn_calculator_inspection_test.exs
@@ -346,7 +346,8 @@ defmodule RPNCalculatorInspectionTest do
 
       assert_receive {:EXIT, ^pid, {:timeout, {Task, task_fn, [_task, 100]}}}
                      when task_fn in [:await, :await_many],
-                     150
+                     150,
+                     "expected to receive a timemout message from Task.await/2 or Task.await_many/2"
 
       Process.flag(:trap_exit, false)
     end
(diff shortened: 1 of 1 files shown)


angelikatyborska:
I'm not 100% sure if this is a good change. Using `await_many` doesn't fulfill [the instructions for this step](https://github.com/exercism/elixir/blob/main/exercises/concept/rpn-calculator-inspection/.docs/instructions.md#4-run-a-concurrent-correctness-check-for-many-inputs):

> Then, for every task started this way, it should await its results for 100ms.

To me this sentence states that each task needs to be awaited separately, not all together. If we were to allow both `await` and `await_many`, we would need to rephrase the instructions to say that either approach can be used, which I would argue will make it more confusing for newcomers. @jiegillet what do you think?

jiegillet:
> To me this sentence states that each task needs to be awaited separately, not all together. If we were to allow both await and await_many, we would need to rephrase the instructions to say that either approach can be used, which I would argue will make it more confusing for newcomers. @jiegillet what do you think?

On one hand, you're right that the instructions really point to using `await`, and I think that's the best doe newcomers.

On the other hand, being a little more permissive doesn't really mean that we have to rephrase the instructions, people should be allowed to come up with alternate solutions if they are pretty much equivalent.

I'm not super familiar with concurrency, am I right in saying that they are equivalent? My intuition is that `await_many` will only take 100ms max, but `await` could be a bit longer if `Enum.map` takes a while to go through the list of tasks?
If that's right, `await_many ` is even stricter so I don't really have an issue with it...

angelikatyborska:
The difference between the two is explained in [`concepts/tasks/about.md`](https://github.com/exercism/elixir/blob/main/concepts/tasks/about.md#timeouts).

TL;DR: `await` is synchronous, it blocks and resets the timeout for each task. It will take max `N * timeout` where N is the number of calls to `await`. If all tasks get started first, and only then awaited, the last task gets a much longer allowed runtime then the first one. `await_many` works for all tasks at once, it will take max `timeout` regardless of how many tasks are awaited.

If you think we don't need to modify the instructions, I'm fine with letting this change be merged.

@madlep can you run `mix format` so that the CI checks pass? 🙂

madlep:
@angelikatyborska that makes sense. Looking at it that way, I can see the intent is that we're wanting to check each task individually only takes 100ms, rather than globally checking if all of them completed in 100ms.

So I get the intent of testing for that.

On the other hand though, it may be difficult for people learning Elixir to understand the difference, and what the exercise is looking for. I know Elixir pretty well, and was playing around with the exercises for fun, but it still took me quite a bit of effort to understand the failing test, and what I needed to do to fix it up. So the motivation for the change is making it more flexible and less confusing to move through the exercise.

So added an expectation failure message, which hopefully describes it a bit better.

(also ran `mix format` as requested 👍 )

angelikatyborska (approved):
Thank you 💜

Issue #1027, opened 2022-01-10
[Recurring] Keep exercises up to date with problem specifications


angelikatyborska:
This is a recurring task that has to be done from time to time. **Do not reference this issue by writing `resolves/fixes #XXX` in your PR** because that will close it.

We want to keep the track up to date with [problem specifications](https://github.com/exercism/problem-specifications), which means regularly checking if anything changes and needs updating.

How to do this task:

## Step 1: fetch configlet

Run:
```
./bin/fetch_configlet.sh
```

## Step 2: check which exercises need updating and in which way

Run:
```
./bin/configlet sync 
```

The output will list all Elixir exercises that are out of sync in some way. Example output:
```
[warn] docs: instructions unsynced: affine-cipher
[warn] docs: instructions unsynced: all-your-base

[warn] metadata: unsynced: acronym
[warn] metadata: unsynced: armstrong-numbers

[warn] collatz-conjecture: missing 2 test cases
       - zero is an error (2187673d-77d6-4543-975e-66df6c50e2da)
       - negative value is an error (ec11f479-56bc-47fd-a434-bcd7a31a7a2e)
```

Choose one or more problems and attempt to fix it. Metadata and doc updates can be all done together for many exercises in a single PR, but please create separate PRs per exercise when changing tests.

### Step 3.1: Sync all docs

Run this command to update all exercise docs (introductions):

```
 ./bin/configlet sync --docs --update
```

Press `y` when asked `sync the above docs ([y]es/[n]o)?`.

### Step 3.2: Sync all metadata

Run this command to update all exercise metadata:

```
 ./bin/configlet sync --metadata --update
```

Press `y` when asked `sync the above docs ([y]es/[n]o)?`.

### Step 3.3: Update tests for an exercise

The general goal is that all Elixir practice exercises follow problem specifications, except for when it doesn't make any sense in Elixir.

Important:
- Get familiar with the [problem specifications](https://github.com/exercism/problem-specifications) repository and the `canonical-data.json` format it uses for describing tests cases. 
- Read `CONTRIBUTING.md` for the general rules of this repository.
- Read `exercises/practice/<exercise>/.meta/design.md` if it exists to learn about design decision that we took in the past for this exercise (e.g. to drop a required function from `resistor-color`). Create this file if you're making unexpected decisions yourself!
- Please create separate a PR per exercise when changing tests.

### Step 3.3.1 Update `tests.toml`

Update `tests.toml` by running this command for a chosen exercise:
```
[...]

PR #1036, opened 2022-01-15
Update `rational-numbers` tests

neduard:
From #1027 

I've also updated docs.  Not sure if this is OK for this PR or if I should restrict it to just updating the tests.


angelikatyborska (approved):
Perfect, thank you! 💜 

Yes, it's fine to throw in a docs update in the same PR, those changes are very easy to review 🙂

PR #1038, opened 2022-01-21
Sync collatz-conjecture

angelikatyborska:
The only difference in the reimplemented test cases is the word "numbers" changed to "integers". We don't use those error messages in the test, so I think we can just assume we "implement" the new test cases without actually changing anything 😁


jiegillet (approved):
Haha, so much work, so much work! We deserve a 20% raise!

PR #1039, opened 2022-01-22
Sync protein-translation

angelikatyborska:
This one was a bit more challenging because Elixir's implementation has two separate functions, `of_codon` and `of_rna`, but problem specs only define a single function. For this reason I took some liberties by duplicating some test cases and changing their order.

The 7 new cases can be split into 3 categories:
- 1 test for empty input: https://github.com/exercism/problem-specifications/blob/main/exercises/protein-translation/canonical-data.json#L3-L16
- 2 tests for two cases where two identical proteins are returned: https://github.com/exercism/problem-specifications/blob/main/exercises/protein-translation/canonical-data.json#L170-L190
- 4 tests for input validation: https://github.com/exercism/problem-specifications/blob/main/exercises/protein-translation/canonical-data.json#L245-L312

Before that, there was no input validation defined in problem specs, but we already had some.

### DIFF
--- a/exercises/practice/protein-translation/.meta/tests.toml
+++ b/exercises/practice/protein-translation/.meta/tests.toml
@@ -1,6 +1,16 @@
-# This is an auto-generated file. Regular comments will be removed when this
-# file is regenerated. Regenerating will not touch any manually added keys,
-# so comments can be added in a "comment" key.
+# This is an auto-generated file.
+#
+# Regenerating this file via `configlet sync` will:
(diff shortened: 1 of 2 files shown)


jiegillet (approved):
You're right, this one is a bit more tricky. It makes sense to have two separate functions but it makes comparing tests with canonical data a bit complicated. I'm good with your changes though!

I must also say that the example solution, while cool with the meta-programming stuff, is actually kind of hard to read with the rna and codon stuff mixed up.

PR #1040, opened 2022-01-22
Sync matching-brackets

angelikatyborska:
Two new test cases: https://github.com/exercism/problem-specifications/blob/main/exercises/matching-brackets/canonical-data.json#L139-L156


jiegillet (approved):
A shut-an-open PR!

PR #1043, opened 2022-01-28
Protein-translation: fix warning in tests

mstranger:
Using `strand` variable in function


jiegillet (approved):
Nice catch! Thank you.

PR #1044, opened 2022-01-29
Sync: Adding one new test case to `isbn_verifier`

jiegillet:
Straightforward one.

### DIFF
--- a/exercises/practice/isbn-verifier/.meta/tests.toml
+++ b/exercises/practice/isbn-verifier/.meta/tests.toml
@@ -1,21 +1,28 @@
-# This is an auto-generated file. Regular comments will be removed when this
-# file is regenerated. Regenerating will not touch any manually added keys,
-# so comments can be added in a "comment" key.
+# This is an auto-generated file.
+#
+# Regenerating this file via `configlet sync` will:
+# - Recreate every `description` key/value pair
+# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
+# - Remove any `include = true` key/value pair (an omitted `include` key implies inclusion)
+# - Preserve any other key/value pair
+#
+# As user-added comments (using the # character) will be removed when this file
+# is regenerated, comments can be added via a `comment` key.
 
 [0caa3eac-d2e3-4c29-8df8-b188bc8c9292]
-description = "valid isbn number"
+description = "valid isbn"
 
 [19f76b53-7c24-45f8-87b8-4604d0ccd248]
 description = "invalid isbn check digit"
 
 [4164bfee-fb0a-4a1c-9f70-64c6a1903dcd]
-description = "valid isbn number with a check digit of 10"
+description = "valid isbn with a check digit of 10"
 
 [3ed50db1-8982-4423-a993-93174a20825c]
 description = "check digit is a character other than X"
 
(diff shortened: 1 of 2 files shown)


exercises/practice/isbn-verifier/.meta/tests.toml
```diff
@@ -1,21 +1,28 @@
...
 [0caa3eac-d2e3-4c29-8df8-b188bc8c9292]
-description = "valid isbn number"
+description = "valid isbn"
 
 [19f76b53-7c24-45f8-87b8-4604d0ccd248]
 description = "invalid isbn check digit"
 
 [4164bfee-fb0a-4a1c-9f70-64c6a1903dcd]
-description = "valid isbn number with a check digit of 10"
+description = "valid isbn with a check digit of 10"
```
kotp, line 19: This line should not be changed, as the change would make it say "international standard book number number".

exercises/practice/isbn-verifier/.meta/tests.toml
```diff
@@ -48,7 +55,10 @@ description = "empty isbn"
 description = "input is 9 characters"
 
 [ed6e8d1b-382c-4081-8326-8b772c581fec]
-description = "invalid characters are not ignored"
+description = "invalid characters are not ignored after checking length"
```
kotp, line 58: Is it important that it is not ignored before or after checking the length?   Or is it only important that the valid ISBN's are coming back as vaildated?  This seems like it is a test that will suggest a specific order of solution, rather than only verify a solution.

kotp: Indeed, some solutions may not only not ignore the invalid characters, but may seek them out and strip them.  I am not sure this is a good test as described now.

kotp (changes requested):

exercises/practice/isbn-verifier/.meta/tests.toml
jiegillet: This change is the opposite, it's removing `number`.

exercises/practice/isbn-verifier/.meta/tests.toml
jiegillet: This change and the other (and all the others from the "Sync" batch of PRs) was done by `configlet sync`. Everything comes from `problem-specifications`, if you want to make another change to needs to be from there.

kotp: I thought at first, until I saw the title change, then I figured it was a suggested change.

It is one that I do not think I would have my track opt into.  But it may be important for Elixir.

kotp: I  can not cancel my review for some reason (the request to change... looking into it now.)

kotp: This comes from [problem-specifications](https://github.com/exercism/problem-specifications) (link provided) if someone else would like to go there to check it out with a click.

exercises/practice/isbn-verifier/.meta/tests.toml
kotp: Thanks.  Yeah, I saw it removed in October by ee7.  Misread it here.

kotp (approved):

neenjaw (approved):

PR #1045, opened 2022-01-29
Sync: Add one test case to `bowling`

jiegillet:
I'm on a roll!

### DIFF
--- a/exercises/practice/bowling/test/bowling_test.exs
+++ b/exercises/practice/bowling/test/bowling_test.exs
@@ -104,7 +104,7 @@ defmodule BowlingTest do
   end
 
   @tag :pending
-  test "efb426ec-7e15-42e6-9b96-b4fca3ec2359" do
+  test "last two strikes followed by only last bonus with non strike points" do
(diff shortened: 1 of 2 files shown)


exercises/practice/bowling/test/bowling_test.exs
```diff
@@ -103,6 +103,14 @@ defmodule BowlingTest do
     assert Bowling.score(game) == {:ok, 30}
   end
 
+  @tag :pending
+  test "efb426ec-7e15-42e6-9b96-b4fca3ec2359" do
```
neenjaw, line 107: Can this be updated with a more descriptive title?

neenjaw (changes requested):
One small request

exercises/practice/bowling/test/bowling_test.exs
jiegillet:
```suggestion
  test "last two strikes followed by only last bonus with non strike points" do
```
Oops , copy/paste error.

neenjaw (approved):

PR #1046, opened 2022-01-29
Sync: Update test in `largest-series-product`

jiegillet:
This one is tiny, it was a change in the error message, but we don't return specific errors in this one.


neenjaw (commented):
Is the added test case in the toml file already covered?

neenjaw (approved):

PR #1047, opened 2022-01-29
Sync: Add error handling to `space-age`

jiegillet:
This one is a breaking change, because the new test required an error message, so I changed the output to `{:ok, answer} | {:error, message}`.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1059,6 +1059,7 @@
         "prerequisites": [
           "floating-point-numbers",
           "atoms",
+          "tuples",
           "multiple-clause-functions",
           "pattern-matching",
           "maps",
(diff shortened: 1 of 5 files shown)


neenjaw:
Why did we add error messages here and not for largest-series-product? Does having to return an error tuple change where this presently falls in the track?

neenjaw (commented):
Why did we add error messages here and not for largest-series-product? Does having to return an error tuple change where this presently falls in the track?

jiegillet:
The reason why I didn't touch the implementation in largest-series-product is that this was  already implemented:
```elixir
      unless Enum.member?(Range.new(0, String.length(number_string)), size) do
        raise ArgumentError
      end
```
It's not necessarily what I would have written, but it was internally consistent, so I didn't touch it.

> Does having to return an error tuple change where this presently falls in the track?
I doubt it, because it requires more complex stuff, but I've added `tuples` to the config just in case. Thank you.

neenjaw (approved):

PR #1050, opened 2022-01-30
Sync: re-write `grade-school`

jiegillet:
This one was a bit involved. The requirements and API changed quite a lot, so I decided to re-write the whole test suite. Here are the main changes I made

- The data structure for the school used to be a `map`, but I changed it so that the student may implement whatever `@type school` they prefer, and the tests don't assume a specific data structure. 
- This mean I had to add `School.new/0`.
- `add` used to return a `school`, but the new tests needed to know if a student insertion was allowed, so now it returns `{:ok | :error, school}`.
- This means I had to add atoms and tuples to the prerequisites.
- I joined some tests that checked the results (list of `:ok | :error`) and the roster separately for the same data.

### DIFF
--- a/exercises/practice/grade-school/lib/school.ex
+++ b/exercises/practice/grade-school/lib/school.ex
@@ -5,7 +5,7 @@ defmodule School do
   Each student is in a grade.
   """
 
-  @type school :: :dummy_type
+  @type school :: any()
 
   @doc """
   Create a new, empty school.
--- a/exercises/practice/grade-school/lib/school.ex
+++ b/exercises/practice/grade-school/lib/school.ex
@@ -5,24 +5,33 @@ defmodule School do
   Each student is in a grade.
   """
 
+  @type school :: :dummy_type
+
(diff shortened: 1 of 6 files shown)


exercises/practice/grade-school/lib/school.ex
```diff
@@ -5,24 +5,33 @@ defmodule School do
   Each student is in a grade.
   """
 
+  @type school :: :dummy_type
```
neenjaw, line 8: Is this a convention we have established anywhere else?

I don't recall it myself and this might not be intuitive to student approaching the problem.

So I think 3 options:
- initial type is `any()`
- specify the underlying data structure
- don't use a defined type and just use `any()` in the typespec

jiegillet: I think I have seen this but I can't find it, maybe it was a different track...
First option sounds good.

neenjaw (approved):

PR #1051, opened 2022-01-31
Update space-age tests.toml

angelikatyborska:
The PR https://github.com/exercism/elixir/pull/1047/files added a new tests case but didn't add it to `tests.toml` 🙂


jiegillet (approved):
Dang it! Thanks for having my back!

PR #1055, opened 2022-02-01
Fix arity typo in the city-office exercise instructions.

veimis:
Fix arity of Form.check_length -function in city-office exercise instructions.


angelikatyborska (approved):
Thank you!

PR #1058, opened 2022-02-04
Sync: add a new test to `rational-numbers`

mstranger:
Additional test for `abs` function and update docs/metadata.


jiegillet (approved):
Sweet! Thank you for also syncing the docs!

PR #1059, opened 2022-02-05
Update instructions.md

rikroos:
The unit test checks in case of `HighScore.update_score/3` if the player is added in case of not existing before. This should be emphasized in the instruction.

### DIFF
--- a/exercises/concept/high-score/.docs/instructions.md
+++ b/exercises/concept/high-score/.docs/instructions.md
@@ -68,7 +68,9 @@ To update a player's score by adding to the previous score, define `HighScore.up
 
 - The first argument is the map of scores.
 - The second argument is the name of the player as a string, whose score you wish to update.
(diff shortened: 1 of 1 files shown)


exercises/concept/high-score/.docs/instructions.md
```diff
@@ -68,7 +68,7 @@ To update a player's score by adding to the previous score, define `HighScore.up
 
 - The first argument is the map of scores.
 - The second argument is the name of the player as a string, whose score you wish to update.
-- The third argument is the score that you wish to **add** to the stored high score.
+- The third argument is the score that you wish to **add** to the stored high score. In case the name of the player does not exist the name will be added to the map of scores with the score argument as the default score.
```
angelikatyborska, line 71: Thank you for the feedback! Let's add that clarification. But I'm going to rephrase this a bit so that the wording matches the rest of the instructions a bit better.

```suggestion
- The third argument is the score that you wish to **add** to the stored high score.

The function should also work if the player doesn't have a previous score - assume the previous score is 0.
```

angelikatyborska (approved):

PR #1063, opened 2022-02-15
Fix wording

tomek-ch:
I think one of the sentences in the introduction is missing a verb which makes it a little confusing to read

### DIFF
--- a/concepts/docs/introduction.md
+++ b/concepts/docs/introduction.md
@@ -2,7 +2,7 @@
 
 Documentation in Elixir is a first-class citizen.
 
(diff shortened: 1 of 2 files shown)


jiegillet:
Thank you, that's a great fix!
This introduction is duplicated somewhere else in the `concept` folder in at least one file, maybe two. 
Could you do a global search to make sure the old sentence doesn't remain?

tomek-ch:
> Thank you, that's a great fix! This introduction is duplicated somewhere else in the `concept` folder in at least one file, maybe two. Could you do a global search to make sure the old sentence doesn't remain?

Thanks, fixed it

jiegillet (approved):
Perfect, thank you!

PR #1064, opened 2022-02-16
Update boutique-inventory test cases

kasperkronborg:
Quick disclaimer: My first PR for this repo, so all feedback is very welcome.

### Description

This PR addresses two small things for the `boutique-inventory` exercise.
- Updates the test case for the first test `works for an empty quantity map` for `task_id: 3` to be more aligned with what in the exercise description. See exercise quote below.
- Updates a test description to be more aligned with the former test with the same `task_id: 3`

> Implement the `increase_quantity/2` function. It should take a single item and a number `n`, and return that item with the quantity for each size increased by `n`.

This PR came about as I was doing the exercise and didn't fully comprehend the test. I believe this is what the test was originally meant to assert. I haven't gone about the practicalities of adding myself to any contribution lists etc.

### DIFF
--- a/exercises/concept/boutique-inventory/test/boutique_inventory_test.exs
+++ b/exercises/concept/boutique-inventory/test/boutique_inventory_test.exs
@@ -75,7 +75,7 @@ defmodule BoutiqueInventoryTest do
     end
 
     @tag task_id: 3
-    test "increases quantity of an item" do
+    test "works for an item with an existing quantity map" do
       assert BoutiqueInventory.increase_quantity(
                %{
                  name: "Green Swimming Shorts",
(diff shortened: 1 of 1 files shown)


angelikatyborska:
Hi @kasperkronborg! Thank you for the PR, but I'm afraid your changes to the tests aren't correct. The function in question is not supposed to _add_ any sizes to the map. It should only increase the counts for sizes that already exist in the map. Is there another change that you think we should do to the test descriptions to make it more clear? Maybe the wording "works with an empty map" is not the best 🤔.

kasperkronborg:
Hi @angelikatyborska! Roger that. That wasn't at all what I got from the exercise description. But good to know that all is as it should be. I will close again 😊

PR #1065, opened 2022-02-17
Make types more descriptive

hauleth:
When I was trying to do this task the most irritating thing was that the description is underspecified and it was tripping me over and over again on simple things that should be either in the docstrings or in types. For example information that what are possible errors returned by the given functions.

### DIFF
--- a/exercises/practice/robot-simulator/lib/robot_simulator.ex
+++ b/exercises/practice/robot-simulator/lib/robot_simulator.ex
@@ -1,15 +1,14 @@
 defmodule RobotSimulator do
   @type robot() :: any()
   @type direction() :: :north | :east | :south | :west
-  @type position :: {integer(), integer()}
+  @type position() :: {integer(), integer()}
 
   @doc """
   Create a Robot Simulator given an initial direction and position.
 
   Valid directions are: `:north`, `:east`, `:south`, `:west`
   """
-  @spec create(direction, position) ::
-    robot() | {:error, "invalid direction" | "invalid position"}
+  @spec create(direction, position) :: robot() | {:error, String.t()}
   def create(direction \\ nil, position \\ nil) do
   end
 
@@ -18,8 +17,7 @@ defmodule RobotSimulator do
 
   Valid instructions are: "R" (turn right), "L", (turn left), and "A" (advance)
   """
-  @spec simulate(robot, instructions :: String.t()) ::
-    robot() | {:error, "invalid instruction"}
+  @spec simulate(robot, instructions :: String.t()) :: robot() | {:error, String.t()}
   def simulate(robot, instructions) do
   end
 
@@ -28,14 +26,14 @@ defmodule RobotSimulator do
 
   Valid directions are: `:north`, `:east`, `:south`, `:west`
   """
-  @spec direction(robot) :: direction
+  @spec direction(robot) :: direction()
   def direction(robot) do
   end
 
   @doc """
   Return the robot's position.
   """
-  @spec position(robot) :: position
+  @spec position(robot) :: position()
   def position(robot) do
   end
 end
--- a/exercises/practice/robot-simulator/lib/robot_simulator.ex
+++ b/exercises/practice/robot-simulator/lib/robot_simulator.ex
@@ -1,10 +1,15 @@
 defmodule RobotSimulator do
+  @type robot() :: any()
+  @type direction() :: :north | :east | :south | :west
+  @type position :: {integer(), integer()}
+
   @doc """
   Create a Robot Simulator given an initial direction and position.
 
   Valid directions are: `:north`, `:east`, `:south`, `:west`
   """
-  @spec create(direction :: atom, position :: {integer, integer}) :: any
+  @spec create(direction, position) ::
+    robot() | {:error, "invalid direction" | "invalid position"}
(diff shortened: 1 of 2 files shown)


exercises/practice/robot-simulator/lib/robot_simulator.ex
```diff
@@ -1,10 +1,15 @@
 defmodule RobotSimulator do
+  @type robot() :: any()
+  @type direction() :: :north | :east | :south | :west
+  @type position :: {integer(), integer()}
```
jiegillet, line 4: Can you add brackets to `position` just for consistency?

exercises/practice/robot-simulator/lib/robot_simulator.ex
```diff
@@ -1,10 +1,15 @@
...
+
   @doc """
   Create a Robot Simulator given an initial direction and position.
 
   Valid directions are: `:north`, `:east`, `:south`, `:west`
   """
-  @spec create(direction :: atom, position :: {integer, integer}) :: any
+  @spec create(direction, position) ::
+    robot() | {:error, "invalid direction" | "invalid position"}
```
jiegillet, line 12: I know this is more descriptive but I wouldn't spell out the individual errors, using `String.t()` would be fine. Same for `simulate` below.

jiegillet (commented):
Thank you for that PR, it certainly is an important topic.

Types are tricky, because we want to leave them open enough so that students can change them to suit their need. Your suggestions are pretty conservative, so that's not too much of a concern for me. I particularly appreciate adding `| {:error, String.t()}`.

You would also need change the example implementation in `.meta` so that the the types match.

@angelikatyborska what do you think?

exercises/practice/robot-simulator/lib/robot_simulator.ex
hauleth: The point is that without spelling of individual errors it is still underspecified. I need to run tests to understand what task expects out of me, as the tests check for **fixed** strings. If it would simply check for `{:error, _}` then it could specify `{:error, String.t()}`. The whole PR is just to state this strings there.

angelikatyborska: I'll take a closer look in a day or two because I'm pretty busy but I just want to say:
>  I need to run tests to understand what task expects out of me

This is the idea behind every single practice exercise on Exercism

neenjaw: I think spelling out the individual error string codes is not typical of Elixir code that I have reviewed in the past.  I would agree with @jiegillet's suggestion to keep it as `String.t()`.


Agree with @angelikatyborska, this is one of the limitations of using a central repository for the problem descriptions that all language tracks suffer from. -- the solution requirements for a particular language are not clearly always spelled out in the problem description.  However, Exercism also encourages TDD in practice tests, so you are reading the tests and encountering those error strings as you progress through the test cases.

hauleth: > I think spelling out the individual error string codes is not typical of Elixir code that I have reviewed in the past. I would agree with @jiegillet's suggestion to keep it as `String.t()`.

Using strings as error values is not common thing in Elixir. And when you use atoms then you often have list of them available somewhere.

Additionally Elixir approach is to return `{:ok, term()} | {:error, term()}`, not `term() | {:error, term()}`. So if the specs for the task would say `{:ok, term()} | {:error, term()}` it would be more clear that the task is to write code in a way that it should handle the errors in meaningful way. With current description I just assumed that the caller will always will be well-behaved and in case of spec mismatch it should fail. Because type explicitly says that the position will be tuple of 2 integers, I just assume that it will never mismatch the specs, as that is counterintuitive. The [tests does that](https://github.com/exercism/elixir/blob/40227b8141577eadcbda120f10128abbf6725b7f/exercises/practice/robot-simulator/test/robot_simulator_test.exs#L45) and it is confusing and frustrating.

---

IMHO the task description should be enough for writing the proper solution on the first try for experienced developer. That is frustrating if I need to look around for external sources of the information if I want to finish a task.

neenjaw: I'm not disagreeing with your suggestion to add the `direction()` type to give context to the atom. Sure, it's a good change.

With regard to your addendum, sure that is an approach to use `{:ok, term()} | {:error, term()}`, and I would be open to that option as well -- but that's not the change you proposed in your PR so I'm sorry if I misunderstood your intention. If @angelikatyborska and/or @jiegillet are open to that change, let's do it if it makes it better and eases frustration.  Likely you are not the only one experiencing this.

hauleth: So my suggestion is one of:

- remove the `spec`s and force user to read tests, as current `spec` is misleading
- remove tests that checks for error values, as these are against the `spec`

angelikatyborska:
Today I learned that putting string or charlist literals in typespecs in not even allowed. I got a hint from my IDE's plugin 😁 
Screenshot 2022-03-19 at 09 08 01

To get this PR merged, I changed the string literals to `String.t` because that's a less invasive change to people that already solved the exercise.

angelikatyborska (approved):
Thank you, the typespecs for this exercise are now much more helpful!

PR #1067, opened 2022-02-22
rpn-calculator-inspection: small wording improvement

glennj:
Very small improvement to the wording: 
from
```
The second argument is a map that serves an accumulator ...
```
to
```
The second argument is a map that serves as an accumulator ...
                                         --
```


angelikatyborska (approved):
Of course! Thank you.

PR #1076, opened 2022-03-06
New concept exercise: `genserver`

angelikatyborska:
Heavily WIP!

I'm opening this PR to signal that I'm working on this, but it's not ready for a first review yet.

### DIFF
--- a/exercises/concept/take-a-number-deluxe/.docs/instructions.md
+++ b/exercises/concept/take-a-number-deluxe/.docs/instructions.md
@@ -1,0 +1,8 @@
+# Instructions
+
+The simple Take-A-Number machine was selling really well, but some advanced users were complaining about a lack of features.
+
+That's why the manufacturer decided to release a deluxe model with more features, and you once again have to write the software for this machine.
+
+The deluxe model give out consecutive numbers and report what was the last number given out just like the previous simple model, but also...
+
--- a/exercises/concept/take-a-number-deluxe/.docs/introduction.md
+++ b/exercises/concept/take-a-number-deluxe/.docs/introduction.md
@@ -1,3 +1,8 @@
 # Introduction
 
 ## GenServer
+
+[impl]: https://hexdocs.pm/elixir/Module.html#module-impl
+[genserver]: https://hexdocs.pm/elixir/GenServer.html
+[start_link]: https://hexdocs.pm/elixir/GenServer.html#start_link/3
+[handle_call]: https://hexdocs.pm/elixir/GenServer.html#c:handle_call/3
--- a/exercises/concept/take-a-number-deluxe/.docs/introduction.md
+++ b/exercises/concept/take-a-number-deluxe/.docs/introduction.md
@@ -1,0 +1,3 @@
+# Introduction
+
+## GenServer
--- a/exercises/concept/take-a-number-deluxe/.meta/design.md
+++ b/exercises/concept/take-a-number-deluxe/.meta/design.md
@@ -1,0 +1,23 @@
+# Design
+
+## Learning objectives
+
+- Know what GenServer is
+- Use all three kinds of callbacks - handle call/cast/info
+
+## Out of scope
+
+## Concepts
+
+- `genserver`
+
+## Prerequisites
+
+- `processes`
+- `pids`
+- `agent`
+
+## Analyzer
+
+- Make sure that `GenServer` is used.
+- Make sure that `@impl` is used.
--- a/exercises/concept/take-a-number-deluxe/.meta/design.md
+++ b/exercises/concept/take-a-number-deluxe/.meta/design.md
@@ -16,6 +16,9 @@
 - `processes`
 - `pids`
 - `agent`
+- `behaviours`
+- `struct`
+- `tuples`
 
 ## Analyzer
 
--- a/exercises/concept/take-a-number-deluxe/.meta/exemplar.ex
+++ b/exercises/concept/take-a-number-deluxe/.meta/exemplar.ex
@@ -1,0 +1,3 @@
+defmodule TakeANumberDeluxe do
+
+end
--- a/exercises/concept/take-a-number-deluxe/.meta/exemplar.ex
+++ b/exercises/concept/take-a-number-deluxe/.meta/exemplar.ex
@@ -1,3 +1,2 @@
 defmodule TakeANumberDeluxe do
-
 end
--- a/exercises/concept/take-a-number-deluxe/lib/take_a_number_deluxe/state.ex
+++ b/exercises/concept/take-a-number-deluxe/lib/take_a_number_deluxe/state.ex
@@ -1,0 +1,62 @@
+defmodule TakeANumberDeluxe.State do
+  defstruct min_number: 1, max_number: 999, queue: [], serving: [], served_count: 0
+
+  def new(min_number, max_number) do
+    if is_integer(min_number) and is_integer(max_number) and min_number < max_number do
+      {:ok, %__MODULE__{min_number: min_number, max_number: max_number}}
+    else
+      {:error, :invalid_configuration}
+    end
+  end
+
+  def queue_new_number(%__MODULE__{} = state) do
+    case find_next_available_number(state) do
+      {:ok, next_available_number} ->
+        {:ok, next_available_number, %{state | queue: [next_available_number | state.queue]}}
+
+      {:error, error} ->
+        {:error, error}
+    end
+  end
+
+  def serve_next_queued_number(%__MODULE__{} = state) do
+    {next_queued_number, new_queue} = List.pop_at(state.queue, length(state.queue) - 1)
+    {:ok, %{state | queue: new_queue, serving: [next_queued_number | state.serving]}}
+  end
+
+  def mark_number_as_served(%__MODULE__{} = state, number) do
+    index = Enum.find_index(state.serving, &(&1 == number))
+
+    if index == nil do
+      {:error, :number_not_found}
+    else
+      {_, new_serving} = List.pop_at(state.serving, index)
+      {:ok, %{state | serving: new_serving, served_count: state.served_count + 1}}
+    end
+  end
+
+  defp find_next_available_number(state) do
+    all_numbers_in_use = state.queue ++ state.serving
+    all_numbers = Enum.to_list(state.min_number..state.max_number)
+
+    case all_numbers_in_use do
+      [] ->
+        {:ok, state.min_number}
+
+      list when length(list) == length(all_numbers) ->
+        {:error, :all_possible_numbers_are_in_use}
+
+      _ ->
+        current_highest_number = Enum.max(all_numbers_in_use)
+
+        next_available_number =
+          if current_highest_number < state.max_number do
+            current_highest_number + 1
+          else
+            Enum.min(all_numbers -- all_numbers_in_use)
+          end
+
+        {:ok, next_available_number}
+    end
+  end
+end
--- a/exercises/concept/take-a-number-deluxe/test/take_a_number_deluxe_test.exs
+++ b/exercises/concept/take-a-number-deluxe/test/take_a_number_deluxe_test.exs
@@ -1,0 +1,7 @@
+defmodule TakeANumberDeluxeTest do
+  use ExUnit.Case
+
+  @tag task_id: 1
+  test "..." do
+  end
+end
--- a/exercises/concept/take-a-number-deluxe/.docs/hints.md
+++ b/exercises/concept/take-a-number-deluxe/.docs/hints.md
@@ -1,0 +1,3 @@
+# Hints
+
+## General
--- a/exercises/concept/take-a-number-deluxe/.formatter.exs
+++ b/exercises/concept/take-a-number-deluxe/.formatter.exs
@@ -1,0 +1,4 @@
+# Used by "mix format"
+[
+  inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}"]
+]
--- a/exercises/concept/take-a-number-deluxe/.gitignore
+++ b/exercises/concept/take-a-number-deluxe/.gitignore
@@ -1,0 +1,24 @@
+# The directory Mix will write compiled artifacts to.
+/_build/
+
+# If you run "mix test --cover", coverage assets end up here.
+/cover/
+
+# The directory Mix downloads your dependencies sources to.
+/deps/
+
+# Where third-party dependencies like ExDoc output generated docs.
+/doc/
+
+# Ignore .fetch files in case you like to edit your project deps locally.
+/.fetch
+
+# If the VM crashes, it generates a dump, let's ignore it too.
+erl_crash.dump
+
+# Also ignore archive artifacts (built via "mix archive.build").
+*.ez
+
+# Ignore package tarball (built via "mix hex.build").
+nil-*.tar
+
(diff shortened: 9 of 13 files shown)


angelikatyborska:
The tests, boilerplate, and exemplar solution are ready for review. I want to get those approved before I continue with detailed instructions and hints.

It was really hard to find something that:
- makes sense in the real world,
- uses all three `handle_call`, `handle_cast`, and `handle_info`, and uses `handle_call` both without user input and with user input.

Overall the exemplar solution is pretty big and this task will require more work than usual, but I don't see a way of making it smaller without compromising on the learning goals.

exercises/concept/take-a-number-deluxe/.meta/design.md
```diff
@@ -0,0 +1,41 @@
...
+## Concepts
+
+- `genserver`
+
+## Prerequisites
+
+- `processes`
+- `pids`
+- `agent`
+- `behaviours`
```
angelikatyborska, line 22: Because GenServer is a behaviour that needs to be implemented, and because its complicated in itself, I think it would be better to teach behaviours separately first. I'll try to think of a story soon.

exercises/concept/take-a-number-deluxe/.meta/exemplar.ex
```diff
@@ -0,0 +1,82 @@
...
+  @impl true
+  def handle_info({:take_a_number, sender_pid}, state) do
+    send(sender_pid, {:error, :basic_model_message_received_by_deluxe_model_server})
+    {:noreply, state}
+  end
+
+  @impl true
+  def handle_info(_, state) do
+    {:noreply, state}
+  end
```
angelikatyborska, line 81: Running `use GenServer` already creates a catch-all `handle_info`, so I had to get creative to push students to implement it themselves and actually be able to assert they did. At first I wanted just a normal catch-all that calls `IO.inspect` but I couldn't figure out how to assert on the IO. The normal capture IO in ExUnit doesn't work because the IO happens in a different process.

So I came up with this, and I'm going to justify it with something like: "when testing the initial prototype of the new machine, somebody did something very weird in the client code and ended up sending the old messages meant for the basic Take-A-Number machine to the new GenServer based one instead. They were really annoyed that they didn't receive any responses or any errors, and asked you to add this response for them so they realize what's going on the next time". A bit far-fetched but I was out of ideas.

jiegillet: If you want another idea for justifying `handle_info`, elixir in Action uses the example of sending messages at regular intervals
```elixir
def init(_) do
  :timer.send_interval(5000, :cleanup)
  ...
end
```
You could say that the number queue should be reset once a day.

exercises/concept/take-a-number-deluxe/.docs/instructions.md
```diff
@@ -0,0 +1,10 @@
+# Instructions
+
+The basic Take-A-Number machine was selling really well, but some users were complaining about its lack of advanced features compared to other models available on the market.
+
+The manufacturer listened to user feedback and decided to release a deluxe model with more features, and you once again have were tasked with writing the software for this machine.
```
jiegillet, line 5:
```suggestion
The manufacturer listened to user feedback and decided to release a deluxe model with more features, and you once again were tasked with writing the software for this machine.
```

exercises/concept/take-a-number-deluxe/.meta/exemplar.ex
```diff
@@ -0,0 +1,82 @@
...
+    GenServer.call(machine, {:serve_next_queued_number, priority_number})
+  end
+
+  def reset_state(machine) do
+    GenServer.cast(machine, :reset_state)
+  end
+
+  # Server callbacks
+
+  @impl true
```
jiegillet, line 28: I know it's most common to use `@impl true`, but I've also read that using `@impl GenServer` might be better, as it's more explicit. Not sure what's best.
I guess it should mainly be addressed in the behaviour concept.

exercises/concept/take-a-number-deluxe/lib/take_a_number_deluxe/state.ex
```diff
@@ -0,0 +1,63 @@
...
+    case find_next_available_number(state) do
+      {:ok, next_available_number} ->
+        {:ok, next_available_number, %{state | queue: [next_available_number | state.queue]}}
+
+      {:error, error} ->
+        {:error, error}
+    end
+  end
+
+  def serve_next_queued_number(%__MODULE__{} = state, priority_number) do
```
jiegillet, line 22: For this function, the logic is a bit hard to follow from the code alone, I think it's coming from the fact that you are using a list to model a queue.
I have two ideas: either add some documentation, or use :queue. Something like (untested)

```elixir
cond do
  :queue.is_empty(state.queue) ->
    {:error, :empty_queue}

  is_nil(priority_number) ->
    {{:value, next_number}, new_queue} = :queue.out(state.queue)
    {:ok, next_number, %{state | queue: new_queue}}

  :queue.member(priority_number, state.queue) ->
    {:ok, priority_number, %{state | queue: :queue.delete(priority_number, state.queue)}}

  true ->
    {:error, :priority_number_not_found}
end    
```
Documentation is probably simpler :)

jiegillet (commented):
Great work!
I think the exercise makes sense in a "real world" kind of way, and I don't think it will be that hard for students to complete, because all of the heavy lifting is done in `TakeANumberDeluxe.State`, all they have to do it call the right functions and check for errors.

I would suggest adding some `@doc`s to the `TakeANumberDeluxe.State` functions, because it's important to quickly grasp what the functions do to pick the right one, but the implementations are not trivial.

exercises/concept/take-a-number-deluxe/lib/take_a_number_deluxe/state.ex
angelikatyborska: I added specs and used `:queue`. In the instructions, I'll make it clear that those functions can be used as they are.

exercises/concept/take-a-number-deluxe/test/take_a_number_deluxe_test.exs
```diff
@@ -0,0 +1,249 @@
...
+    test "updates the state accordingly" do
+      {:ok, pid} = TakeANumberDeluxe.start_link(min_number: 1, max_number: 99)
+      assert TakeANumberDeluxe.queue_new_number(pid) == {:ok, 1}
+      assert TakeANumberDeluxe.queue_new_number(pid) == {:ok, 2}
+      assert TakeANumberDeluxe.queue_new_number(pid) == {:ok, 3}
+      assert TakeANumberDeluxe.serve_next_queued_number(pid) == {:ok, 1}
+      assert TakeANumberDeluxe.serve_next_queued_number(pid) == {:ok, 2}
+
+      queue =
+        :queue.delete(2, :queue.delete(1, :queue.in(3, :queue.in(2, :queue.in(1, :queue.new())))))
```
angelikatyborska, line 162: This looks ridiculous, it's not equal to `from_list([3])` and the pipe cannot be used 😂

exercises/concept/take-a-number-deluxe/.meta/exemplar.ex
Tyyagoo: Another way to push students to implement the `handle_info` callback is to use GenServer's own [timeout][genserver-timeout]. You can say in history that for "economy" reasons the machine should be turned off after a period of inactivity.

Example:

```elixir
@timeout 1500

def init(_), do: {:ok, ..., @timeout}

def handle_call(..., ..., ...), do: {:reply, ..., ..., @timeout}
def handle_cast(..., ...), do: {:noreply, ..., @timeout}
def handle_info(:timeout, ...), do: exit(:normal) # energy_is_expensive
```

That way, if the student doesn't implement this callback, the process will keep running and the tests should fail, since the _match all_ `handle_info` just calls a `:logger.error`

```elixir
Process.sleep(1000)
assert Process.alive?(pid)
Process.sleep(1000)
refute Process.alive?(pid)
```

_This behavior should be easy to test, but these timers in the tests scares me_

[genserver-timeout]: https://hexdocs.pm/elixir/GenServer.html#module-timeouts

angelikatyborska: @Tyyagoo That sounds brilliant! I totally forgot about the "timeouts" feature. I will try to incorporate your idea.

exercises/concept/take-a-number-deluxe/test/take_a_number_deluxe_test.exs
jiegillet: That's one reason why we code in Elixir and not in Erlang :)

concepts/genserver/about.md
```diff
@@ -0,0 +1,3 @@
+# About
+
+TODO: copy exercise introduction
```
angelikatyborska, line 3: TODO before merge, after exercise introduction is approved.

concepts/genserver/introduction.md
```diff
@@ -0,0 +1,3 @@
+# Introduction
+
+TODO: copy exercise introduction
```
angelikatyborska, line 3: TODO before merge, after exercise introduction is approved.

exercises/concept/take-a-number-deluxe/.docs/introduction.md
```diff
@@ -0,0 +1,147 @@
+# Introduction
+
+## GenServer
+
+~~~~exercism/caution
+This is an advanced topic.
+~~~~
```
angelikatyborska, line 7: I added this as a warning because I'm feeling a bit nervous about how big this exercise ended up being. Do you think this is a good idea?

exercises/concept/take-a-number-deluxe/lib/take_a_number_deluxe/state.ex
```diff
@@ -0,0 +1,85 @@
...
+    cond do
+      :queue.is_empty(state.queue) ->
+        {:error, :empty_queue}
+
+      is_nil(priority_number) ->
+        {{:value, next_number}, new_queue} = :queue.out(state.queue)
+        {:ok, next_number, %{state | queue: new_queue}}
+
+      :queue.member(priority_number, state.queue) ->
+        {:ok, priority_number, %{state | queue: :queue.delete(priority_number, state.queue)}}
```
angelikatyborska, line 46: Looks like `:queue.delete` is only available from OTP 24. @jiegillet what would you propose as a fix? Back to my weird list-based queue? Or add a check with `function_exported?(:queue, :delete, 2)` and use `to_list`, `Enum.filter`, and `from_list` to fake `delete`?

jiegillet: None of these suggestions sparked joy, so I implemented and pushed a queue module. Feel free to revert if it's not to your liking :)
I figured this exercise is already complicated, so this won't make a real difference :)

exercises/concept/take-a-number-deluxe/.docs/introduction.md
jiegillet: Honestly, I don't think it's that advanced. 
I first implemented a `GenServer` on my own in the `bank-account` exercise, and it was very very smooth, the official documentation is very clear for anyone who needs more help.

It may have been complicated to implement `TakeANumberDeluxe.State` but the students don't have to, all they have to do is connect the dots, there is basically no logic to implement.

I think `top-secret` was more complex.

exercises/concept/take-a-number-deluxe/lib/take_a_number_deluxe/state.ex
angelikatyborska: Thank you! I'm glad you didn't make me implement a queue myself 😁. I just added an explanation comment at the top of the queue module.

angelikatyborska:
Something vague is broken with Elixir 1.9 + OTP 20 according to CI:
```

  1) test start_link/1 min and max numbers get validated (TakeANumberDeluxeTest)
     test/take_a_number_deluxe_test.exs:24
     ** (EXIT from #PID<0.122.0>) :invalid_configuration
```

It will take me a while to debug because installing OTP below 22 doesn't work (either out of the box or at all?) on my new mac. I will have to debug via Docker.

angelikatyborska:
I had to build my own Docker image for OTP 20 because the highest Elixir version on DockerHub with OTP 20 is 1.6.

For some reason, `start_link` sends an exit signal in OTP 20 🤷 that shouldn't happen as far as I understand the docs.

We can either add a hack to the test like this:
```elixir
otp_version = String.to_integer(System.otp_release())

if otp_version <= 20, do: Process.flag(:trap_exit, true)

# assert stuff

if otp_version <= 20, do: Process.flag(:trap_exit, false)
```

Or drop OTP 20 support and pair 1.8 and 1.9 with OTP 21. Or try OTP 20.3 instead of 20.0 on CI, hoping that this was a bug that they somehow fixed at some point.

According to https://www.erlang.org/patches/otp-21.0

OTP 21 was first released on 2018-06-19, almost 4 years ago, and OTP 20 was last released on 2020-02-28, 2 years ago.

exercises/concept/take-a-number-deluxe/.docs/instructions.md
```diff
@@ -0,0 +1,128 @@
...
+
+Implement the `report_state/1` function and the necessary `GenServer` callback. The machine should reply to the caller with its current state.
+
+```elixir
+{:ok, machine} = TakeANumberDeluxe.start_link(min_number: 1, max_number: 10)
+TakeANumberDeluxe.report_state(machine)
+# => %TakeANumberDeluxe.State{
+#      max_number: 10,
+#      min_number: 1,
+#      queue: {[], []},
```
jiegillet, line 45:
```suggestion
#      queue: %TakeANumberDeluxe.Queue{in: [], out: []},
```

exercises/concept/take-a-number-deluxe/.docs/introduction.md
```diff
@@ -0,0 +1,143 @@
...
+  def init(_init_arg) do
+    # the initial count of questions asked is always 0
+    state = 0
+    {:ok, state}
+  end
+
+  @impl GenServer
+  def handle_call(:are_we_there_yet?, _from, state) do
+    reply =
+      case state do
```
jiegillet, line 46: `cond` might work better here? 
Not very important at all , feel free to ignore :)

exercises/concept/take-a-number-deluxe/.meta/exemplar.ex
```diff
@@ -0,0 +1,93 @@
+defmodule TakeANumberDeluxe do
+  use GenServer
+
+  # Client API
+
+  @spec start_link(keyword()) :: {:ok, pid()}
```
jiegillet, line 6:
```suggestion
  @spec start_link(keyword()) :: {:ok, pid()} | {:error, any()}
```

jiegillet (approved):
Great work! You had to jump through so many hoops on this one :)

I left a bunch of small comments before we merge.

exercises/concept/take-a-number-deluxe/.docs/introduction.md
angelikatyborska: Yeah, it looks better thanks to no repetition of "n when n..."

angelikatyborska:
@jiegillet Big thanks for the two huge reviews 🙏 I am so happy we will finally fill the GenServer shaped hole in the syllabus...

PR #1078, opened 2022-03-06
Go out of beta for all concept exercises

angelikatyborska:
This practically doesn't change anything, but I think all concept exercises have been around long enough to mark them as "active", not "beta".


jiegillet (approved):
Congratulations on all these active exercises :)

PR #1079, opened 2022-03-07
Fix flakiness test from take-a-number

### LINKED ISSUE #1021: Race condition in `Take-A-Number`
Hi,

There is a problem with `Take-A-Number` exercise, in my opinion cased by async nature of erlang processes.

Here is my code:
```elixir
defmodule TakeANumber do
  def start() do
    spawn(fn -> loop(0) end)
  end

  def loop(state) do
    IO.inspect("INIT INBOX: #{Keyword.get(Process.info(self()), :message_queue_len)}")
    receive do
      {:report_state, sender_pid} ->
        IO.inspect("report_state INBOX: #{Keyword.get(Process.info(self()), :message_queue_len)}")
        send(sender_pid, state)
        loop(state)
      {:take_a_number, sender_pid} ->
        IO.inspect("take_a_number INBOX: #{Keyword.get(Process.info(self()), :message_queue_len)}")
        send(sender_pid, state+1)
        loop(state+1)
      :stop ->
        nil
      _ ->
        IO.inspect("DEFAULT INBOX: #{Keyword.get(Process.info(self()), :message_queue_len)}")
        loop(state)
    end
  end
end
```

output of `mix test test/take_a_number_test.exs:75` vary depending on how quick process information of messages queue will be flushed it may be like this:
```
  * test ignores unexpected messages and keeps working [L#75]"INIT INBOX: 3"
"DEFAULT INBOX: 2"
"INIT INBOX: 2"
"DEFAULT INBOX: 1"
"INIT INBOX: 1"
"take_a_number INBOX: 0"
"INIT INBOX: 0"
"report_state INBOX: 0"
"INIT INBOX: 0"
  * test ignores unexpected messages and keeps working (2.7ms) [L#75]

  1) test ignores unexpected messages and keeps working (TakeANumberTest)
     test/take_a_number_test.exs:75
     Assertion with == failed
     code:  assert Keyword.get(Process.info(pid), :message_queue_len) == 0
     left:  1
     right: 0
     stacktrace:
       test/take_a_number_test.exs:87: (test)



Finished in 0.08 seconds (0.00s async, 0.08s sync)
7 tests, 1 failure, 6 excluded

Randomized with seed 0
```

or this:

```
  * test ignores unexpected messages and keeps working [L#75]"INIT INBOX: 3"
"DEFAULT INBOX: 2"
"INIT INBOX: 2"
"DEFAULT INBOX: 1"
"INIT INBOX: 1"
"take_a_number INBOX: 0"
"INIT INBOX: 0"
"report_state INBOX: 0"
"INIT INBOX: 0"
  * test ignores unexpected messages and keeps working (0.6ms) [L#75]


Finished in 0.08 seconds (0.00s async, 0.08s sync)
7 tests, 0 failures, 6 excluded

Randomized with seed 0
```

Using exercise online editor gives almost always error. 

Not sure about the proper solution but, putting small `:timer.sleep(100)` just before `assert Keyword.get(Process.info(pid), :message_queue_len) == 0` makes test working as expected every time.

```elixir
  @tag task_id: 5
  test "ignores unexpected messages and keeps working" do
    pid = TakeANumber.start()

    send(pid, :hello?)
    send(pid, "I want to speak with the manager")

    send(pid, {:take_a_number, self()})
    assert_receive 1

    send(pid, {:report_state, self()})
    assert_receive 1
    :timer.sleep(100)
    assert Keyword.get(Process.info(pid), :message_queue_len) == 0
  end
```

Thanks!

angelikatyborska:
Resolves #1021

I was just about to open a PR that just removes this test and then checks in the analyzer if a catch-all was used, but i realized it would be terrible student-experience. The last test, and the only test for the last 5th task, would pass without the student having to do any work. They might not even notice that the 5th task was there if they do it task by task and rely on the test runner output to know when they're done, and then they would be hit with an analyzer warning...

I decided to do a dirty tricky and add a retry mechanism to the assertion to fight the flakiness. I know that the students on the web won't see the content of this new private function, so I tried to describe it.

### DIFF
--- a/exercises/concept/take-a-number/test/take_a_number_test.exs
+++ b/exercises/concept/take-a-number/test/take_a_number_test.exs
@@ -84,30 +84,13 @@ defmodule TakeANumberTest do
     send(pid, {:report_state, self()})
     assert_receive 1
 
-    # wait_until/3 is a custom helper function that tries to
-    # evaluate the given expression a few times until it returns true.
     # This is necessary because `Process.info/1` is not guaranteed to return up-to-date info immediately.
+    dirty_hacky_delay_to_ensure_up_to_date_process_info = 200
+    :timer.sleep(dirty_hacky_delay_to_ensure_up_to_date_process_info)
 
     # Do not use `Process.info/1` in your own code.
     # It's meant for debugging purposes only.
     # We use it here for didactic purposes because there is no alternative that would achieve the same result.
-    assert wait_until(fn ->
-             Keyword.get(Process.info(pid), :message_queue_len) == 0
-           end)
-  end
-
-  defp wait_until(condition_fn, max_retries \\ 5, current_try \\ 1) do
-    condition_fn_return = condition_fn.()
-
-    cond do
-      condition_fn_return == true ->
-        true
-
-      max_retries > current_try ->
-        :timer.sleep(100) && wait_until(condition_fn, max_retries, current_try + 1)
-
-      max_retries <= current_try ->
-        condition_fn_return
-    end
+    assert Keyword.get(Process.info(pid), :message_queue_len) == 0
   end
 end
--- a/exercises/concept/take-a-number/test/take_a_number_test.exs
+++ b/exercises/concept/take-a-number/test/take_a_number_test.exs
@@ -84,6 +84,30 @@ defmodule TakeANumberTest do
     send(pid, {:report_state, self()})
     assert_receive 1
 
-    assert Keyword.get(Process.info(pid), :message_queue_len) == 0
+    # wait_until/3 is a custom helper function that tries to
+    # evaluate the given expression a few times until it returns true.
+    # This is necessary because `Process.info/1` is not guaranteed to return up-to-date info immediately.
+
+    # Do not use `Process.info/1` in your own code.
+    # It's meant for debugging purposes only.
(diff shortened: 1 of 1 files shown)


jiegillet (approved):
This works, although, yes, it does feel a bit dirty :)
I couldn't think of an alternative though. 

I think the comments will not be visible though, since the code is going through the AST stage... Maybe the comments should be a multiline string, python style?

EDIT: very possibly a multiline string will be transformed to a single line one...

angelikatyborska:
Damn, you're right that the comment will be deleted :| I'm not sure if it's ok to add a test with an unknown function and no comment...

Maybe a "fake" multiline string could be achieved in the same way as in problem specs, concatenation?
```elixir
"line 1"
+ "line 2"
```

But that will be weird 🙈

jiegillet:
Then we could simplify the code
```elixir
delay_for_synchronization = 500
:timer.sleep(delay_for_synchronization)
Keyword.get(Process.info(pid), :message_queue_len) == 0
```

angelikatyborska:
I hate arbitrary sleeps that don't check the condition before the sleep and don't re-check the condition on failure 😭 but it might be the only option that would be understandable to students via the web editor. I'll make the commit 💔.

jiegillet:
💔

Issue #1069, opened 2022-02-24
Change charlist errors to string/atom errors in OCR numbers


angelikatyborska:
Charlist errors are unusual. Let's fix this exercise to use either strings or atoms for errors.

https://github.com/exercism/elixir/blob/f452285f2243d04d88ba0afcc097844929012782/exercises/practice/ocr-numbers/test/ocr_numbers_test.exs#L52

https://github.com/exercism/elixir/blob/f452285f2243d04d88ba0afcc097844929012782/exercises/practice/ocr-numbers/test/ocr_numbers_test.exs#L65

PR #1083, opened 2022-03-11
Change charlist errors to string

richardjypark:
solved issue: #1069

Updated all references to charlist in the documentation and test code, including any examples (i.e. '?' )

### DIFF
--- a/exercises/practice/ocr-numbers/.docs/instructions.md
+++ b/exercises/practice/ocr-numbers/.docs/instructions.md
@@ -27,7 +27,7 @@ Is converted to "0"
 
 Is converted to "1"
 
-If the input is the correct size, but not recognizable, your program should return '?'
+If the input is the correct size, but not recognizable, your program should return "?"
 
 If the input is the incorrect size, your program should return an error.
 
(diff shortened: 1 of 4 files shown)


exercises/practice/ocr-numbers/.docs/instructions.md
```diff
@@ -27,7 +27,7 @@ Is converted to "0"
 
 Is converted to "1"
 
-If the input is the correct size, but not recognizable, your program should return '?'
+If the input is the correct size, but not recognizable, your program should return "?"
```
angelikatyborska, line 30: All the instructions documents come from a language-independent central repository (problem-specifications) and we cannot make Elixir-specific changes in them. They would be overwritten the next time we synchronize with that repository anyway:
```suggestion
If the input is the correct size, but not recognizable, your program should return '?'
```

angelikatyborska (approved):
Excellent! Thank you 💜

angelikatyborska:
(Note to other maintainers: I know that this exercise still has "charlists" in prerequisites and I decided to keep them so that people can decide for themselves if they think charlists are needed here or not)

PR #1084, opened 2022-03-13
Small fixes to `basics` intro

angelikatyborska:
Prompted by @SleeplessByte's comment https://github.com/exercism/elixir-analyzer/issues/267#issuecomment-1066008048, I reviewed the basics intro to check if it mentions that module names can be skipped in local calls. It did, but in a way that was very easy to miss.

I rewrote the basics intro a bit so that:
- There is a code example of calling a function without a module name.
- Both `def` and `defp` are clearly mentioned by name and shown in code examples. The design document claimed `depf` is out of scope, but that's wrong, no other concept mentions `defp` later and we are using it in other exercises.
- Remove the phrase "access modifier" because it doesn't apply in Elixir like it does in other languages where "public" or "private" are stand-alone keywords.

### DIFF
--- a/concepts/basics/introduction.md
+++ b/concepts/basics/introduction.md
@@ -23,25 +23,46 @@ end
 
 ## Named functions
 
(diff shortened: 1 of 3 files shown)


SleeplessByte:
🎉

jiegillet (approved):
Nice!
A little more spread out and readable.

PR #1085, opened 2022-03-14
Reduce list length for test "huge sublist not in huge list"

jnnks:
Reduce the number of tested items to reduce execution time.

O(n²) algorithms will take a long time and timeout the browser test evaluation

### DIFF
--- a/exercises/practice/sublist/test/sublist_test.exs
+++ b/exercises/practice/sublist/test/sublist_test.exs
@@ -68,8 +68,8 @@ defmodule SublistTest do
   end
 
   @tag :pending
-  test "huge sublist not in huge list" do
-    assert Sublist.compare(Enum.to_list(10..1_000_001), Enum.to_list(1..1_000_000)) == :unequal
+  test "large sublist not in large list" do
+    assert Sublist.compare(Enum.to_list(10..50_001), Enum.to_list(1..50_000)) == :unequal
   end
 
   @tag :pending


angelikatyborska:
Thank you for the feedback!

Writing an efficient sublist check is one of the challenges of the exercise. It is possible to come up with a solution that passes the tests in the browser fast enough.

Decreasing the list length would make the exercise less challenging for all users, and I'm not sure we want to do that because it is otherwise already pretty simple. One thing that we can do is to tag this specific test as slow, which would disable it when using the web editor, but still keep it as challenging as it is now when working on your own computer. We did a similar thing for example for list ops:
- [example slow test](https://github.com/exercism/elixir/blob/main/exercises/practice/list-ops/test/list_ops_test.exs#L43-L46)
- [slow test explanation in instructions](https://github.com/exercism/elixir/blob/main/exercises/practice/list-ops/.docs/instructions.append.md#slow-tests)

@jiegillet @neenjaw what are your opinions about this?

neenjaw:
I'd be okay to tag it as slow, but I think the test case adds a particular challenge appropriate to the problem exercise as it is.

jiegillet:
The exercises we tagged as slow are ones that are slow even for the example implementations, or quite difficult to work out. I agree with @neenjaw, I think it's appropriate to keep this exercise as it is. I won't make a fuss if we tag it as slow though.

jnnks:
I understand that writing efficient algorithms is a learning goal, but I also feel like an easy exercise is not the learning level to expect that. Maybe you can consider adding a hint regarding performance.

angelikatyborska:
@jnnks After we merge https://github.com/exercism/elixir/pull/1087, this test won't be run on the website, and there will be a warning in the exercise instructions that some of the tests are too slow to run on the website.

PR #1088, opened 2022-03-17
clarify the integer expectations on secrets exercise

maco:
I had a mentee who got confused about why the secrets exercise wanted integer division.


angelikatyborska (approved):
Makes sense 👍

PR #1089, opened 2022-03-17
Add doc tests to the doc concept and corresponding city-office exercise

maco:
The docs concept didn't explain doc tests or check that people could write them, and that seemed like a gap. This fills that gap.

(The test for this is based on ExUnit's own tests.)

### DIFF
--- a/bin/test_exercises.sh
+++ b/bin/test_exercises.sh
@@ -93,8 +93,14 @@ do
       module_name=$(cat "${test_file}" | sed -rn 's/^defmodule (.*)Test do$/\1 /p')
       doctest_code="doctest ${module_name}"
 
-      # .bak and rm part left here to simplify GNU/BSD sed options compatibility
-      sed -i.bak 's/use ExUnit.Case\(.*\)/use ExUnit.Case\1\n'" ${doctest_code}"'\n/g' "${test_file}" && rm -f "${test_file}.bak"
+      if [ ${module_name} != "Form" ]
+      then
+        # This module is skipped because it contains tests *for* doc tests.
+        # It already *has* a `doctest ...` line, and mix really doesn't like seeing that in a module twice.
+
+        # .bak and rm part left here to simplify GNU/BSD sed options compatibility
+        sed -i.bak 's/use ExUnit.Case\(.*\)/use ExUnit.Case\1\n'" ${doctest_code}"'\n/g' "${test_file}" && rm -f "${test_file}.bak"
+      fi
 
       # perform unit tests
       set +e
--- a/bin/test_exercises.sh
+++ b/bin/test_exercises.sh
@@ -93,14 +93,8 @@ do
       module_name=$(cat "${test_file}" | sed -rn 's/^defmodule (.*)Test do$/\1 /p')
       doctest_code="doctest ${module_name}"
 
-      if [ ${module_name} != "Form" ]
-      then
-        # This module is skipped because it contains tests *for* doc tests.
-        # It already *has* a `doctest ...` line, and mix really doesn't like seeing that in a module twice.
-
-        # .bak and rm part left here to simplify GNU/BSD sed options compatibility
-        sed -i.bak 's/use ExUnit.Case\(.*\)/use ExUnit.Case\1\n'" ${doctest_code}"'\n/g' "${test_file}" && rm -f "${test_file}.bak"
-      fi
+      # .bak and rm part left here to simplify GNU/BSD sed options compatibility
+      sed -i.bak 's/^  use ExUnit.Case\(.*\)/  use ExUnit.Case\1\n'" ${doctest_code}"'\n/' "${test_file}" && rm -f "${test_file}.bak"
 
       # perform unit tests
       set +e
@@ -141,7 +135,7 @@ do
 done
 
 # clean up
-rm -rf tmp-exercises
+#rm -rf tmp-exercises
 
 # report
 printf -- '-%.0s' {1..80}; echo ""
--- a/bin/test_exercises.sh
+++ b/bin/test_exercises.sh
@@ -135,7 +135,7 @@ do
 done
 
 # clean up
-#rm -rf tmp-exercises
+rm -rf tmp-exercises
 
 # report
 printf -- '-%.0s' {1..80}; echo ""
--- a/concepts/docs/about.md
+++ b/concepts/docs/about.md
@@ -6,6 +6,7 @@ Elixir documentation:
 - Written in [**Markdown**][markdown].
 - Added by using special module attributes.
 - Typically uses the heredoc syntax for multiline strings.
+- Can include unit tests.
 
 Module attributes used for documentation:
 
@@ -91,6 +92,10 @@ Many modern IDEs that support Elixir can parse and display documentation and typ
 ## Internal modules and function
 
 If a module or a function is intended for internal usage only, you can mark it with `@moduledoc false` or `@doc false`. Those modules and functions will not be included in the generated documentation. Note that that doesn't make them private. They can still be invoked and/or imported. Check the [official documentation about hiding internal modules and functions][hiding-internal-modules-and-functions] to learn about potential solutions to this problem.
+
+## Doc tests
+
+You can attach unit tests directly to functions by including them in the `@doc` attribute. These are offset by indenting them 4 spaces. They start with a `iex>` prompt, just like in an `iex` shell. The output from the test is on a new line, indented the same 4 spaces. A blank space goes between each test. See the example at the top of the page.
 
 [markdown]: https://docs.github.com/en/github/writing-on-github/basic-writing-and-formatting-syntax
 [official-documentation]: https://hexdocs.pm/elixir/
--- a/exercises/concept/city-office/test/form_test.exs
+++ b/exercises/concept/city-office/test/form_test.exs
@@ -210,9 +210,8 @@ defmodule FormTest do
         doctest Form, only: [{:check_length, 2}], import: true
       end
 
-      output = capture_io(fn -> ExUnit.run() end)
-      IO.inspect(output, label: "output")
-      assert output =~ "2 doctests, 0 failures"
+      doctest_results = ExUnit.run()
+      assert doctest_results == %{excluded: 0, failures: 0, skipped: 0, total: 2}
       assert_doc({:check_length, 2}, "\n    :ok")
       assert_doc({:check_length, 2}, ~r/\n    {:error, \d+}/)
     end
--- a/exercises/concept/city-office/test/form_test.exs
+++ b/exercises/concept/city-office/test/form_test.exs
@@ -1,6 +1,5 @@
 defmodule FormTest do
   use ExUnit.Case
-  import ExUnit.CaptureIO
 
   # Dear Elixir learner,
   # If you're reading this test suite to gain some insights,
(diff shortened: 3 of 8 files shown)


angelikatyborska:
(Note: I'm leaving a short, not well-research opinion first, to kick off a discussion with other maintainers)

Hmm, I'm not entirely sure how to feel about this idea. On one hand, doctests are awesome and I would expect all Elixir devs to know them. On the other, learning writing tests was never in scope for students on Exercism because tests are always already given.

We also need to check if this is going to work in the test runner. If it works in the test runner, and if we agree that it's a good idea to teach doctests, then I think this is a good exercise to add them to. I would probably even want to add a doctest writing step to each function, each as a separate step in instructions, so that the exercise is symmetrical.

@jiegillet @neenjaw I'm looking forward to your opinions here

angelikatyborska:
I did a quick check in the test runner (commit https://github.com/exercism/elixir-test-runner/commit/b1ade5c98160364163e63e2d6eb7c2b21502db62 based on Jie's unmerged work in https://github.com/exercism/elixir-test-runner/pull/86, command: `./bin/smoke_test.sh`). It doesn't work in the test runner. The new test fails. I'm not yet sure why.

```
parsing /tmp/solution_pKj7YEPWYN/test/form_test.exs for metadata
transforming /tmp/solution_pKj7YEPWYN/test/form_test.exs
3c3
<   "status": "pass",
---
>   "status": "fail",
118c118,119
<       "message": null,
---
>       "message": "  1) test check_length/2 has 2 passing doctests (FormTest)\n     \u001b[1m\u001b[30mtest/form_test.exs:139\u001b[0m\n     \u001b[31mAssertion with =~ failed\u001b[0m\n     \u001b[36mcode:  \u001b[0massert output =~ \"2 doctests, 0 failures\"\n     \u001b[36mleft:  \u001b[0m\"\u001b[48;5;88m\u001b[0m\"\n     \u001b[36mright: \u001b[0m\"\u001b[32m2 doctests, 0 failures\u001b[0m\"\n     \u001b[36mstacktrace:\u001b[0m\n       test/form_test.exs:146: (test)\n",
>       "name": "test check_length/2 has 2 passing doctests",
120c121
<       "status": "pass",
---
>       "status": "fail",
🔥 test/city-office: expected /tmp/output_lLFCXZHBPy/results.json to equal test/city-office/expected_results.json on successful run 🔥
```
For readability, formatted message is this:
Screenshot 2022-03-17 at 08 04 50

jiegillet:
I do like the idea of talking about doctests, and actually, why not adding a testing concept? It's an important topic, and it would be fun to figure out how to test tests :)

I thought that the test-runner could handle doctests, I'll take a closer look soon and figure out why it doesn't here.

jiegillet:
I played around for a bit, and I think the test-runner is not happy because before running the tests, it's modifying the source code os the tests by traversing the AST, getting rid of the docs and therefore the doctests in the process.

It's the second use case in as many weeks where we would actually prefer keeping the comments. Since 1.13 it should be possible to keep the comments, but it's a bunch of work. What do you think @angelikatyborska?

I also noticed that the CI is failing:
```
* test check_length/2 has 2 passing doctests (60023.9ms)
  1) test check_length/2 has 2 passing doctests (FormTest)
     test/form_test.exs:207
     ** (exit) exited in: GenServer.call(ExUnit.Server, {:take_async_modules, 1}, 60000)
         ** (EXIT) time out
     stacktrace:
       (elixir) lib/gen_server.ex:989: GenServer.call/3
```
I haven't gotten to the bottom of that yet.

exercises/concept/city-office/test/form_test.exs
```diff
@@ -202,6 +203,19 @@ defmodule FormTest do
...
+
+    test "has 2 passing doctests" do
+      defmodule CheckLengthOnly do
+        use ExUnit.Case
+        doctest Form, only: [{:check_length, 2}], import: true
+      end
+
+      output = capture_io(fn -> ExUnit.run() end)
+      IO.inspect(output, label: "output")
+      assert output =~ "2 doctests, 0 failures"
```
jiegillet, line 215:
```suggestion
      doctest_results = ExUnit.run()
      assert doctest_results == %{excluded: 0, failures: 0, skipped: 0, total: 2}
```

jiegillet: Also with this, we don't need ` import ExUnit.CaptureIO`.

bin/test_exercises.sh
```diff
@@ -93,8 +93,14 @@ do
       module_name=$(cat "${test_file}" | sed -rn 's/^defmodule (.*)Test do$/\1 /p')
       doctest_code="doctest ${module_name}"
 
-      # .bak and rm part left here to simplify GNU/BSD sed options compatibility
-      sed -i.bak 's/use ExUnit.Case\(.*\)/use ExUnit.Case\1\n'" ${doctest_code}"'\n/g' "${test_file}" && rm -f "${test_file}.bak"
+      if [ ${module_name} != "Form" ]
```
jiegillet, line 96: I'm not super comfortable with hardcoding a file name. 
Maybe detecting the use of doctests in the each file before using `sed` would be better?

jiegillet (commented):
The tests failed in the test-runner because of the formatter, the IO output wasn't passed properly, maybe because the formatter is a GenServer? I have no idea how to fix that, so instead I realized that `ExUnit.run()` returns a map with the results and we can use that. 
That might also fix the CI?

angelikatyborska:
Great! If we can make the tests pass without modifying the test runner like Jie says, and is Jie likes the idea of the doc tests too, then I think we should add them 🎉.

I think ideally every exercise step that says "Add documentation and a typespec to the ... function." would also be extended to include adding a doctest. It's great that the `doctest` macro accepts an `:only` option, that will allow us to create a separate unit test for each step.

@maco are you interested in making those changes and do you have time for that? If not, I can also take over.

exercises/concept/city-office/test/form_test.exs
```diff
@@ -202,6 +203,19 @@ defmodule FormTest do
         ":ok | {:error, pos_integer()}"
       )
     end
+
+    test "has 2 passing doctests" do
```
angelikatyborska, line 207: Every test needs a `@task_id` tag

concepts/docs/about.md
```diff
@@ -92,6 +93,10 @@ Many modern IDEs that support Elixir can parse and display documentation and typ
 
 If a module or a function is intended for internal usage only, you can mark it with `@moduledoc false` or `@doc false`. Those modules and functions will not be included in the generated documentation. Note that that doesn't make them private. They can still be invoked and/or imported. Check the [official documentation about hiding internal modules and functions][hiding-internal-modules-and-functions] to learn about potential solutions to this problem.
 
+## Doc tests
```
angelikatyborska, line 96: This file shows up to students only after completing the learning exercise. We need to explain doc tests in those two files as well:
- `concepts/docs/introduction.md`
- `exercises/concept/city-office/.docs/introduction.md`

maco:
Yes, I'm fine to make those changes.

concepts/docs/about.md
maco: Oh, I'd been trying to figure out what was the difference between the various markdown files.

maco:
Ok, I made a bunch of changes, but not the part involving sed yet. I need to noodle that a bit because crossing newline boundaries with regex is tricky.

bin/test_exercises.sh
jiegillet: I thought about this a bit more...
The problem is that there would be two `doctest ModuleName` in a row right? Maybe we can simply use `sed` again to remove double appearances? Is that too clunky?

jiegillet:
@maco do you still have time/motivation to work on these, or would you like us to take over?

maco:
@jiegillet a month later, with fresh eyes, I realize the `DefModule` regex starts at `^` to avoid matching nested ones, and I can do the same on the `use ExUnit.Case` regex.

I'm unclear on why this timed out in CI, though.

bin/test_exercises.sh
```diff
@@ -135,7 +135,7 @@ do
 done
 
 # clean up
-rm -rf tmp-exercises
+#rm -rf tmp-exercises
```
jiegillet, line 138: This should probably be uncommented.

jiegillet:
I took a look at the CI failure, but I'm not sure what to do. The problem is probably related to running tests within tests (GenServers within GenServers...). It fails in Elixir 1.8 and 1.9, and succeeds in 1.12 and 1.13 (1.10 has been hanging for 17 minutes, not a good sign).

Not the cleanest suggestion, but we could simply rely on checking that well formatted doctests have been written. The doctests are being tested, so if they fail, it will be caught.

neenjaw:
If the code is correct by current elixir syntax/patterns, the easiest thing to do is drop `1.8`, `1.9`, `1.10` in CI.

I have no comment on the correctness of this PR, haven't looked at it in detail, but arguably if it doesn't work for those versions, yet does for later ones -- there was a bug/limitation in ex_unit which then was parched/extended.

bin/test_exercises.sh
maco: done ✅

angelikatyborska:
Just FYI, I think it's this bug: https://github.com/elixir-lang/elixir/issues/9601

Fixed in https://github.com/elixir-lang/elixir/commit/bae1cc7cdf1a5beb27530ff8d85869e4c845cda6 which was part of 1.10, but not lower

angelikatyborska:
>  It fails in Elixir 1.8 and 1.9, and succeeds in 1.12 and 1.13 (1.10 has been hanging for 17 minutes, not a good sign).

I tested in the official Docker containers `elixir:1.10-alpine` and `elixir:1.11-alpine`, and it hangs for a very long time in those too, which means only 1.12 and 1.13 work as expected with this approach.

jiegillet:
I don't think we want to drop all these versions :(

Also, after all this trouble, I checked and `elixir-test-runner` doesn't actually run doctests at the moment. When adding the line `doctest Form` to `city-office/test/form_test.exs`, the doctests do run, and test runner output is
```json
{
  "message": null,
  "status": "pass",
  "tests": [
    {
      "message": null,
      "metadata-error": "metadata missing for test",
      "name": "doctest Form.blanks/1 (1)",
      "output": null,
      "status": "pass",
      "test_code": null
    },
   ...

```
when everything passes and
```json

{
  "message": null,
  "status": "fail",
  "tests": [
    {
      "message": "  1) doctest Form.blanks/1 (1) (FormTest)\n     \u001b[1m\u001b[30mtest/form_test.exs:3\u001b[0m\n     \u001b[31mDoctest failed\u001b[0m\n     \u001b[36mdoctest:\u001b[0m\n       iex> Form.blanks(3)\n       \"XXXXX\"\n     \u001b[36mcode:  \u001b[0mForm.blanks(3) === \"XXXXX\"\n     \u001b[36mleft:  \u001b[0m\"XXX\"\n     \u001b[36mright: \u001b[0m\"XXX\u001b[32mXX\u001b[0m\"\n     \u001b[36mstacktrace:\u001b[0m\n       lib/form.ex:17: Form (module)\n",
      "metadata-error": "metadata missing for test",
      "name": "doctest Form.blanks/1 (1)",
      "output": null,
      "status": "fail",
      "test_code": null
    },
...
```
when the first doctest fails (and everything else passes).
The tests do run and the final result is accurate, but the messages are really bad. It would require quite a bit of work to make it nicer. 

I have a terrible, terrible, hacky alternative, please let me know what you think. The following test works
```elixir
    test "has a passing doctest, regex edition" do
      doctest_regex = ~r'''
          iex> Form.blanks\((?<input>\d*)\)
          "(?<output>.*)"
      '''

      assert_doc({:blanks, 1}, doctest_regex)

      assert %{"input" => input, "output" => output} =
               doc_named_captures({:blanks, 1}, doctest_regex)

      assert Form.blanks(String.to_integer(input)) == output
    end
```
where `doc_named_captures` is the exact copy of `assert_doc` expect for the last line that is changed to `Regex.named_captures(unquote(expected_doc), actual_doc)` (to be be refactored later I guess).

jiegillet:
@maco an Exercism-wide cleanup wave is on the rise.
Are you still interested in pursuing this PR, maybe along the line of what I suggested above? 
If not, I think we will drop it.

PR #1091, opened 2022-03-19
Remove duplicate "position" in darts typespec

rikroos:
The @type "position" reflexes a tuple like {x,y}. 
I think the @spec should not repeat "position" twice in the description of the first and only parameter.

### DIFF
--- a/exercises/practice/darts/.meta/example.ex
+++ b/exercises/practice/darts/.meta/example.ex
@@ -4,7 +4,7 @@ defmodule Darts do
   @doc """
   Calculate the score of a single dart hitting a target
   """
-  @spec score(position :: position) :: integer
+  @spec score(position) :: integer
   def score({x, y}) do
     radius = :math.sqrt(x * x + y * y)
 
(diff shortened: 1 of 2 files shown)


jiegillet:
Thank you, yes, I guess that makes perfect sense.
Would you mind changing it in the example solution too?

rikroos:
I will change the the example solution too soon. I was not aware of the existence. That will become a separate pull request I think, I will refer to Tthis #1091.

angelikatyborska:
@rikroos It doesn't make sense to split it into two PRs. Can you push a second commit to this PR instead?

angelikatyborska:
I'm going to assume that you're using GitHub's UI to push changes so it might not be immediately obvious how to add more changes to this PR. One way to achieve that could be to open a PR from `rikroos:patch-4` into `rikroos:patch-3` and then you merge that immediately because both branches are yours. Then, the changes would appear in this PR 🙂. There are probably other ways, but this should be the easiest since the other branch with the other change already exists.

rikroos:
I have deleted the unnecessary second PR and added a new commit (example.ex) to the old patch3.
I am not very familiar with github. Apologies for some overhead.

jiegillet (approved):
Thanks a lot for the change and for managing to do it all in this PR.
The GitHub interface can be a little tricky, you have a lot more control when you use git in the command line.

PR #1092, opened 2022-03-19
Update example.ex

rikroos:
See also PR#1091.
The @type position in the @spec should be referenced only once.


rikroos:
This PR is not necessary. Change will be added to a previous branch.

PR #1094, opened 2022-03-22
Enable configlet fmt on CI

angelikatyborska:


jiegillet (approved):
What's not to like about a formatting PR :)

Issue #1097, opened 2022-03-23
Missing .gitignore files


Tyyagoo:
### Problem

Some exercises doesn't have a .gitignore file, allowing inattentive students to end up committing unwanted files.
_e.g. _build folder_

### Examples

- affine-cipher
- circular-buffer
- complex-numbers
- darts
- house
- rational-numbers
- react
- resistor-color-duo
- resistor-color-trio
- square-root
- variable-length-quantity
- zebra-puzzle

_If this is really a unexpected behavior, I can submit a pull request to fix it._

glennj:
I think they might be referring to students commit exercise to a personal git repo. I do this

Tyyagoo:
I double-checked all the exercises and found only those others that were missing .gitignore

- book-store
- food-chain
- go-counting
- knapsack
- pov
- proverb
- rectangles
- satellite
- sgf-parsing
- two-bucket
- word-search

PR #1099, opened 2022-03-23
Add missing `.gitignore` files

Tyyagoo:
Fixed exercism/elixir#1097


angelikatyborska (approved):
Perfect! Thank you for finding this problem and fixing it 💜

Issue #1071, opened 2022-02-27
Set 5 seconds time limit for test in `rpn-calculator-inspection`


jiegillet:
In #1041 @bitboxx introduced a clever mechanism to make the long running tests fail within a time limit.

There is another test in the same exercise that can also take too long for faulty solutions: "returns a map when input list has 1000 elements and the calculator takes 50ms for each calculation". I just mentored a solution like that, and the student was wondering why it passed locally but not online. 

It would be great to make that test fail within 5 seconds.

Faulty solution for testing purposes (doing everything inside of `Enum.reduce`):
```elixir
  def reliability_check(calculator, inputs) do
    oldvalue = Process.flag(:trap_exit, true)
    result_map =
      Enum.reduce(inputs, %{}, fn input, acc ->
                    start_reliability_check(calculator, input)
                    |> await_reliability_check_result(acc) end)
    Process.flag(:trap_exit, oldvalue)
    result_map
  end
```

jiegillet:
Closed via #1102.

PR #1041, opened 2022-01-22
Set 5 seconds time limit for test to pass.

bitboxx:
Allow test to timeout and fail when exercise solution is taking too much time to finish. This improves user experience and to fix the issue that the tests pass on local but not on-site.

PR #1102, opened 2022-03-26
#1071 - Set 5 second timeout limit to another test

butterknight:
👋 

I've attempted a solution for #1071 based on #1041


jiegillet (approved):
Looks great, thank you very much!

PR #1103, opened 2022-03-27
New concept exercise `dancing-dots` (`use` and `behaviour`)

angelikatyborska:
~Heavily WIP!~

~I'm opening this PR to signal that I'm working on this.~

This will be necessary as a prerequisite for the new GenServer exercise (https://github.com/exercism/elixir/pull/1076)


exercises/concept/dancing-dots/lib/dancing_dots/dot_group.ex
```diff
@@ -0,0 +1,60 @@
+defmodule DancingDots.DotGroup do
+  # This module is an example of how behaviours can be used in practice.
+  # You don't need to read it to solve this exercise.
+  # It's here for the curious :)
```
angelikatyborska, line 4: I'm not sure if it's a good idea to leave this code here. But one of my biggest problems with understanding behaviours is actually imagining how they can be used once I have some modules that all implement that one behaviour.

angelikatyborska:
@jiegillet The example solution, tests, and instructions are ready for review. Can you take a look?

exercises/concept/dancing-dots/lib/dancing_dots/dot_group.ex
jiegillet: Yeah, I'm with you, it's nice to see how it can be used. `render_dots/2` is great because it's very readable, but I think maybe `new/2` is a bit too complex... Maybe it could be changed to filter the valid options rather than throwing an error for the sake of simplicity?

(untested idea)
```elixir
def new(dots, animations_with_opts) do
  valid_animations_with_opts =
  for {animation_module, opts} <- animations_with_opts,
      # using Animation's init/1 callback
      {:ok, opts} <- animation_module.init(opts) do
    {animation_module, opts}
  end

  {:ok,  %__MODULE__{dots: dots, animations_with_opts: valid_animations_with_opts }}
end

```

exercises/concept/dancing-dots/.meta/design.md
```diff
@@ -0,0 +1,23 @@
+# Design
+
+## Learning objectives
+
+- Know how to define a behaviour with callbacks
+- Know how to provide a default implementation of a callback
+- Know how to use a behaviour
+- `@impl`
+
+## Out of scope
```
jiegillet, line 10: -  `@macrocallback`
- `@optional_callbacks`

exercises/concept/dancing-dots/.meta/exemplar.ex
```diff
@@ -0,0 +1,47 @@
...
+  use DancingDots.Animation
+
+  @impl DancingDots.Animation
+  def init(opts) do
+    velocity = Keyword.get(opts, :velocity)
+
+    if is_number(velocity) do
+      {:ok, [velocity: velocity]}
+    else
+      {:error, "Expected required option :velocity to be a number, got: #{inspect(velocity)}"}
```
jiegillet, line 39: I would suggest to add one type of errors here: "Missing option :velocity". And some tests for that.

exercises/concept/dancing-dots/test/dancing_dots/animation_test.exs
```diff
@@ -0,0 +1,199 @@
...
+    test "every frame after the first, handle_frame/3 grows the dot's radius by the given velocity" do
+      dot1 = %DancingDots.Dot{x: 90, y: 90, radius: 100, opacity: 0.5}
+      dot2 = %DancingDots.Dot{x: 100, y: 0, radius: 400, opacity: 0.9}
+
+      assert DancingDots.Zoom.handle_frame(dot1, 1, velocity: 3) == %{dot1 | radius: 100}
+      assert DancingDots.Zoom.handle_frame(dot1, 2, velocity: 3) == %{dot1 | radius: 103}
+      assert DancingDots.Zoom.handle_frame(dot1, 3, velocity: 3) == %{dot1 | radius: 106}
+      assert DancingDots.Zoom.handle_frame(dot1, 101, velocity: 3) == %{dot1 | radius: 400}
+
+      assert DancingDots.Zoom.handle_frame(dot2, 1, velocity: -1) == %{dot2 | radius: 400}
```
jiegillet, line 173: How about splitting this test in two here and mention negative velocity? Just because the instructions mention it but the test descriptions don't.

jiegillet (commented):
Very very cool! The story is great and there is a lot to learn.

I took my time solving with only the instructions and the Elixir documentation. What a mess!! (the docs, not your instructions) All the ingredients you need for behaviours are scattered everywhere in like 3 or 4 different pages. This really is a needed exercise! I think even a nice blog post about the topic would be great.

For reference, here is my final submission, it's very very similar to yours:
```elixir
defmodule DancingDots.Animation do
  @type dot :: DancingDots.Dot.t()
  @type opts :: keyword
  @type error :: any 
  @type frame_number :: pos_integer

  @callback init(opts) :: {:ok, opts} | {:error, error}
  defmacro __using__(_opts) do
    quote do
      @behaviour DancingDots.Animation
      def init(opts) do
        {:ok, opts}
      end 

      defoverridable init: 1
    end 
  end 

  @callback handle_frame(dot, frame_number, opts) :: dot 
end

defmodule DancingDots.Flicker do
  use DancingDots.Animation
  alias DancingDots.Dot

  @impl DancingDots.Animation
  def handle_frame(%Dot{opacity: opacity} = dot, frame_number, _opts)
      when rem(frame_number, 4) == 0 do
    %Dot{dot | opacity: opacity / 2}
  end 

  def handle_frame(%Dot{} = dot, _frame_number, _opts), do: dot 
end

defmodule DancingDots.Zoom do
  use DancingDots.Animation
  alias DancingDots.Dot

  @impl DancingDots.Animation
  def init(opts) do
    velocity = Keyword.get(opts, :velocity)

    if is_number(velocity) do
      {:ok, opts}
    else
      {:error, "Expected required option :velocity to be a number, got: #{inspect(velocity)}"}
    end 
  end 

  @impl DancingDots.Animation
  def handle_frame(%Dot{radius: radius} = dot, frame_number, opts) do
    n = frame_number - 1 
    velocity = Keyword.get(opts, :velocity)
    %Dot{dot | radius: radius + n * velocity}
  end 
end
```

Outside of finding information in the docs, the only place that gave me pause was the error message in `DancingDots.Zoom`, I left some comments about that.

exercises/concept/dancing-dots/.meta/exemplar.ex
angelikatyborska: I'm not sure I understand the suggestion, are you saying there need to be two different errors, one for when velocity was not given, and one for when it's given but not valid? There already is a test for when it's missing:
```elixir
      assert DancingDots.Zoom.init([]) ==
               {:error, "Expected required option :velocity to be a number, got: nil"}
```

exercises/concept/dancing-dots/lib/dancing_dots/dot_group.ex
angelikatyborska: I agree, `new` was too complex. I think the worst offender there was the `reduce_while` call. To make it simpler, I decided to split `new` into two functions - `new` that just returns a group with no animations, and `add_animation` to add a single animation module to the list. That allows getting rid of any `Enum` calls.

I didn't want to silently ignore initialization errors in this demo because that would show an anti-pattern IMO.

concepts/behaviours/about.md
```diff
@@ -0,0 +1,3 @@
+# About
+
+TODO: copy from exercise's introduction
```
angelikatyborska, line 3: TODO before merging

concepts/behaviours/introduction.md
```diff
@@ -0,0 +1,3 @@
+# Introduction
+
+TODO: copy from exercise's introduction
```
angelikatyborska, line 3: TODO before merging

concepts/use/about.md
```diff
@@ -0,0 +1,3 @@
+# About
+
+TODO: copy from exercise's introduction
```
angelikatyborska, line 3: TODO before merging

angelikatyborska:
This is now ready for the final full review.

exercises/concept/dancing-dots/.docs/introduction.md
```diff
@@ -0,0 +1,91 @@
...
+    end
+  end
+end
+```
+
+The options can be given as a second argument when calling `use`, e.g. `use ExUnit.Case, async: true`. When not given explicitly, they default to an empty list.
+
+## Behaviours
+
+Behaviours allow us to define interfaces (sets of functions and macros) in a _behaviour module_ that can be later implemented by different _callback modules_. Thanks to the shared interface, those callback modules can be used interchangeably.
```
kotp, line 39: Same here for English (US) spelling.

exercises/concept/dancing-dots/.docs/introduction.md
```diff
@@ -0,0 +1,91 @@
...
+
+~~~~exercism/note
+Note the British spelling of "behaviours".
+~~~~
+
+### Defining behaviors
```
kotp, line 45: Even if we accept that this exercise is the exception to the US English spelling of words (or perhaps "behaviours" is the only exception in this exercise) then it should at least be consistently spelled.

kotp: Sorry, this is all I have time for now…

exercises/concept/dancing-dots/.docs/hints.md
```diff
@@ -0,0 +1,52 @@
+# Hints
+
+## General
+
+- Read about behaviours in the official [Getting Started guide][getting-started-behaviours].
+- Read about behaviours on [elixirschool.com][elixir-school-behaviours].
+- Read about behaviours in the [documentation][doc-behaviours].
```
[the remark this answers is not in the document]
angelikatyborska: Thanks, but in Elixir, "behaviour" is a concept and whenever we refer to this concept, we need to use the UK spelling because it affects the code too (e.g. you need to write `@behaviour DancingDots.Animation`, not `@behavior`). Other times, when we use "behavior" in its regular life non-Elixir meaning, we use the US spelling.

There was one mistake in my PR, I fixed it: [`5827335` (#1103)](https://github.com/exercism/elixir/pull/1103/commits/5827335fb684a5c9fe7914f6f09efcc84d7a1909)

exercises/concept/dancing-dots/.docs/hints.md
```diff
@@ -0,0 +1,52 @@
...
+- In the quoted expression, implement the `init/1` function.
+- The default implementation of the `init/1` function should wrap the given `opts` argument in `:ok` tuple.
+- There is [a macro][defoverridable] that can mark a function as overridable.
+
+## 3. Implement the `Flicker` animation
+
+- Make use of `DancingDots.Animation` `__using__/1` macro by calling [this one special macro][doc-use] in the `DancingDots.Flicker` module.
+- You do not need to implement the `init/1` function. Its default implementation is enough.
+- You need to implement the `handle_frame/3` function.
+- To detect "every 4th frame", you can check if the [remainder][rem] when dividing it by 4 is equal to 0.
```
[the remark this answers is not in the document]
angelikatyborska: We use the real function/macro names in link names, hence `rem`.

exercises/concept/dancing-dots/.docs/hints.md
```diff
@@ -0,0 +1,52 @@
...
+- The macros' argument can be ignored.
+- The macro must return a [quoted expression][quote].
+- In the quoted expression, use `@behaviour` so that calling `use DacingDots.Animation` sets `DacingDots.Animation` as the using module's behaviour.
+- In the quoted expression, implement the `init/1` function.
+- The default implementation of the `init/1` function should wrap the given `opts` argument in `:ok` tuple.
+- There is [a macro][defoverridable] that can mark a function as overridable.
+
+## 3. Implement the `Flicker` animation
+
+- Make use of `DancingDots.Animation` `__using__/1` macro by calling [this one special macro][doc-use] in the `DancingDots.Flicker` module.
```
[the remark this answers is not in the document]
angelikatyborska: I disagree, this makes it much harder for us to know where the link leads to. The text "this one special macro" is cryptic on purpose, to hide from the students what it is, but as a maintainer I want to see immediately that this link leads to the documentation about "use".

exercises/concept/dancing-dots/.docs/instructions.md
```diff
@@ -0,0 +1,77 @@
+# Instructions
+
+Your friend, an aspiring artist, reached out to you with a project idea. Let's combine his visual creativity with your technical expertise. It's time to dabble in [generative art][generative-art]!
+
+Constraints help creativity and shorten project deadlines, so you've both agreed to limit your masterpiece to a single shape - the circle. But there's going to be many many circles. And they can move around! You'll call it... dancing dots.
```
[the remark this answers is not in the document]
angelikatyborska: The word was duplicated on purpose as a way to straighten its meaning. Does that need a comma? I'm not the best with punctuation... Anyway, I'll just remove the duplication.
```suggestion
Constraints help creativity and shorten project deadlines, so you've both agreed to limit your masterpiece to a single shape - the circle. But there's going to be many circles. And they can move around! You'll call it... dancing dots.
```

exercises/concept/dancing-dots/.docs/instructions.md
```diff
@@ -0,0 +1,77 @@
...
+
+## 4. Implement the `Zoom` animation
+
+Use the `Animation` behaviour to implement a zooming animation.
+
+This animation takes one option - velocity. Velocity can be any number. If it's negative, the dot gets zoomed out instead of zoomed in.
+
+Implement the `init/1` callback. It should validate that the passed options is a keyword list with a `:velocity` key. The value of velocity must be a number. If it's not a number, return the error `"Expected required option :velocity to be a number, got: #{inspect(velocity)}"`.
+
+Implement the `handle_frame/3` callback. It should return the dot with its radius increased by the `n` times velocity, where `n` is the current frame number minus one.
```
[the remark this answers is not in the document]
angelikatyborska: Thanks! Your suggestion does sound much better.

exercises/concept/dancing-dots/.meta/exemplar.ex
jiegillet: Yes, I was suggesting two different errors, but actually a simple rephrasing might do.

If no `:velocity` option is given, the error message says "Expected required option `:velocity` to be a number", which feels like you passed something wrong, not nothing at all.

How about:
"The `:velocity` option is required, and its value must be a number. Got: nil"
It's a small change, but it feels a bit better maybe?

concepts/behaviours/.meta/config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "blurb": "Behaviours allow us to define interfaces in a behaviour module_that can be later implemented by different callback modules.",
```
jiegillet, line 2:
```suggestion
  "blurb": "Behaviours allow us to define interfaces in a behaviour module that can be later implemented by different callback modules.",
```

exercises/concept/dancing-dots/.docs/introduction.md
```diff
@@ -0,0 +1,91 @@
...
+
+  def mark_as_read(collection, book) do
+    # other function unrelated to the Countable behaviour
+  end
+end
+```
+
+### Default callback implementations
+
+When defining a behaviour, it is possible to provide a default implementation of a callbacks. The `__using__/1` macro can be used for this purpose. To make it possible for the user of the behaviour module to override the default implementation, call the `defoverridable/1` macro after the function implementation. It accepts a keyword list of function names as keys and function arities as values.
```
jiegillet, line 84: This is a bit tricky to figure out, so maybe being a bit more precise would help?

```suggestion
When defining a behaviour, it is possible to provide a default implementation of a callbacks. This implementation should be defined in the quoted expression of the `__using__/1` macro. To make it possible for users of the behaviour module to override the default implementation, call the `defoverridable/1` macro after the function implementation. It accepts a keyword list of function names as keys and function arities as values.
```

Would it be worth adding an example as well?

```elixir
defmodule Countable do
  @callback count(collection :: any) :: pos_integer
  @callback indexed_from() :: pos_integer

  defmacro __using__(_) do
    quote do
      @behaviour Countable
      def indexed_from(), do: 1
      defoverridable indexed_from: 0
    end 
  end 
end
```

exercises/concept/dancing-dots/.meta/design.md
```diff
@@ -0,0 +1,26 @@
...
+- `behaviours`
+- `use`
+
+## Prerequisites
+
+- `typespecs`
+- `structs`
+- `ast`
+- `enum`
+- `import`
```
jiegillet, line 26: Is there a need for an analyzer as well?

The only thing I can thing of is that `DancingDots.Zoom.init/1` uses `Keyword` and `is_number` but it's very minor.

exercises/concept/dancing-dots/lib/dancing_dots/dot_group.ex
jiegillet: Yes, this new version is great!

jiegillet (approved):
Awesome work 🎉 
I left some comments, but this is already good enough to merge.

exercises/concept/dancing-dots/.meta/exemplar.ex
angelikatyborska: Totally, that does sound better. Here: [`ccfcabe` (#1103)](https://github.com/exercism/elixir/pull/1103/commits/ccfcabeb06e86eda16a84d68ffea071b3b5951f4)

exercises/concept/dancing-dots/.meta/design.md
angelikatyborska: I'm not sure if those two things even matter, using the access behaviour (`opts[:velocity]`) is also fine. I guess the only "bad idea" that technically works would be pattern matching on the keyword list. As for `is_number`, we could add more tests that would assert that it also allows floats or something, but I don't think that's really necessary.

I could think of two things, relevant to the concept:

[`362ceec` (#1103)](https://github.com/exercism/elixir/pull/1103/commits/362ceeca1b02e8256a11d383a404dd26d6bd9d6f)

jiegillet: Those are better, yes :)

exercises/concept/dancing-dots/.docs/instructions.md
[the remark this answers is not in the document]
kotp: This is already in, but "there are going to be very many circles" though yes, "many, many" would be the way to strengthen the idea of "a whole bunch of" them.

PR #1104, opened 2022-04-02
Sync yacht

angelikatyborska:
One new test from https://github.com/exercism/problem-specifications/pull/1993


jiegillet (approved):
High five! 🎉

PR #1105, opened 2022-04-08
Run older Elixir with OTP 21 instead of 20

angelikatyborska:
The usage of OTP 20 is preventing the new GenServer learning exercise from being finished. See: https://github.com/exercism/elixir/pull/1076#issuecomment-1089187674


jiegillet (approved):
I'm not too sure if I should be looking at specific things, but the CI is passing so... Looks good to me!

PR #1106, opened 2022-04-10
docs(freelancer-rates): fix typo

peter-gy:
Change "a hourly rate" to "an hourly rate" in step 4 of the instructions.


jiegillet (approved):
Good catch, thank you!

PR #1107, opened 2022-04-10
Add a dancing-dots test for defoverridable

angelikatyborska:
I noticed that if you forget to make `init/1` overridable, you will only get a test failure in task 4, and it was very cryptic because it made it look like your implementation of `init/1` is wrong. Hence I want to add an explicit test in task 2 for this.


jiegillet (approved):
Good change!

PR #1110, opened 2022-04-12
Use stop-tuple instead of exit in TakeANumberDeluxe

angelikatyborska:
I'm not 100% sure if it makes a difference, but using the stop-tuple looks more elegant 😅 I tried this at first before I used `exit` but I forgot that it needs the third element, state, to work 🤦


jiegillet (approved):
Sounds good.
We will need to change the first test in the analyzer, but that can come with the full `@impl` check.

PR #1114, opened 2022-04-13
[language-list] use less suggestive function name

solnic:
Calling this function "exciting_list?" suggests that other languages aren't exciting. Small things like that matter 😄

### DIFF
--- a/exercises/concept/language-list/.docs/instructions.md
+++ b/exercises/concept/language-list/.docs/instructions.md
@@ -58,13 +58,13 @@ LanguageList.new()
 # => 2
 ```
 
-## 6. Define a function to determine if the list is exciting
+## 6. Define a function to determine if the list includes a functional language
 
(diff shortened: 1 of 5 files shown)


exercises/concept/language-list/.docs/instructions.md
```diff
@@ -60,11 +60,11 @@ LanguageList.new()
 
 ## 6. Define a function to determine if the list is exciting
```
angelikatyborska, line 61: Can you also update the this headline and its twin brother in `hints.md`?

angelikatyborska (commented):
Thank you! I agree, it's a bit arrogant that this function suggests other languages aren't exciting 😇. To be fair, other tracks implement similar exercises with their own programming language names, e.g. https://github.com/exercism/csharp/blob/main/exercises/concept/tracks-on-tracks-on-tracks/.meta/Exemplar.cs#L41-L54 But I totally agree we can be better!

exercises/concept/language-list/.docs/instructions.md
solnic: done

angelikatyborska (approved):

PR #1115, opened 2022-04-15
New practice exercise `state-of-tic-tac-toe`

jiegillet:
This is somewhat nostalgic :)

### DIFF
--- a/config.json
+++ b/config.json
@@ -2219,25 +2219,6 @@
         "difficulty": 5
       },
       {
-        "slug": "state-of-tic-tac-toe",
-        "name": "State of Tic Tac Toe",
-        "uuid": "905ac84d-7ba1-4208-bee1-55d8efde0c01",
-        "prerequisites": [
-          "strings",
-          "tuples",
-          "atoms",
-          "cond",
-          "lists",
-          "enum",
-          "maps",
-          "list-comprehensions",
-          "case",
-          "charlists"
-        ],
-        "practices": ["cond"],
-        "difficulty": 5
-      },
-      {
         "slug": "transpose",
         "name": "Transpose",
         "uuid": "ce270a34-add1-422c-bb86-53b310f05e27",
@@ -2403,6 +2384,25 @@
         "practices": [
           "recursion"
         ],
+        "difficulty": 6
+      },
+      {
+        "slug": "state-of-tic-tac-toe",
+        "name": "State of Tic Tac Toe",
+        "uuid": "905ac84d-7ba1-4208-bee1-55d8efde0c01",
+        "prerequisites": [
+          "strings",
+          "tuples",
+          "atoms",
+          "cond",
+          "lists",
+          "enum",
+          "maps",
+          "list-comprehensions",
+          "case",
+          "charlists"
+        ],
+        "practices": ["cond"],
         "difficulty": 6
       },
       {
--- a/config.json
+++ b/config.json
@@ -2230,9 +2230,11 @@
           "lists",
           "enum",
           "maps",
-          "list-comprehensions"
+          "list-comprehensions",
+          "case",
+          "charlists"
         ],
-        "practices": [],
+        "practices": ["cond"],
         "difficulty": 5
       },
       {
--- a/config.json
+++ b/config.json
@@ -2219,6 +2219,23 @@
         "difficulty": 5
       },
       {
+        "slug": "state-of-tic-tac-toe",
+        "name": "State of Tic Tac Toe",
+        "uuid": "905ac84d-7ba1-4208-bee1-55d8efde0c01",
+        "prerequisites": [
+          "strings",
+          "tuples",
+          "atoms",
+          "cond",
+          "lists",
+          "enum",
+          "maps",
+          "list-comprehensions"
+        ],
+        "practices": [],
+        "difficulty": 5
+      },
+      {
         "slug": "transpose",
         "name": "Transpose",
         "uuid": "ce270a34-add1-422c-bb86-53b310f05e27",
--- a/exercises/practice/state-of-tic-tac-toe/.meta/example.ex
+++ b/exercises/practice/state-of-tic-tac-toe/.meta/example.ex
@@ -6,8 +6,8 @@ defmodule StateOfTicTacToe do
   @doc """
   Determine the state a game of tic-tac-toe where X starts.
   """
-  @spec gamestate(board :: String.t()) :: {:ok, :win | :ongoing | :draw} | {:error, String.t()}
-  def gamestate(board) do
+  @spec game_state(board :: String.t()) :: {:ok, :win | :ongoing | :draw} | {:error, String.t()}
+  def game_state(board) do
     board =
       for {line, row} <- board |> String.split("\n", trim: true) |> Enum.with_index(),
           {player, col} <- line |> to_charlist() |> Enum.with_index(),
--- a/exercises/practice/state-of-tic-tac-toe/.meta/example.ex
+++ b/exercises/practice/state-of-tic-tac-toe/.meta/example.ex
@@ -22,12 +22,23 @@ defmodule StateOfTicTacToe do
     o_wins = wins?(board, ?O)
 
     cond do
-      x_count - o_count > 1 -> {:error, "Wrong turn order: X went twice"}
-      o_count - x_count > 0 -> {:error, "Wrong turn order: O started"}
-      x_wins and o_wins -> {:error, "Impossible board: game should have ended after the game was won"}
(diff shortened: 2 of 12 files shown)


exercises/practice/state-of-tic-tac-toe/.meta/example.ex
```diff
@@ -0,0 +1,40 @@
+defmodule StateOfTicTacToe do
```
angelikatyborska, line 1: My solution (without looking at yours first)
```elixir
defmodule StateOfTicTacToe do
  @symbol_map %{
    "X" => :x,
    "O" => :o
  }

  @doc """
  Determine the state a game of tic-tac-toe where X starts.
  """
  @spec gamestate(board :: String.t()) :: {:ok, :win | :ongoing | :draw} | {:error, String.t()}
  def gamestate(board) do
    board
    |> parse_board()
    |> check_state()
  end

  defp parse_board(board) do
    board
    |> String.split("\n", trim: true)
    |> Enum.map(fn row ->
      String.split(row, "", trim: true)
      |> Enum.map(&@symbol_map[&1])
    end)
  end

  defp check_state(parsed_board) do
    case validate_state(parsed_board) do
      :ok -> find_winners(parsed_board)
      {:error, error} -> {:error, error}
    end
  end

  defp frequencies(parsed_board) do
    plays = parsed_board |> List.flatten() |> Enum.filter(& &1)
    Map.merge(%{x: 0, o: 0}, Enum.frequencies(plays))
  end

  defp validate_state(parsed_board) do
    frequencies = frequencies(parsed_board)

    cond do
      frequencies.o > frequencies.x -> {:error, "Wrong turn order: O started"}
      frequencies.x > frequencies.o + 1 -> {:error, "Wrong turn order: X went twice"}
      true -> :ok
    end
  end

  defp find_winners(parsed_board) do
    full_board? = parsed_board |> List.flatten() |> Enum.reject(& &1) |> Kernel.==([])

    row_winners = row_winners(parsed_board)
    column_winners = column_winners(parsed_board)
    diagonal_winners = diagonal_winners(parsed_board)

    winners =
      (column_winners ++ row_winners ++ diagonal_winners)
      |> Enum.filter(& &1)
      |> Enum.uniq()

    cond do
      winners == [] && full_board? ->
        {:ok, :draw}

      winners == [] ->
        {:ok, :ongoing}

      length(winners) == 1 ->
        {:ok, :win}

      length(winners) == 2 ->
        symbol = find_who_should_have_won(parsed_board)
        {:error, "Impossible board: game should have ended after #{symbol} won"}
    end
  end

  defp row_winners(parsed_board) do
    parsed_board
    |> Enum.map(&winner(&1))
  end

  defp column_winners(parsed_board) do
    parsed_board
    |> Enum.zip()
    |> Enum.map(&winner(Tuple.to_list(&1)))
  end

  defp diagonal_winners(parsed_board) do
    down_right =
      parsed_board
      |> Enum.with_index()
      |> Enum.map(fn {_, i} -> Enum.at(Enum.at(parsed_board, i), i) end)

    up_right =
      parsed_board
      |> Enum.with_index()
      |> Enum.map(fn {_, i} -> Enum.at(Enum.at(parsed_board, length(parsed_board) - i - 1), i) end)

    [winner(down_right), winner(up_right)]
  end

  defp winner(list) do
    if length(Enum.uniq(list)) == 1, do: hd(list), else: nil
  end

  defp find_who_should_have_won(parsed_board) do
    frequencies = frequencies(parsed_board)

    # this is not technically true because it is impossible to say who should have won in certain cases
    # without knowing the actual play older, e.g.:
    #
    #   XXX
    #   OOO
    #   X..
    #
    # In this case, the winner depends on what was X's last move.
    # If it was in the first column bottom row, then X won first.
    # If it was in last column first row, then O won first.
    should_have_won =
      cond do
        frequencies.x <= frequencies.o -> :x
        frequencies.x > frequencies.o -> :o
      end

    reverted_symbol_map = Map.new(@symbol_map, fn {key, val} -> {val, key} end)
    reverted_symbol_map[should_have_won]
  end
end
```

exercises/practice/state-of-tic-tac-toe/test/state_of_tic_tac_toe_test.exs
```diff
@@ -0,0 +1,300 @@
...
+    test "Invalid board (3)" do
+      board = """
+      XXX
+      OOO
+      ...
+      """
+
+      assert StateOfTicTacToe.gamestate(board) ==
+               {:error, "Impossible board: game should have ended after X won"}
+    end
```
angelikatyborska, line 298: I am not a fan of this test. It forces you to handle a case of two winners, but more detailed tests cases are lacking because it's impossible to handle two winners correctly in all cases. It felt wrong writing code handling this... I even left a comment in my solution. Basically in this case, assuming the right round order was followed, it's clear that X won because X started. But in this case, it's no longer clear:

```
XXX
OOO
X..
```

The real winner depends on which X was played in which order. If the bottom row X was last, then X won. But it wasn't last, then O won.

I wish this corner case of "two winners" was not even touched in the problem specs but I'm not sure if it makes sense to argue. What do you think?

config.json
```diff
@@ -2218,6 +2218,23 @@
...
+        "prerequisites": [
+          "strings",
+          "tuples",
+          "atoms",
+          "cond",
+          "lists",
+          "enum",
+          "maps",
+          "list-comprehensions"
+        ],
```
angelikatyborska, line 2234: My solution included a `case`, and yours included a charlist:
```suggestion
        "prerequisites": [
          "strings",
          "tuples",
          "atoms",
          "cond",
          "lists",
          "enum",
          "maps",
          "list-comprehensions",
          "case",
          "charlists"
        ],
```

config.json
```diff
@@ -2218,6 +2218,23 @@
...
+        "practices": [],
+        "difficulty": 5
```
angelikatyborska, line 2236: What? You're saying this is easier than killer sudoku helper? 😱  It took me 4x longer to solve this than killer sudoku helper... I guess this can be a 5 or 6 if we bring killer sudoku helper down to 4, or maybe killer sudoku helper should be a 3 instead? I'm not sure... I haven't done any practice exercises in a long time.

config.json
```diff
@@ -2218,6 +2218,23 @@
...
+          "strings",
+          "tuples",
+          "atoms",
+          "cond",
+          "lists",
+          "enum",
+          "maps",
+          "list-comprehensions"
+        ],
+        "practices": [],
```
angelikatyborska, line 2235: Cond isn't full yet, what do you think?
```suggestion
        "practices": ["cond"],
```

angelikatyborska (commented):
I was really enjoying this one until the last test (not your fault) 😞

exercises/practice/state-of-tic-tac-toe/lib/state_of_tic_tac_toe.ex
```diff
@@ -0,0 +1,8 @@
+defmodule StateOfTicTacToe do
+  @doc """
+  Determine the state a game of tic-tac-toe where X starts.
+  """
+  @spec gamestate(board :: String.t()) :: {:ok, :win | :ongoing | :draw} | {:error, String.t()}
+  def gamestate(board) do
```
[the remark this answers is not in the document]
jiegillet: Yes, I guess you're right.
I never thought of choosing the right functions names as going against the problem specs, especially for something as little as an `_`. The data is important, not the variable names.

config.json
jiegillet: Haha, it didn't take me that long, I guess our initial approaches were different. I started out with the error cases and everything was smooth from there. 

Also, every time the input is a "2D" grid string, I got in the habit of turning into an "indexed" map of `{{row, col}, cell}` which really helps manipulation (1 line frequency count and 3 line `win?/2`). This pattern was born from solving so many practice exercises.

exercises/practice/state-of-tic-tac-toe/test/state_of_tic_tac_toe_test.exs
jiegillet: [Open a PR here](https://github.com/exercism/problem-specifications/pull/2012).

SaschaMann: > I was really enjoying this one until the last test (not your fault) 😞

Sorry! 😅

@jiegillet I like that change, it mirrors how I expected people to solve it (counting the number of wins, not caring about the order)

angelikatyborska (approved):

angelikatyborska:
Awesome 👏

PR #1120, opened 2022-04-21
Fix links in ABOUT.md

poulinjulien:
I noticed there is a dead link in the ABOUT page.

It's hard to tell what was at the original link, but this PR suggests another one with what seems to be the same content.


angelikatyborska (approved):
Thanks a lot! I don't really remember what the link was because I never checked it out 😅 but judging by the URL, I think you found the right video.

jiegillet:
[The wayback machine confirms it too](https://web.archive.org/web/20170407062244/http://erlangcentral.org/videos/euc-2014-robert-virding-hitchhikers-tour-of-the-beam/#.WOcwPrjLfL8).

PR #1121, opened 2022-04-22
Specify what should happen with punctuation

adamwight:


angelikatyborska:
Thank you, but we cannot accept this PR - for technical reasons. I do agree with the additions 🙂. Did you take a look at `CONTRIBUTING.md`? There is an explanation there: https://github.com/exercism/elixir/blob/609b7c2ddebe2767fffc038fd2b626c724547fde/CONTRIBUTING.md?plain=1#L43

All `instructions.md` on Exercism must be identical for all tracks. They're automatically copied from this repository: https://github.com/exercism/problem-specifications/. If you want to propose a change to instructions, you have to do it in that repository. If it gets accepted, we will automatically pull it to the Elixir track.

One exception to this rule would be changes to instructions that are caused by the Elixir track not following problem specifications exactly and implementing extra test cases, but this is not the case here. The test cases for punctuation come from problem specifications.

If you open a PR in the problem specifications repository with those changes, I'll approve them 👍

adamwight:
Thank you both for the guidance, I completely missed the fact that the exercises are generated from a language-independent repo—how cool!

PR #1129, opened 2022-05-02
Test updated for State of Tic Tac Toe: Corrects an invalid game state

chriseyre2000:
This fixes one of the tac-tac-toe games.

The game had the same number of X's and O's while X went first and won.
This is therefore an invalid game state but is used to test simple victory.

The test specifications have been updated so this matches the new version.


angelikatyborska:
Thanks! But you also need to update `tests.toml` via configlet. Take a look at this issue: https://github.com/exercism/elixir/issues/1027 (step 1 and step 3.3 apply here)

jiegillet:
There are 3 missing test cases, the 2 about invalid boards have actually been added, but I forgot to update the `tests.toml`.

jiegillet (approved):
Woah, you went above and beyond and changed all the test names as well!
Thanks a lot!

PR #1130, opened 2022-05-02
Dancing dots: sync example with instructions

danferreira:
Just a minor fix.
The error reason is different from instructions above


angelikatyborska (approved):
👍 Thank you

PR #1132, opened 2022-05-05
React: fix tests

danferreira:
The tests were passing, even with an incomplete or non existent implementation of _remove_callback_.

The callbacks were still being emitted but tests were checking for a incomplete message pattern


angelikatyborska (approved):
Thank you a lot for noticing the problem and fixing it 💜

Issue #1124, opened 2022-04-27
city-office address_map documentation test should be insensitive to key order


adamwight:
In the type documentation test [here](https://github.com/exercism/elixir/blob/1b21ae1ca610de97db79e76db890503ba75ce466/exercises/concept/city-office/test/form_test.exs#L211), we assert that the `address_map` keys are listed in a specific order, matching the order of keys in the `address_tuple`.  This should be relaxed to allow the keys in any order, by normalizing the docstring or making the test dynamic.  A standard practice is to alphabetize map keys, this should be considered correct as well.

angelikatyborska:
Sorry for the late response 🙏. I opened a PR that will fix this.

PR #1138, opened 2022-05-27
City-office: allow different order of values in `address` and `address_map` types

angelikatyborska:
Closes https://github.com/exercism/elixir/issues/1124

### DIFF
--- a/exercises/concept/city-office/test/form_test.exs
+++ b/exercises/concept/city-office/test/form_test.exs
@@ -85,15 +85,25 @@ defmodule FormTest do
           declaration == :type && type_name == unquote(type_name)
         end)
 
+      assert type,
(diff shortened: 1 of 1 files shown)


exercises/concept/city-office/test/form_test.exs
```diff
@@ -85,15 +85,25 @@ defmodule FormTest do
           declaration == :type && type_name == unquote(type_name)
         end)
 
+      assert type,
+             "expected the module #{unquote(module_name)} to have a public type named #{unquote(type_name)}"
+
```
angelikatyborska, line 90: This is an unrelated bug fix. This assertion was too late, causing a cryptic `no match of right hand side value: nil` error when a type was missing.

jiegillet (approved):
Looks good, thank you for closing the issue!

PR #1140, opened 2022-05-29
Update lasagna.ex

JVMartyns:
My solution


angelikatyborska:
The purpose of the file `exercises/concept/lasagna/lib/lasagna.ex` is to provide a starting point for all students that solve this exercise. It cannot contain a solution. The solution is in `exercises/concept/lasagna/.meta/exemplar.ex`.

PR #1141, opened 2022-05-31
Normalize type specs

adamwight:
Order doesn't matter for maps and union types, this change makes the
test insensitive by always sorting the names alphabetically.

Also makes it possible to match a single expectation explicitly.

Follows up on #1138.

### DIFF
--- a/exercises/concept/city-office/test/form_test.exs
+++ b/exercises/concept/city-office/test/form_test.exs
@@ -92,18 +92,22 @@ defmodule FormTest do
 
       {:"::", _, [_, type_definition]} = Code.Typespec.type_to_quoted(type)
 
-      actual_type_definition = Macro.to_string(type_definition)
+      # Alphabetize by label so that the test is insensitive to order.
+      normalized_type_definition =
+        case type_definition do
+          {:%{}, opts, map_keys} ->
+            {:%{}, opts, List.keysort(map_keys, 0)}
 
-      if is_list(unquote(expected_type_definition)) do
-        if actual_type_definition in unquote(expected_type_definition) do
-          assert true
-        else
-          # we know this will fail at this point, but we're using it to provide a nice failure message
-          assert actual_type_definition == hd(unquote(expected_type_definition))
+          {:|, opts, tuple_elements} ->
+            {:|, opts, Enum.sort_by(tuple_elements, fn {name, _, _} -> name end)}
+
+          x ->
+            x
         end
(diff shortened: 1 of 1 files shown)


jiegillet:
Hi @adamwight, thank you for the contribution.
Unfortunately, I'm going to turn it down, because it is mostly a refactor that students cannot see therefore benefit from and also it makes the error messages a bit worse.

While the order of keys in maps shouldn't matter, in this case we would like to encourage students to match the order of `address_tuple()`, because it makes the code more readable, so we want to show `%{street: String.t(), postal_code: String.t(), city: String.t()}` as the expected answer (while accepting other orderings silently).

It's also a bit confusing that if I write `@type address() :: map() | address_tuple()` the error message says
```
  2) test custom types in the Form module has a custom 'address' type that is a union of 'address_map' and 'address_tuple' (FormTest)
     test/form_test.exs:239
     Assertion with == failed
     code:  assert actual_type_definition == expected_type_definition
     left:  "address_tuple() | map()"
     right: "address_map() | address_tuple()"
     stacktrace:
       test/form_test.exs:242: (test)
```
but `"address_tuple() | map()"` is **not** the "actual type definition", it was modified somewhere and the student has no way of understanding that change.

jiegillet:
Hi @adamwight, thank you for the contribution.
Unfortunately, I'm going to turn it down, because it is mostly a refactor that students cannot see therefore benefit from and also it makes the error messages a bit worse.

While the order of keys in maps shouldn't matter, in this case we would like to encourage students to match the order of `address_tuple()`, because it makes the code more readable, so we want to show `%{street: String.t(), postal_code: String.t(), city: String.t()}` as the expected answer (while accepting other orderings silently).

It's also a bit confusing that if I write `@type address() :: map() | address_tuple()` the error message says
```
  2) test custom types in the Form module has a custom 'address' type that is a union of 'address_map' and 'address_tuple' (FormTest)
     test/form_test.exs:239
     Assertion with == failed
     code:  assert actual_type_definition == expected_type_definition
     left:  "address_tuple() | map()"
     right: "address_map() | address_tuple()"
     stacktrace:
       test/form_test.exs:242: (test)
```
but `"address_tuple() | map()"` is **not** the "actual type definition", it was modified somewhere and the student has no way of understanding that change.

adamwight:
@jiegillet Thanks for the review, and for sharing your thoughtful analysis!

PR #1143, opened 2022-06-06
Do not teach about Bitwise operators (^^^), only not-operator macros (bxor)

### LINKED ISSUE #1142: ^^^/2: bitwise XOR no longer exists
As of Elixir 1.12, `^^^/2: bitwise XOR` stopped existing.

Can the Secrets exercise be updated to match the latest Elixir 1.13?

angelikatyborska:
Bitwise.^^^ got deprecated, so there's no full set of operators available. Since we need a XOR, it's just easier to teach about the (not operator) macros instead. The goal of the exercise is to learn about bit manipulation in general, it shouldn't matter whether it's with operators or not.

Closes #1142

Related elixir-lang PR: https://github.com/elixir-lang/elixir/commit/d9a23d0d38ce1ada6583eca1c6ce6c861114fca7

Related elixir-lang issue: https://github.com/elixir-lang/elixir/issues/11590

> [josevalim](https://github.com/josevalim) [on 24 Jan](https://github.com/elixir-lang/elixir/issues/11590#issuecomment-1019777554)
> The operator will change precedence in Elixir v2.0.0 but not be removed. So my advice is to consider it as if it will be removed as the precedence change may break code.

### DIFF
--- a/concepts/bit-manipulation/about.md
+++ b/concepts/bit-manipulation/about.md
@@ -2,26 +2,26 @@
 
 [Bitwise binary functions][bitwise-wiki] can be performed on integers using functions from the [Bitwise module][bitwise-hexdocs].
 
-- [`&&&/2`][bitwise-and]: bitwise AND
-- [`<<</2`][bitwise-shift-left]: bitwise SHIFT LEFT
-- [`>>>/2`][bitwise-shift-right]: bitwise SHIFT RIGHT
-- [`^^^/2`][bitwise-xor]: bitwise XOR
-- [`|||/2`][bitwise-or]: bitwise OR
-- [`~~~/1`][bitwise-not]: bitwise NOT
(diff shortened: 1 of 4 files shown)


jiegillet (approved):
Good change, even though I liked the `^^^` operator, it looks like a happy alien.

I did find some uses of `&&&` in `reference/exercise-concepts/secret-handshake.md`, I don't know if we need to address that.

angelikatyborska:
> I did find some uses of &&& in reference/exercise-concepts/secret-handshake.md, I don't know if we need to address that.

We can just delete those docs: https://github.com/exercism/elixir/pull/1144

The usages of other bitwise operators in example solutions of practice exercises are fine, none of them are `^^^` and students never interact with those example solutions anyway.

PR #1144, opened 2022-06-07
Remove outdated concept reference

angelikatyborska:
It's not accurate when compared to the actual prerequisites for those exercises. The docs have been written a long time ago, before the work on v3 started.


jiegillet (approved):
Alright.
I've never interacted with those files, but they do seem updated.

PR #1145, opened 2022-06-07
change dot-dsl pratices to ast

dongdigua:
I think it should be ast :)

### DIFF
--- a/config.json
+++ b/config.json
@@ -2740,7 +2740,8 @@
           "ast"
         ],
         "practices": [
-          "ast"
+          "ast",
+          "keyword-lists"
         ],
         "difficulty": 8
       },
--- a/config.json
+++ b/config.json
@@ -2740,7 +2740,7 @@
           "ast"
         ],
         "practices": [
-          "keyword-lists"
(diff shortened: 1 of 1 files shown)


angelikatyborska:
I agree that `ast` could be added as `practices` to `dot-dsl` but I'm not sure about removing `keyword-lists`. There is no other exercise in the track that practices `keyword-lists`, a situation that should be avoided. @jiegillet You have solved this exercise before, would you agree that it's damaging to recommend it to practice `keyword-lists`? I looked at community solutions but it's still not clear to me.

jiegillet:
I also agree on adding `ast`, and keyword list are important too (to a lesser extent) so we could leave both, that's no problem.
I am a bit concerned that this is the only exercise to practice keyword lists though, because it is very advanced. Maybe we should find a couple more exercises more suited for it (not in this PR necessarily).

dongdigua:
ok, I added keyword-lists to dot-dsl.

angelikatyborska (approved):
Thank you 🎉

PR #1150, opened 2022-06-19
Use correct wording for functions

pdgonzalez872:
Hi! I saw we had some legacy terms in the docs. This fixes it.


angelikatyborska:
Oops, what a blunder! Thank you 🙂 The `lasagna` instructions definitely needs fixing, but we cannot change the `nth-prime` instructions because they're not track-specific and will be overwritten automatically ([explanation](https://github.com/exercism/elixir/blob/91ccf91940f32aff3726c772695b2de167d8192a/CONTRIBUTING.md?plain=1#L43)). Can you revert changes to the `nth-prime` file?

angelikatyborska (approved):

PR #1153, opened 2022-07-06
Dancing Dots: Fix test

LuarRaNa:
The test `"__using__ does not provide a default implementation of handle_frame/3"` was passing, even with the wrong Module.
And according to the documentation for [function_exported](https://www.erlang.org/doc/man/erlang.html#function_exported-3), it is because the Module was not loaded in which case returns `false` and the test succeeds.


angelikatyborska (approved):
Thank you 💜

PR #1155, opened 2022-07-14
File sniffer: Add instruction and test to return nil for unknown inputs

### LINKED ISSUE #1152: Analysis of "File Sniffer"
Hi, I was doing the "File Sniffer" exercise, and this was my solution:
```
defmodule FileSniffer do
  def type_from_extension("exe"), do: "application/octet-stream"
  def type_from_extension("bmp"), do: "image/bmp"
  def type_from_extension("png"), do: "image/png"
  def type_from_extension("jpg"), do: "image/jpg"
  def type_from_extension("gif"), do: "image/gif"

  def type_from_binary(<<0x7F, 0x45, 0x4C, 0x46, _rest::binary>>), do: "application/octet-stream"
  def type_from_binary(<<0x42, 0x4D, _rest::binary>>), do: "image/bmp"
  def type_from_binary(<<0x89, 0x50, 0x4E, 0x47, 0x0D, 0x0A, 0x1A, _rest::binary>>), do: "image/png"
  def type_from_binary(<<0xFF, 0xD8, 0xFF, _rest::binary>>), do: "image/jpg"
  def type_from_binary(<<0x47, 0x49, 0x46, _rest::binary>>), do: "image/gif"

  def verify(file_binary, extension) do
    case {type_from_binary(file_binary), type_from_extension(extension)} do
      {type, type} -> {:ok, type}
      _ -> {:error, "Warning, file format and file extension do not match."}
    end
  end
end
```
But for some reason it still gave me this feedback:
> Essential
Even though there could be other valid ways to solve this exercise, for the sake of mastering bitstrings please match the first few bytes of the argument of type_from_binary/1 with the special form <<>>.

I tried solving for the problem myself, but couldn't find what was actually causing it.

Lgdev07:
Closes #1152 and https://github.com/exercism/elixir-analyzer/issues/304

The instructions for the exercise say that the student needs to implement type_from_binary/1 by waiting for the binary signature.

The problem is that we only have tests for this in the analyzer.

So if the student creates type_from_binary/1 expecting only the first hex, it will pass the local tests, but fail in the analyzer tests.

This PR creates tests to ensure that the program will fail if the file does not have a complete binary signature.

### DIFF
--- a/exercises/concept/file-sniffer/.docs/instructions.md
+++ b/exercises/concept/file-sniffer/.docs/instructions.md
@@ -14,21 +14,28 @@ Use the following table for reference:
 
 ## 1. Given an extension, return the expected media type
 
-Implement the `type_from_extension/1` function. It should take a file extension (string) and return the media type (string).
+Implement the `type_from_extension/1` function. It should take a file extension (string) and return the media type (string) or nil if the extension does not match with the expected ones.
 
 ```elixir
 FileSniffer.type_from_extension("exe")
 # => "application/octet-stream"
+
+FileSniffer.type_from_extension("txt")
+# => nil
 ```
 
 ## 2. Given a binary file, return the expected media type
 
-Implement the `type_from_binary/1` function. It should take a file (binary) and return the media type (string).
+Implement the `type_from_binary/1` function. It should take a file (binary) and return the media type (string) or nil if the extension does not match with the expected ones.
 
 ```elixir
 file = File.read!("application.exe")
 FileSniffer.type_from_binary(file)
 # => "application/octet-stream"
+
+file = File.read!("example.txt")
+FileSniffer.type_from_binary(file)
+# => nil
 ```
 
 Don't worry about reading the file as a binary. Assume that has been done for you and is provided by the tests as an argument.
--- a/exercises/concept/file-sniffer/.meta/exemplar.ex
+++ b/exercises/concept/file-sniffer/.meta/exemplar.ex
@@ -4,6 +4,7 @@ defmodule FileSniffer do
   def type_from_extension("jpg"), do: "image/jpg"
   def type_from_extension("gif"), do: "image/gif"
   def type_from_extension("exe"), do: "application/octet-stream"
+  def type_from_extension(_file_extension), do: nil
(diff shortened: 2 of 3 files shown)


angelikatyborska:
Big thanks, this is an excellent addition 🙏.

Unfortunately I am not sure if we can use `assert_raise FunctionClauseError`. This exercise is about learning binaries and doesn't force you to use multiple clause functions. You might just as well use `cond`, and then the error will be different:

```
  5) test raises error when given uncompleted signature file exe (FileSnifferTest)
     test/file_sniffer_test.exs:102
     Expected exception FunctionClauseError but got CaseClauseError (no case clause matching: "\dEL")
     code: assert_raise(FunctionClauseError, fn ->
     stacktrace:
       (file_sniffer 0.1.0) lib/file_sniffer.ex:9: FileSniffer.type_from_binary/1
       test/file_sniffer_test.exs:103: (test)
```

But this test case is very valuable because multiple people already complained about the analyzer being wrong because of this 🤔.

I'm thinking that we should define the behavior for unknown inputs in `type_from_extension` and `type_from_binary`. The concept `nil` was already taught by this point in the learning journey, so we can just define that for any other input than the ones defined in the exercise, the functions should return `nil`. Then, the new tests can just assert that the result is `nil` instead of that it raises an error. What do you think?

Lgdev07:
Thanks for the review :)

Yeah, I didn't think in this scenario that the student could use cond, changing the instructions is actually a better approach.

I will make the changes,

angelikatyborska:
> I will make the changes

@Lgdev07 Are you still going to do that, or should we take over and finish this task for you?

Lgdev07:
> @Lgdev07 Are you still going to do that, or should we take over and finish this task for you?

Sorry for the delay Angelika, I'll finish by the end of this weekend for sure, I moved to another city and was very busy :)

angelikatyborska:
Oh, that's very exciting 🏙️ ✨. I also moved apartments recently, so I remember very well how busy that makes you 🚚 . No worries.

angelikatyborska (approved):
Thank you!

PR #1160, opened 2022-07-25
Fix typespec in meetup.ex

deadshot465:
`:calendar.date()` in Erlang is defined as `{year(), month(), day()}` tuple, whereas `Date.t()` and `Calendar.date()` in Elixir are maps. The function `assert_dates_equal` in `meetup_test.exs` expects the first argument to be a `%Date{}`, which doesn't match the typespec in `meetup.ex` (and Dialyzer will issue a warning when returning a `Date.t()` as well). This tiny PR fixes this.


angelikatyborska (approved):
Thank you!

PR #1163, opened 2022-07-30
Fix links in log-parser admonition

angelikatyborska:
Relevant issue: https://github.com/exercism/exercism/issues/6347

I can also see that I forgot to put the references for the URLs at the bottom of one of those documents, but that's not the cause of this issue. The links don't render on the concept document either, and that has the right references, so I'm sure this PR is necessary.


jiegillet (approved):
👍

Issue #1171, opened 2022-08-12
Rewrite `language-list` examples in instructions without using the pipe


angelikatyborska:
The pipe operator is only taught later in the concept tree (at `strings` (high school sweetheart) and shouldn't be used in the code examples in exercises that come before. Please rewrite those examples without the pipe.

https://github.com/exercism/elixir/blob/main/exercises/concept/language-list/.docs/instructions.md?plain=1#L20-L21

PR #1172, opened 2022-08-12
Remove pipe operator from language-list instructions

ghecho:
Update instructions so they don't use the pipe operator since it's taught later.

Resolves https://github.com/exercism/elixir/issues/1171


angelikatyborska (approved):
Perfect, thank you!

PR #1176, opened 2022-08-13
Fix the first test which is set as pending for binary_search_test

meatball133:
This is a fix for the binary search exercises


angelikatyborska (approved):
Thanks!

PR #1177, opened 2022-08-14
Remove outdated language-list hints

angelikatyborska:
Same as https://github.com/exercism/elixir/pull/1173, pointed out by Exhort students 🙂


jiegillet (approved):
Thank you, Exhort students

PR #1181, opened 2022-08-22
Incorrect sentence in the Introduction

Morsicus:
I think the sentence is unclear and look wrong to me. Here a proposal. What do you think?

### DIFF
--- a/concepts/lists/introduction.md
+++ b/concepts/lists/introduction.md
@@ -9,7 +9,7 @@ two_item_list = [1, 2]
 multiple_type_list = [1, :pi, 3.14, "four"]
 ```
 
-Elixir implements lists as a linked list, where each node stores the reference to the next list. The first item in the list is referred to as the _head_ and the remaining list of items is called the _tail_. We can use this notation in code:
+Elixir implements lists as a linked list, where each node stores the reference to the next element in the list. The first item in the list is referred to as the _head_ and the remaining list of items is called the _tail_. We can use this notation in code:
 
 ```elixir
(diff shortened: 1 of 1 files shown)


concepts/lists/introduction.md
```diff
@@ -9,7 +9,7 @@ two_item_list = [1, 2]
 multiple_type_list = [1, :pi, 3.14, "four"]
 ```
 
-Elixir implements lists as a linked list, where each node stores the reference to the next list. The first item in the list is referred to as the _head_ and the remaining list of items is called the _tail_. We can use this notation in code:
+Elixir implements lists as a linked list, where each node stores the reference to the next element in the list. The first item in the list is referred to as the _head_ and the remaining list of items is called the _tail_. We can use this notation in code:
```
angelikatyborska, line 12: Thanks! Indeed that's a confusing sentence. Your interpretation of it isn't really correct, that's on us 😞.

Here's my fix:
```suggestion
Elixir implements lists as a linked list, where each node stores two values: the first item and another list with all the remaining items. The first item in the list is referred to as the _head_ and the remaining list of items is called the _tail_. We can use this notation in code:
```

jiegillet: I'm good with your proposal, @angelikatyborska. 
Does this make it clearer to you @Morsicus?

angelikatyborska (approved):

jiegillet (approved):

angelikatyborska:
Thank you!

PR #1182, opened 2022-08-23
Add more test cases to german-sysadmin

pdmoore:
- Added a simpler test for lowercase retention that matches the example from the README
- Renamed the test that keeps lowercase to indicate it also keeps Latin and German
- Switched the order of the asserts so you get the information that the counts do not match before the strings are compared (this would have helped me a lot)
- Used the 'allows' language from the README in the underscores test

### DIFF
--- a/exercises/concept/german-sysadmin/test/username_test.exs
+++ b/exercises/concept/german-sysadmin/test/username_test.exs
@@ -8,18 +8,38 @@ defmodule UsernameTest do
     end
 
     @tag task_id: 1
-    test "it removes everything but lowercase letters" do
+    test "it allows lowercase Latin letters" do
+      assert Username.sanitize('anne') == 'anne'
+    end
+
+    @tag task_id: 1
+    test "it allows the whole lowercase Latin alphabet" do
+      lowercase_latin_letters = 'abcdefghijklmnopqrstuvwxyz'
+
+      assert Username.sanitize(lowercase_latin_letters) == lowercase_latin_letters
+    end
+
+    @tag task_id: 1
+    test "it removes numbers" do
       assert Username.sanitize('schmidt1985') == 'schmidt'
     end
 
     @tag task_id: 1
-    test "it keeps all lowercase Latin letters (and ignores German letters that require special handling)" do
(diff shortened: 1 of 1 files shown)


angelikatyborska:
Thank you for the suggestions!

I agree that the old test descriptions aren't that great and that we need simpler tests. A lot of students struggle with charlists. What if we added even more simple tests for the first task? And the big test with a very long input is very tricky... on one hand, we need a test that checks that everything that should be deleted, is deleted. On other hand, we cannot mention in the first task that underscores and German letters are special. That comes in later tasks. Maybe I can rewrite that too, hmm... 

Here are all the tests for task 1 that I would suggest having. If you like that suggestion, you can copy-paste it to your PR and I'll approve (you'll get the reputation points on Exercism 😉). 

```elixir
    @tag task_id: 1
    test "works for an empty charlist" do
      assert Username.sanitize('') == ''
    end

    @tag task_id: 1
    test "it allows lowercase latin letters" do
      assert Username.sanitize('anne') == 'anne'
    end

    @tag task_id: 1
    test "it allows the whole lowercase latin alphabet" do
      lowercase_latin_letters = 'abcdefghijklmnopqrstuvwxyz'

      assert Username.sanitize(lowercase_latin_letters) == lowercase_latin_letters
    end

    @tag task_id: 1
    test "it removes numbers" do
      assert Username.sanitize('schmidt1985') == 'schmidt'
    end

    @tag task_id: 1
    test "it removes punctuation" do
      assert Username.sanitize('*fritz*!$%') == 'fritz'
    end

    @tag task_id: 1
    test "it removes whitespace" do
      assert Username.sanitize(' olaf ') == 'olaf'
    end

    @tag task_id: 1
    test "it removes all disallowed characters" do
      allowed_characters = 'abcdefghijklmnopqrstuvwxyz_ßäöü'
      input = Enum.to_list(0..0x10FFFF) -- allowed_characters

      assert Username.sanitize(input) == ''
    end
```

pdmoore:
Thanks, Angelika! I've updated the pull request with a new commit (I think) after I confirmed all new tests pass the existing implementation locally. There should be 9 tests now.

angelikatyborska (approved):
Thank you 💜

PR #1183, opened 2022-08-24
Make newsletter exercise tests run async

halfdan:
I just went through this exercise as part of the Exhort. It seemed trivial to make this exercise run tests asynchronous by generating random temp files using some core utils.

### DIFF
--- a/exercises/concept/newsletter/test/newsletter_test.exs
+++ b/exercises/concept/newsletter/test/newsletter_test.exs
@@ -1,6 +1,5 @@
 defmodule NewsletterTest do
-  # run test synchronously to be able to use the same file path for all tests without write conflicts
-  use ExUnit.Case, async: false
+  use ExUnit.Case
 
   @temp_file_path Path.join(["assets", "temp.txt"])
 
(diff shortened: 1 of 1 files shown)


zorn:
My understanding is that marking a test module as `async` only enables concurrent testing against other modules. The individual tests of a single module always run sequentially. 🤷‍♂️ 

https://hexdocs.pm/ex_unit/main/ExUnit.Case.html

angelikatyborska:
@zorn You're absolutely right. This existing code with the comment in the tests should never have been added:

```elixir
  # run test synchronously to be able to use the same file path for all tests without write conflicts
  use ExUnit.Case, async: false
```

Our bad 🙈. But that also means this PR isn't really needed. We just need to remove the snippet I posted above to avoid future confusion.

halfdan:
Hah, well TIL.

Happy to update/change the PR to remove that little snippet. Thanks to both of you.

angelikatyborska:
> Happy to update/change the PR to remove that little snippet

That would be much appreciated!

jiegillet:
@halfdan are you still interested in finishing this PR?
We can add the hacktoberfest label to it if you're interested.

halfdan:
Yep, still happy to finish this. The tag would be a nice perk.

jiegillet (approved):
Sorry I took over in the end. Thanks for bringing this up though!

halfdan:
@jiegillet Perfectly fine - sorry for dropping the ball here.

PR #1184, opened 2022-08-31
Scale Generator: Fix return type in typespec

aifrak:
Improve typespec from `step/3` because it returns String and not a list.


angelikatyborska:
Thanks! Could you add the same fix to the example solution in `.meta/example.ex`?

aifrak:
Done: `.meta/example.ex` has been also fixed.

angelikatyborska (approved):

angelikatyborska:
Thank you 💜

PR #1185, opened 2022-09-01
Update high_score_test.exs

ingluisandres:
fixes update score for existing player with non-zero score adds score to previous


angelikatyborska:
The current test is correct. It is desired behavior that `update_score` adds to the previous score if it exists.

PR #1186, opened 2022-09-02
Clarify wording in Secrets instructions

### LINKED ISSUE #1170: Could instructions for Secrets subtract be clearer?
I may be wrong on this, but it caught out me, Brian and Erik, so I think I'm not.

---

The [example for subtract](https://github.com/exercism/elixir/blob/main/exercises/concept/secrets/.docs/instructions.md#2-create-a-subtractor) is:
```
subtractor = Secrets.secret_subtract(2)
subtractor.(3)
# => 1
```

And the [tests are](https://github.com/exercism/elixir/blob/main/exercises/concept/secrets/test/secrets_test.exs#L18-L37):
```
subtract = Secrets.secret_subtract(3)
subtract.(6) 
=> 3
```

However, the [instructions](https://github.com/exercism/elixir/blob/main/exercises/concept/secrets/.docs/instructions.md#2-create-a-subtractor) are: "It should return a function which takes one argument and subtracts from it the secret passed in to secret_subtract."

The functions above are subtracting the secret from the argument, not the argument from the secret. [My passing solution](https://exercism.org/tracks/elixir/exercises/secrets/solutions/iHiD) subtracts the secret from the argument too.

kytrinyx:
The instructions for the Secrets concept exercise has left people
confused with respect to the `secret_subtract` method.

People were not sure whether they should subtract the argument to
the anonymous function from the argument to the named function, or
vice versa.

After some discussion on GitHub the conclusion was that the phrase
_subtracts from it_ was at the root of the confusion, as
people were reversing the order of _it_ and _from_ without noticing
it.

This rephrases the sentence in an attempt to clarify.

Closes #1170

### DIFF
--- a/exercises/concept/secrets/.docs/instructions.md
+++ b/exercises/concept/secrets/.docs/instructions.md
@@ -18,7 +18,7 @@ adder.(2)
 
 ## 2. Create a subtractor
 
-Implement `Secrets.secret_subtract/1`. It should return a function which takes one argument and subtracts from it the secret passed in to `secret_subtract`.
+Implement `Secrets.secret_subtract/1`. It should return a function which takes one argument and subtracts the secret passed in to `secret_subtract` from that argument.
 
 ```elixir
 subtractor = Secrets.secret_subtract(2)


angelikatyborska (approved):

angelikatyborska:
Thank you!

PR #1187, opened 2022-09-02
Fix mixup between arguments and parameters

kytrinyx:
Arguments are passed to a function, but a function definition has
parameters.

Here we were referring to both concepts with the same word.

### DIFF
--- a/concepts/basics/introduction.md
+++ b/concepts/basics/introduction.md
@@ -37,7 +37,7 @@ defmodule Calculator do
 end
 ```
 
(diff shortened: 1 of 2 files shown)


angelikatyborska:
There is no concept of "parameters" in the Elixir world, except when maybe used colloquially. Functions have arguments and we try to avoid the word "parameter" to make it very clear it should not be used to mean "arguments". Can this sentence be edited in a way that makes more sense to you but doesn't use the word "parameter"?

kytrinyx:
Oh wow! That's really interesting, and good to know.

If that's the case then there's no change necessary 😄  ❤️ Thank you

PR #1192, opened 2022-09-02
Update bitwise shift left example

### LINKED ISSUE #1189: Should we tweak the bitwise example in Secrets introduction?
The introduction says:

> Here is an example how to use a bitwise function:

The example it gives is:

```
Bitwise.bsl(1, 2)
# => 4
```

However, it's not immediately clear which number is being shifted and which number represents how many places it is shifted, because `Bitwise.bsl(1, 2)` and `Bitwise.bsl(2, 1)` both result in 4.

Would it make sense to use a different value in the example to help clarify? E.g.

```
Bitwise.bsl(1, 3)
# => 8
```

If this seems like a good change, I'd be happy to put in a PR.

kytrinyx:
The introduction to bitwise shift used `Bitwise.bsl(1, 2)`
as the example. This is ambiguous, as
`Bitwise.bsl(1, 2)` and `Bitwise.bsl(2, 1)` both result in 4.

Here I updated the example to shift 1 left 3 places, resulting
in a value of 8. This is unambiguous because 3 shifted left once
results in a value of 6.

Closes #1189

### DIFF
--- a/concepts/bit-manipulation/introduction.md
+++ b/concepts/bit-manipulation/introduction.md
@@ -13,7 +13,7 @@ Here is an example how to use a bitwise function:
 
 ```elixir
 Bitwise.bsl(1, 3)
-# => 3
+# => 8
 ```
 
 All bitwise functions only work on integers.
--- a/concepts/bit-manipulation/introduction.md
+++ b/concepts/bit-manipulation/introduction.md
@@ -12,8 +12,8 @@ Elixir supports many functions for working with bits found in the _Bitwise modul
 Here is an example how to use a bitwise function:
 
 ```elixir
-Bitwise.bsl(1, 2)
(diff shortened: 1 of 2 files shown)


concepts/bit-manipulation/introduction.md
```diff
@@ -12,8 +12,8 @@ Elixir supports many functions for working with bits found in the _Bitwise modul
 Here is an example how to use a bitwise function:
 
 ```elixir
-Bitwise.bsl(1, 2)
-# => 4
+Bitwise.bsl(1, 3)
+# => 3
```
angelikatyborska, line 16:
```suggestion
# => 8
```

angelikatyborska (approved):
Thanks!

PR #1195, opened 2022-09-04
Community garden: require incremental plot ids

angelikatyborska:
This change is inspired by the Exhort students discussing how they were unsure how to handle the id generation. One student used random ids, another created a second Agent. I think the id generation isn't supposed to be a riddle in this exercise, so we might as well make our expected solution explicit - ids should be incremental, you're supposed to keep an id counter in the agent's state.


tim-austin-7shifts (approved):

angelikatyborska:
@tim-austin-7shifts @neenjaw I appreciate the approval but you seem to have used your second account which doesn't have maintainer permissions 😅

neenjaw:
Lol, was using the wrong laptop

neenjaw (approved):
Approved for real this time. :D

PR #1196, opened 2022-09-05
Update captains_log_test.exs

pdmoore:
Combine stardate range checks into one test with two asserts. Currently, as two separate tests, the first test passes (>= 41_000.0) with a nil implementation. This new combined test follows the pattern already demonstrated in this test file as "ends with a random integer between 1000 and 9999"

### DIFF
--- a/exercises/concept/captains-log/test/captains_log_test.exs
+++ b/exercises/concept/captains-log/test/captains_log_test.exs
@@ -59,15 +59,9 @@ defmodule CaptainsLogTest do
     end
 
     @tag task_id: 3
-    test "is equal to or greater than 41_000.0" do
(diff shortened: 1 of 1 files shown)


neenjaw:
Appreciate the initiative for a pr, but this makes the test suite worse in my opinion as it then combines two behaviours into one test making it harder for more novice students to see that there are two cases to handle.

If we can reword the existing two test cases to make it more clear for you, I am open to that. 

In its current state I would not merge this PR.

pdmoore:
I felt there's inconsistency in the suite. The test for the behavior "Registry numbers start with the prefix 'NCC-' and then use a number from 1000 to 9999 (inclusive)" is verified by one test that confirms the range 1000..9999 through two asserts in the one test.

PR #1199, opened 2022-09-12
Update instructions.md for roman-numerals and added two new tests

tradfursten:
Updated the instructions with the sync script.


angelikatyborska (approved):
Thank you!

PR #1200, opened 2022-09-12
Fix-test-case-and-instructions-of-high-score-exercise

lacf95:
Found a failing test and a wrong instruction's example output on High Score exercise, this PR fixes it


lacf95:
Not really, closed!

PR #1206, opened 2022-09-15
Pacman rules: Added missing testcase

lauramcastro:
The first function of the Pacman exercise is intended to be solved by using the `and` operator. However, with the test for double `false` missing, an implementation that used `==` instead passed the tests and was uncaught by automated feedback.

I realized this during a mentoring session; perhaps there is a preferred course of action, but I thought I could contribute the missing test case, just in case.


angelikatyborska (approved):
Thank you, this additional test makes a lot of sense 🙂

PR #1208, opened 2022-09-21
Added assertion to bread-and-potions test

nash8114:
* ensure that the bread-and-potion ManaPotion strength attribute is used, and not just a value of 6.
I ran into this by just staticky using "6" as a value for all mana potions, and both tests happened to pass. (Any other value would fail). This small addition should fix being able to use a static value for mana.


jiegillet (approved):
Thank you, this is an nice improvement!

PR #1211, opened 2022-10-03
Add a link to hex docs for match operator (=/2)

rlamsal1256:


jiegillet:
Thank you for the contribution, unfortunately I will turn it down.
The [Exercism docs](https://exercism.org/docs/building/tracks/concepts) mention about `introduction.md` that

> Links should be used sparingly, if at all. While a link explaining a complex topic like recursion might be useful, for most concepts the links will provide more information than needed so explaining things concisely inline should be the aim.

The link is more appropriate for `about.md`, and actually is already present there.
I appreciate the PR but in cases like this we like to follow the Exercism guidelines.

PR #1217, opened 2022-10-20
Update bin/fetch-configlet script

kytrinyx:
The script has been tweaked to work better across all platforms.

Ref https://github.com/exercism/exercism/issues/6591


ErikSchierboom (approved):

jiegillet:
I didn't notice the name was different, my bad.
I'm not sure why we had a different name, but personally, I'm in favor of forcing all tracks to have the same name, it's just easier to centralize these things.

PR #1219, opened 2022-10-21
RPN Calculator: two additional test cases so that a naïve implementation cannot pass the test suite

lauramcastro:
Got a student which implementation was passing all tests, but whose implementation was not actually using the result of `operation.(stack)` at any time. Instead, it was manually returning `:ok` and `{:ok, "operation completed"}`.

This was passing the test suite because only one `operation` was being used as test case for either `calculate!/2`, `calculate/2` and `calculate_verbose/2`. Simply adding another test case that uses a different `operation` function should prevent such naïve implementation from passing all the tests.

### DIFF
--- a/exercises/concept/rpn-calculator/test/rpn_calculator_test.exs
+++ b/exercises/concept/rpn-calculator/test/rpn_calculator_test.exs
@@ -2,8 +2,13 @@ defmodule RPNCalculatorTest do
   use ExUnit.Case
 
   @tag task_id: 1
-  test "calculate! returns an :ok atom" do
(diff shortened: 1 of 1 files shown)


glennj:
> ... was not actually using the result of `operation.(stack)` at any time. Instead, it was manually returning `:ok` and `{:ok, "operation completed"}`.

Test-driven development at its most pure! ;)

jiegillet (approved):
Hahaha yes, there are such students.
Thank you for the extra tests. If you feel like it, you are welcome to add your name to the lists of contributors of the exercise.

PR #1223, opened 2022-10-27
Add missing spaces in `input` of `rectangles`

aifrak:
The test `rectangles must have four sides` from [problem-specifications](https://github.com/exercism/problem-specifications/blob/main/exercises/rectangles/canonical-data.json) has 2 trailing spaces on the 4th line.

Spaces have been added like in the test `two rectangles without shared parts` to avoid the formatter to remove them.


jiegillet (approved):
Thank you very much 💛

PR #1226, opened 2022-11-06
Update high_score_test.exs

GDM99:
It could be either:
@tag task_id: 5
    test "update score for existing player with non-zero score adds score to previous" do
      scores =
        HighScore.new()
        |> HighScore.add_player("José Valim")
        |> HighScore.update_score("José Valim", 1)
        |> HighScore.update_score("José Valim", 486_373)

      assert scores == %{"José Valim" => 486_373}
    end
  end

OR

@tag task_id: 5
    test "update score for existing player with non-zero score adds score to previous" do
      scores =
        HighScore.new()
        |> HighScore.add_player("José Valim")
        |> HighScore.update_score("José Valim", 1)
        |> HighScore.update_score("José Valim", 486_373)
        |> HighScore.update_score("José Valim", 486_374)
      assert scores == %{"José Valim" => 486_374}
    end
  end

### DIFF
--- a/exercises/concept/high-score/test/high_score_test.exs
+++ b/exercises/concept/high-score/test/high_score_test.exs
@@ -139,7 +139,7 @@ defmodule HighScoreTest do
         |> HighScore.update_score("José Valim", 1)
         |> HighScore.update_score("José Valim", 486_373)
 
-      assert scores == %{"José Valim" => 486_374}
+      assert scores == %{"José Valim" => 486_373}
     end
   end
 


jiegillet:
Hi @GDM99,
General pieces of advice:
It's usually preferable to open a PR with an explanation of the change instead of dumping code in there, especially when there is no issue associated with it.
When you want to make a change, please do some research in issues and PR to see if it came up before.

In this case, it did come up before (I think more than once).
The instructions mention 
> - The third argument is the score that you wish to **add** to the stored high score.
So the test is correct.

PR #1229, opened 2022-11-09
[Fixes #1209] Captain's Log instructions change

### LINKED ISSUE #1209: Captain's Log instructions should nudge the student towards the erlang call requirement.
https://github.com/exercism/elixir/blob/be6101a052f015a8fd74e2fe233d0bb1af1b0543/exercises/concept/captains-log/.docs/instructions.md?plain=1#L44-L53

I solved this early on with the following but then was dumbfounded why the test would be specifically looking for an `ArgumentError`. When I attempted to add guards I got `FunctionClauseError` which of course did not match.

```elixir
  def format_stardate(stardate) do
    Float.round(stardate, 1) |> Float.to_string()
  end
```

**Suggestion**

I suggest we add a line to the instructions hinting that solving this using an erlang call is a requirement.

zorn:
Fixes #1209

Adds explicit prose and error example to help express what the current tests are looking for.

### DIFF
--- a/exercises/concept/captains-log/.docs/instructions.md
+++ b/exercises/concept/captains-log/.docs/instructions.md
@@ -47,7 +47,7 @@ CaptainsLog.random_stardate()
 
 In the captain's log, stardates are usually rounded to a single decimal place.
 
-Implement the `format_stardate/1` function that will take a floating point number and return a string with the number rounded to a single decimal place. 
+Implement the `format_stardate/1` function that will take a floating point number and return a string with the number rounded to a single decimal place.
 
 The implementation should use an Erlang function.
 
--- a/exercises/concept/captains-log/.docs/instructions.md
+++ b/exercises/concept/captains-log/.docs/instructions.md
@@ -47,7 +47,12 @@ In the captain's log, stardates are usually rounded to a single decimal place.
 
 Implement the `format_stardate/1` function that will take a floating point number and return a string with the number rounded to a single decimal place.
 
+While there are valid approaches to this solution that might use an Elixir module, the implementation we are testing for expects an implementation using an Erlang function.
+
 ```elixir
 CaptainsLog.format_stardate(41458.15721310934)
 # => "41458.2"
+
+CaptainsLog.format_stardate(41411)
+# => ** (ArgumentError) argument error
 ```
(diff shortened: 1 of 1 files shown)


exercises/concept/captains-log/.docs/instructions.md
```diff
@@ -47,7 +47,12 @@ In the captain's log, stardates are usually rounded to a single decimal place.
...
 Implement the `format_stardate/1` function that will take a floating point number and return a string with the number rounded to a single decimal place.
 
+While there are valid approaches to this solution that might use an Elixir module, the implementation we are testing for expects an implementation using an Erlang function.
+
 ```elixir
 CaptainsLog.format_stardate(41458.15721310934)
 # => "41458.2"
+
+CaptainsLog.format_stardate(41411)
+# => ** (ArgumentError) argument error
```
jiegillet, line 57: I appreciate where this coming from, but something we see often when we make those things too explicit is that a surprising number of students will implement those explicitly

```elixir
def format_stardate(n) when is_integer(n) do
  raise ArgumentError
end
```
which is not idiomatic and not easy to test for (although it could be an analyzer check).
Another point of view is that you wouldn't point out that `format_stardate("hi")` is an argument error.

So I think I would rather leave that out.

exercises/concept/captains-log/.docs/instructions.md
```diff
@@ -47,7 +47,12 @@ In the captain's log, stardates are usually rounded to a single decimal place.
 
 Implement the `format_stardate/1` function that will take a floating point number and return a string with the number rounded to a single decimal place.
 
+While there are valid approaches to this solution that might use an Elixir module, the implementation we are testing for expects an implementation using an Erlang function.
```
jiegillet, line 50: This is well suited for an analyzer comment, but is a bit long for here, and the motivation for doing that is not for testing, it's for learning to use Erlang, so we shouldn't mention tests.
 
Like I suggested, "The implementation should use an Erlang function." or something like that would be best.

The short sentence can also go to task 3. I saw your comment on the issue, but we have been talking about changing the analyzer check [over here](https://github.com/exercism/elixir-analyzer/issues/340) to check for an Erlang function, so the requirement should be explicit.

exercises/concept/captains-log/.docs/instructions.md
```diff
@@ -45,7 +47,9 @@ CaptainsLog.random_stardate()
 
 In the captain's log, stardates are usually rounded to a single decimal place.
 
-Implement the `format_stardate/1` function that will take a floating point number and return a string with the number rounded to a single decimal place.
+Implement the `format_stardate/1` function that will take a floating point number and return a string with the number rounded to a single decimal place.
```
jiegillet, line 50:
```suggestion
Implement the `format_stardate/1` function that will take a floating point number and return a string with the number rounded to a single decimal place.
```

jiegillet (approved):
Thank you very much for your understanding and contribution, I appreciate it.

PR #1230, opened 2022-11-17
Change refute to assert equals false 

mrnovalles:
When coding the exercise, the instructions are clear that alert_recipient/2 should return a `false` value in case none of the conditions match. Returnnig a nil would still make the tests pass.
This commit changes that to use an assertion for the `false` value instead.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,6 +97,7 @@ jobs:
         with:
           otp-version: ${{matrix.otp}}
           elixir-version: ${{matrix.elixir}}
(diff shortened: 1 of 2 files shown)


jiegillet:
Thank you, that's a nice change.
Could I ask that you flip the order
```diff
- assert false == LogLevel.alert_recipient(0, false)
+ assert LogLevel.alert_recipient(0, false) == false
```
This is the usual convention in Elixir tests (or at least in this repo).

jiegillet (approved):

PR #1237, opened 2022-11-18
README: reflect renaming of `fetch-configlet`

ee7:
I forgot to include this in 213ef5b3860b. Sorry for the noise.


jiegillet (approved):
Hahaha if I didn't know better, I would report you for spamming :p 
Thanks for all the effort.

PR #1238, opened 2022-11-18
Update (a bit) text formatting and spec for Kindergarten Garden

cr0t:
Just a minor stylistic updates to the Kindergarten Garden task main file.

_Something that I noticed when opened this file._


jiegillet:
Thank you, but we are not really looking for contributions for matters so small.
Also, `String.t()` is the norm, I don't think we need to change that.

PR #1242, opened 2022-11-26
Fix erlef/setup-beam not finding Erlang versions

angelikatyborska:
I noticed that the CI was failing on a PR that doesn't touch any Elixir code: https://github.com/exercism/elixir/pull/1241

It was suspicious because it started failing without any changes. I assume github changed what "ubuntu-latest" means.

I chose a lower Ubuntu version according to this table: https://github.com/erlef/setup-beam#compatibility-between-operating-system-and-erlangotp

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ on:
 
 jobs:
   precheck:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


jiegillet (approved):
I made the same change in [a9ec12d](https://github.com/exercism/elixir/pull/1230/commits/a9ec12d06d727a917d369cc29b7017c782a0abc9) but it didn't take.
Maybe because I missed pr.ci.yml 🤷

PR #1246, opened 2022-12-13
Add test cases for matching_brackets

pichi:
There are some solutions using recursive regular expressions {PCRE containing (?R) or (?n) } which could fail those tests despite passing all previous tests.

### DIFF
--- a/.github/workflows/pr-comment.yml
+++ b/.github/workflows/pr-comment.yml
@@ -6,7 +6,7 @@ jobs:
   pr-comment:
     runs-on: ubuntu-latest
     steps:
-      - uses: exercism/pr-commenter-action@v1.3.0
+      - uses: exercism/pr-commenter-action@v1.3.1
         with:
           github-token: "${{ github.token }}"
           config-file: ".github/pr-commenter.yml"


jiegillet:
Hi @pichi, thank you for the contribution.
All of our tests cases are synced from the [problem-specifications](https://github.com/exercism/problem-specifications) central repository, we add test cases whenever they are added in the exercise [canonical data](https://github.com/exercism/problem-specifications/blob/main/exercises/matching-brackets/canonical-data.json).

Normally, I would redirect you to that repo, but currently most of Exercism is on a paused state, so all contributions from non-maintainers are closed automatically [with this message](https://github.com/exercism/problem-specifications/pull/2163#issuecomment-1342996811). 

So instead, I will redirect you to the forum. You have valid arguments, it shouldn't be hard to get people to agree. I will mark this PR as paused in the mean time.

Same reply for #1247.

pichi:
Thanks, @jiegillet for the advice. I have created a topic there
[Matching brackets tests don’t cover more than one balanced bracket group](https://forum.exercism.org/t/matching-brackets-tests-dont-cover-more-than-one-balanced-bracket-group/1655)

PR #1248, opened 2022-12-13
Add test case to pig-latin

pichi:
x anywhere except word beginning doesn't make vowel sound even in alien language.


angelikatyborska:
I believe the topic was abandoned by the author on the forum a year ago (https://forum.exercism.org/t/matching-brackets-tests-dont-cover-more-than-one-balanced-bracket-group/1655/4) so I'm closing this PR.

PR #1252, opened 2022-12-19
Remove unused and confusing :quantity key from options

MartinCapraro:
The [boutique suggestions exercise](https://exercism.org/tracks/elixir/exercises/boutique-suggestions) requires the use of a list comprehension filter to remove item pairs exceeding a maximum price. The maps in the tests include a field called `:quantity`, which one might reasonably conclude need to be multiplied by the price to get the total price, but this is not the case.

This confused me when I was doing the exercise, so this PR removes them.


angelikatyborska (approved):
Thanks! `:quantity` isn't needed for anything in this exercise so we can safely remove it.

PR #1253, opened 2022-12-22
Edit note on how to use `Bitwise` module on v1.14.0

fernandocanizo:
Maybe we should remove all Elixir version references and just go with the way it should be used on current stable Elixir version (v1.14.0). This is a text for people learning the language, they should not be learning it with older versions.


jiegillet:
Thank you, but this change is not accurate, in the current version you don't need to call `require Bitwise` anymore. It is used just like any other module, so it's not particularly worth highlighting how to call it.
For people who, for a reason or another don't have access to more recent versions of the language, I would like to keep the note as it is.

PR #1255, opened 2022-12-24
Add script to check autogenerated files on CI

angelikatyborska:
I noticed that we have no check in place that the autogenerated files are kept up to date.

This is my attempt at adding such a check.

The CI failure is expected (it should be proof that the changes work). The CI should start passing after https://github.com/exercism/elixir/pull/1256 is merged and incorporated into this branch.


bin/check_autogenerated_files.sh
```diff
@@ -0,0 +1,24 @@
...
+bin/configlet generate
+
+echo "Checking git diff"
+git diff --exit-code HEAD
+GIT_DIFF_EXIT_CODE="$?"
+
+if [ "$GIT_DIFF_EXIT_CODE" -ne 0 ]
+then
+  echo "Autogenerated files check failed. Please run \`configlet generate\` and commit in the changes."
+  exit 1;
```
angelikatyborska, line 20: If I set `set -euo pipefail` in this script, then it always exists after the git diff before echoing the explanation :( so I didn't add it. We risk it "passing" because something else goes wrong.

jiegillet: You could also grab the output, which should always succeed and exist depending on the content

```bash
GIT_DIFF=$(git diff)

if [ -z "$GIT_DIFF" ]
then 
  echo "Autogenerated files check passed"
  exit 0
else
  echo "Autogenerated files check failed. Please run \`configlet generate\` and commit in the changes."
  exit 1
fi  
```

jiegillet:
The CI kept failing even after merging #1256, so I closed and reopened. Seems to have worked.

bin/check_autogenerated_files.sh
angelikatyborska: Thanks, that's better. I had to wrap `echo $GIT_DIFF` into `echo "$GIT_DIFF"`, otherwise new lines weren't printed.

jiegillet: Oh that's a great trick. I think I should have used that in that gettext PR at work -_-

jiegillet (approved):
Oops, sorry for leaving that hanging for so long, I must have missed the update!
Great addition!

bin/check_autogenerated_files.sh
jiegillet: Oooh, even fancier. I should really invest more time learning shell syntax

jiegillet: Ah wait, but `git diff` always succeeds here, it doesn't fail when there is no difference.

glennj: I guess that's Angelika added the `--exit-code` option.

PR #1260, opened 2023-01-08
Rename MixProject module for library-fees exercise

dmarcoux:
I guess CaptainsLog came from the captains-log exercise, but they aren't related at all.


angelikatyborska (approved):
Thanks!

PR #1261, opened 2023-01-09
Rename MixProject modules and application names to match their directory

dmarcoux:
In the same vein as #1260.

I stumbled on this while trying to run the tests of solved exercises in an umbrella project.

### DIFF
--- a/exercises/concept/bread-and-potions/mix.exs
+++ b/exercises/concept/bread-and-potions/mix.exs
@@ -1,4 +1,4 @@
-defmodule RPG.MixProject do
+defmodule BreadAndPotions.MixProject do
   use Mix.Project
 
   def project do
--- a/exercises/concept/city-office/mix.exs
+++ b/exercises/concept/city-office/mix.exs
@@ -1,4 +1,4 @@
-defmodule Form.MixProject do
+defmodule CityOffice.MixProject do
   use Mix.Project
 
   def project do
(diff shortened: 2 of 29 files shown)


angelikatyborska:
As far as I can tell, all of those mix project module names are like this to match the top-level module name in `lib`.

angelikatyborska:
@jiegillet Do you have an opinion on this? I'm 80% sure we shouldn't merge this (we should keep the module name in `mix.exs` the same as it is in `lib`) but I wanted to check what do you think?

jiegillet:
Yes, I also think so. Internal consistency is better than confirming to the slug. Of course we could change every file to match the slug, but we would have very little to gain for the noise generated.

angelikatyborska:
I'm going to close this then, sorry! @dmarcoux If you want, you could instead open an issue about the problem of running exercises as part of the umbrella app, telling us what's the advantage of being able to do that and what are the current problems with that. If more people get interested, we'll reassess how much work it would be to fix everything properly and whether it's worth doing.

dmarcoux:
I store the solutions to the Exercism Elixir track in an umbrella project in order to keep the solutions code all together in a single Git repository. It was also an excuse to apply what I have learned about continuous integration for Elixir projects.

I totally understand the points you all brought up here. It's all good. Thank you for taking the time to have a look 🙂.

PR #1267, opened 2023-02-01
Change the price inside the colour check test.

robindrost:
I accidentally removed the colour check and found out my tests where still passing. This was due to the price check. I have lowered the price in the colour test to make sure the check on the colour match is in the code.


jiegillet (approved):
Yes, thank you, that's a really nice improvement!

PR #1269, opened 2023-02-02
Add missing comma

solar05:


jiegillet:
Thank you @solar05.
The CI doesn't pass, because there is another `introduction.md` in one of the exercises that has to match. Could you either find the file and modify it or run `bin/configlet generate`?

concepts/pattern-matching/introduction.md
```diff
@@ -17,7 +17,7 @@ The use of pattern matching is dominant in assertive, idiomatic Elixir code. You
 
 Remember, matches occur from the right side to the left side.
 
-In the last example if we don't need a variable in a pattern match, we can discard it by referencing `_`. Any variable starting with an `_` is not tracked by the runtime.
+In the last example, if we don't need a variable in a pattern match, we can discard it by referencing `_`. Any variable starting with an `_` is not tracked by the runtime.
```
jiegillet, line 20:
```suggestion
In the last example, if we don't need a variable in a pattern match, we can discard it by referencing `_`.
Any variable starting with an `_` is not tracked by the runtime.
```
Also, changing it to one line per sentence would be great.

solar05: Done 👌

jiegillet (approved):
Sweet, thanks!

PR #1280, opened 2023-02-23
[freelancer-rates] Make it clear that discount can apply to under 22 days of work

iHiD:
With reference to [this forum discussion](https://forum.exercism.org/t/freelancer-rates-exercise-issue/3188/7), this exercise seems to be confusing people. I think this change will remove the confusion.

### DIFF
--- a/exercises/concept/freelancer-rates/.docs/instructions.md
+++ b/exercises/concept/freelancer-rates/.docs/instructions.md
@@ -8,7 +8,7 @@ We first establish a few rules between the freelancer and the project manager:
 - The daily rate is 8 times the hourly rate.
 - A month has 22 billable days.
 
-The freelancer is offering to apply a discount if the project manager chooses to let the freelancer bill at the end of each month (regardless of how many days have been worked), which can come in handy if there is a certain budget the project manager has to work with.
+Sometimes, the freelancer is offering to apply a discount on their daily rate (for example for their most loyal customers or for non-for-profit customers).
 
 Discounts are modeled as fractional numbers representing percentage, for example `25.0` (25%).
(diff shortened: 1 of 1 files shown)


exercises/concept/freelancer-rates/.docs/instructions.md
```diff
@@ -8,7 +8,7 @@ We first establish a few rules between the freelancer and the project manager:
 - The daily rate is 8 times the hourly rate.
 - A month has 22 billable days.
 
-The freelancer is offering to apply a discount if the project manager chooses to let the freelancer bill per month, which can come in handy if there is a certain budget the project manager has to work with.
+The freelancer is offering to apply a discount if the project manager chooses to let the freelancer bill at the end of each month (regardless of how many days have been worked), which can come in handy if there is a certain budget the project manager has to work with.
```
angelikatyborska, line 11: So, If I understand correctly, the issue is that some people think that a discount should only be applied to 22-day chunks of work? Because of the phrase "per month"?

If I got that right, then changing "per month" to "at the end of each month" doesn't fix things. Yes, the addition in the parenthesis makes a helpful clarification, but... let's just remove the source of confusion altogether so that the clarification is not even needed?

```suggestion
Sometimes, the freelancer is offering to apply a discount on their daily rate (for example for their most loyal customers or for non-for-profit customers).
```

Does that work? It now clearly says that the discount applies to the daily rate, like in the example solution.

iHiD: Yes, that's much better :)

angelikatyborska (approved):

PR #1285, opened 2023-03-04
rpg-character-sheet: IO inspect label problems

angelikatyborska:
As reported here https://forum.exercism.org/t/rpg-character-sheet-label-hint-suggestion/4279 and here http://forum.exercism.org/t/bug-report-tracking/1334/10, people are struggling to notice that IO.inspect adds a color and a space for them. The output of the one test that fails if you misuse the label is not easy to parse as it's a regex match and prints out the whole string, with the double color hidden somewhere in the middle, hard to notice (see screenshot, you can test your own eyes 🙂 ).

To help out with this, I'm adding a new test that will fail explicitly on misuse of the label option, and a new hint.

Screenshot 2023-03-04 at 08 43 35


jiegillet (approved):
That new test is super clear, nicely done :)

PR #1294, opened 2023-03-20
update divide/2 for code consistency

ctstacey:


jiegillet:
What is the motivation for this change? What is inconsistent at the moment?

ctstacey:
- The surrounding code does not use the `_by` post-fix in function names.
  For example, it uses `multiply` rather than `multiply_by`.

- The surrounding code uses `a` and `b` as parameter names, rather than `num` and `den`.

https://github.com/exercism/elixir/blob/12cfced545f501fcd0b74f6ef85b340a81564507/exercises/practice/rational-numbers/lib/rational_numbers.ex#L21-L37

jiegillet:
I  see, but in my opinion, the current version is better, because division is not commutative.

> * The surrounding code does not use the `_by` post-fix in function names.
>       For example, it uses `multiply` rather than `multiply_by`.

I think this name is great for pipes

```elixir
area = (base * height)
  |> divide_by(2)
```
because you pronounce it the same way as you would in English, and it really helps with knowing which is the numerator and denominator. `divide` is a bit more ambiguous. For multiplication, it doesn't matter which is which, so `multiply_by` is not as necessary (although it does sound better to read out).

> * The surrounding code uses `a` and `b` as parameter names, rather than `num` and `den`.

`num` and `den` help disambiguate which parameter is divided by which. In general, descriptive parameter names are preferable. Again, addition an multiplication are commutative, so it doesn't help. 

If anything, `subtract` could use better names, but there is no common word for the value you subtract, and `subtract_from` would flip the conventional parameter order...

ctstacey:
Some interesting points. 👍

Personally I don't have issue with `a |> divide(b)`, and I prefer `add`, `subtract`, `multiply`, `divide`, but that's just me.

I'm happy to close the PR and wish you a good day! 🙂

PR #1295, opened 2023-03-25
Easier learning exercise for bitstrings - paint-by-numbers

angelikatyborska:
Over the months, I've seen complaints here and there about the dna-encoding exercise. People struggle with it because it attempts to teach two very difficult concepts all at once - bitstrings and tail call recursion.

As a first step to fix this, I would like to create an easier exercise that only teaches bitstrings and keep dna-encoding as a learning exercise only for tail call recursion.

According to the data from https://exercism.org/tracks/elixir/build, only 4 other learning exercises have a lower completion ratio than dna encoding, and those exercises are much later in the concept tree.

Screenshot 2023-03-25 at 19-43-13 Elixir on Exercism


exercises/concept/paint-by-number/.docs/instructions.md
```diff
@@ -0,0 +1,159 @@
...
+* . . . . . . . . . . . *
+* . . . . . . . . . . . *
+* . . . . . . . . . . . *
+  * . . . . . . . . . *
+    * . . . . . . . *
+      * . . . . . *
+        * . . . *
+          * . *
+            *
+```
```
angelikatyborska, line 37: I'm repeating myself, high-school sweetheart already has an ASCII heart 😉 but a heart was the most recognizable shape I could do in ASCII that could be made out of "pixels" and not infinitely many arbitrary text characters like ASCII art usually is.

I saw that the sudoku killer exercise comes with an image. I think I will make this into an image, with some different shape... like a smiley? Or this bee https://ih1.redbubble.net/image.93120918.6282/flat,800x800,075,f.jpg ?

exercises/concept/paint-by-number/.docs/instructions.md
```diff
@@ -0,0 +1,159 @@
...
+
+This picture uses 3 colors.
+Let's say we assign indices to those colors:
+- `0` (binary: `0b00`) for white,
+- `1` (binary: `0b01`) for black`,
```
angelikatyborska, line 42:
```suggestion
- `1` (binary: `0b01`) for black,
```

config.json
```diff
@@ -282,13 +282,29 @@
...
       },
       {
         "slug": "dna-encoding",
         "name": "DNA Encoding",
         "uuid": "994c8760-8a58-4ef3-8a28-a0f9182f4d9f",
         "concepts": [
-          "bitstrings",
           "tail-call-recursion"
         ],
         "prerequisites": [
```
neenjaw, line 310: Wouldn't bitstrings be listed as a prerequisite here?

exercises/concept/paint-by-number/.docs/instructions.md
neenjaw: Am good to follow your intuition here. I think the heart is good even though we have it somewhere else as it is simple and not likely to be mistaken or misinterpreted.

neenjaw (approved):
Looks good, makes sense to insert another exercise.

config.json
angelikatyborska: Yes, absolutely, I forgot!

meatball133:
Okay I solved the exercise without looking at the exemplar.
I am not an amazing elixir developer but I can get myself around the language at least.
It perhaps isnt the perfect solution, like the first function I would likely have used pow, with ceil instead of recursion.
But since the instructions said to do this way I did.

```elixir
defmodule PaintByNumber do
  def palette_bit_size(color_count, count \\ 0) do
    if color_count > Integer.pow(2, count) do
      palette_bit_size(color_count, count + 1)
    else
      count
    end
  end

  def empty_picture() do
    <<>>
  end

  def test_picture() do
    <<0::2, 1::2, 2::2, 3::2>>
  end

  def prepend_pixel(picture, color_count, pixel_color_index) do
    count = palette_bit_size(color_count)
    pixel = <<pixel_color_index::size(count)>>
    << pixel::bitstring,picture::bitstring>>
  end

  def get_first_pixel(picture, color_count) do
    if picture != <<>> do
      count = palette_bit_size(color_count)
      <<value::size(count), rest::bitstring>> = picture
      value
    else
      nil
    end
  end

  def drop_first_pixel(picture, color_count) do
    if picture != <<>> do
      count = palette_bit_size(color_count)
      <<value::size(count), rest::bitstring>> = picture
      rest
    else
      <<>>
    end
  end

  def concat_pictures(picture1, picture2) do
    <<picture1::bitstring, picture2::bitstring>>
  end
end
```
I think this is a fun and good exercise, I found it pretty easy getting through I have some feedback:

There is in instruction a place where perhaps the note cold be replace with a note special block.

prepend_pixel wasn't the easiest to get around, after some studying I figured out what to do, but it wasn't super straight forward, my feedback:

Have an exercise there you create a pixel, perhaps remove test_picture and replace it with create_pixel.
If I am doing what you are suppose to do in concat_pictures so I sorta feel like there is no point in having the last function, or perhaps I did the prepend_pixel incorrect.

vaeng:
I think this is a lot clearer than the DNA exercise. I am one of the people who ruin the AST exercise statistics. So take my opinion with a grain of salt :D

I did the exercise like this:
```elixir
defmodule PaintByNumber do
  def palette_bit_size(color_count) when color_count == 0, do: 0
  def palette_bit_size(color_count) when color_count <= 2, do: 1
  def palette_bit_size(color_count) do
    1 + palette_bit_size(color_count/2)
  end


  def empty_picture() do
    <<>>
  end

  def test_picture() do
    <<0::size(2), 1::size(2), 2::size(2), 3::size(2)>>
  end

  def prepend_pixel(picture, color_count, pixel_color_index) do
    bit_size = palette_bit_size(color_count)
    <<pixel_color_index::size(bit_size) , picture::bitstring>>
  end

  def get_first_pixel(<<>>, color_count), do: nil
  def get_first_pixel(picture, color_count) do
    bit_size = palette_bit_size(color_count)
    <<pixel::size(bit_size), _rest::bitstring>> = picture
    pixel
  end

  def drop_first_pixel(<<>>, color_count), do: <<>>
  def drop_first_pixel(picture, color_count) do
    bit_size = palette_bit_size(color_count)
    <<_droppedpixel::size(bit_size), rest::bitstring>> = picture
    rest
  end

  def concat_pictures(picture1, picture2) do
   <<picture1::bitstring, picture2::bitstring>>
  end
end

```

I stumbled upon the first exercise and was not really sure how to solve it in a good way. As soon as I code Elixir; I see recursion everywhere, so that's what I did. It is not really needed for the concept of the exercise. If you want to cut it somewhere, you could just hand the implementation to the students.

I was also a thrown of a bit, when I tried `<<_droppedpixel::size(palette_bit_size(color_count)), rest::bitstring>> = picture`and the compiler complained, that it want's an Integer. So I had to it change to:
```elixir
bit_size = palette_bit_size(color_count)
<<_droppedpixel::size(palette_bit_size(bit_size )), rest::bitstring>> = picture
```

Another thing that I asked myself was if  `<<something::size(2)>>` or `<<something::2>`is more idiomatic. I did not really get that from the docs.

glennj:
<details><summary>My take is largely similar to vaeng's:</summary>

```elixir
defmodule PaintByNumber do
  #use Bitwise

  def palette_bit_size(color_count) do
    #do_palette_bit_size(color_count, 1)
    :math.log2(color_count) |> Float.ceil() |> trunc()
  end

  #defp do_palette_bit_size(color_count, bits) do
  #  case 1 <<< bits do
  #    n when n >= color_count -> bits
  #    _ -> do_palette_bit_size(color_count, bits + 1)
  #  end
  #end

  def empty_picture(), do: <<>>

  def test_picture(), do: <<0::2, 1::2, 2::2, 3::2>>

  def prepend_pixel(picture, color_count, pixel_color_index) do
    bits = palette_bit_size(color_count)
    <<pixel_color_index::size(bits), picture::bitstring>>
  end

  def get_first_pixel(<<>>, _), do: nil
  def get_first_pixel(picture, color_count) do
    bits = palette_bit_size(color_count)
    <<pixel::size(bits), _::bitstring>> = picture
    pixel
  end

  def drop_first_pixel(<<>>, _), do: <<>>
  def drop_first_pixel(picture, color_count) do
    bits = palette_bit_size(color_count)
    <<_::size(bits), picture::bitstring>> = picture
    picture
  end

  def concat_pictures(picture1, picture2) do
    <<picture1::bitstring, picture2::bitstring>>
  end
end
```
</details>

It's been a while since I went through the elixir syllabus, but I recall bitstrings as being a bit confusing. I appreciate the expanded concept introduction.

This is the only bit where I stumbled a little:

> Pattern matching can also be done to obtain the value from within the special form. The `::bitstring` type must be used to pattern match on a bitstring of an unknown size. It can only be used for the last fragment of the bitstring.

And it was only the last "It" that tripped me up. I suggest

> Pattern matching can also be done to obtain values from the special form. You have to know the number of bits for each fragment you want to capture, with one exception: the `::bitstring` type can be used to pattern match on a bitstring of an unknown size, but this can only be used for the **last** fragment.

exercises/concept/paint-by-number/.docs/hints.md
```diff
@@ -0,0 +1,57 @@
...
+- Read about [bitstrings][bitstring] in the Getting Started guide.
+- Review the documentation for the [bitstring special form][bitstring-form].
+- Review [how to convert decimal numbers to binary numbers]([decimal-to-binary-youtube]).
+- Elixir supports [different ways of writing integers][integer-literal], including binary notation.
+
+## 1. Calculate palette bit size
+
+- This task doesn't use the bitstring data type.
+- Find the smallest positive number `n` such that 2 raised to the power of `n` is greater than or equal than the number of colors we need to represent.
+- Create a recursive function for this.
```
vaeng, line 14: I did not understand the instructions in th same way. I thought I was supposed to find a direct way.

exercises/concept/paint-by-number/.docs/hints.md
```diff
@@ -0,0 +1,57 @@
...
+## 6. Drop the first pixel from a picture
+
+- The [bitstring special form][bitstring-form] can be used to [pattern match bitstrings][bitstring-matching].
+- Use the previously implemented `PaintByNumber.palette_bit_size/1` function to get the bit size of one pixel.
+- Use the special `::bitstring` type to specify that the rest of the bitstring is of unknown size.
+
+## 7. Concatenate two pictures
+
+- The [bitstring special form][bitstring-form] can be used to concatenate two bitstrings.
+- Use the special `::bitstring` type to specify that the two bitstring fragments are of unknown size.
```
vaeng, line 49: I was not confused by this, but the student might conclude that it is only possible to use the bitstring type once (for the last part of the construction process of a new bitstring).

angelikatyborska:
Thank you all, I'll read and reply to your comments in a few days because right now my ass is getting kicked by the common cold 😷

angelikatyborska:
> If you want to cut it somewhere, you could just hand the implementation to the students.

I was considering this, but there's a lot of hurdles.
- I cannot hand them the implementation of `palette_bit_size`. I want students to implement it, so that they internalize better what the size actually means and where it comes from.
- I cannot hand them an implementation that uses the Erlang library - that goes against the assumptions of the learning mode. They shouldn't see something that wasn't explained yet.
- I would need to write my own logarithm function, but only using concepts introduced in the syllabus so far. I'm not great at computer science and algorithms. I think I can do a basic `log2` function that only works with integers (e.g. `log2(9) == 8`), but if I do this and build-in the floor operation in the process, that will remove one degree of understanding what's happening. I would love to implement a proper `log2` that returns floats, but... how? @vaeng do you know?

angelikatyborska:
> If I am doing what you are suppose to do in concat_pictures so I sorta feel like there is no point in having the last function, or perhaps I did the prepend_pixel incorrect.

@meatball133 Hehe, you did implement `prepend_pixel` in a way that I didn't expect. You did:
```elixir
  def prepend_pixel(picture, color_count, pixel_color_index) do
    count = palette_bit_size(color_count)
    pixel = <<pixel_color_index::size(count)>>
    <<pixel::bitstring, picture::bitstring>>
  end
```

But it could have been just:
```elixir
  def prepend_pixel(picture, color_count, pixel_color_index) do
    count = palette_bit_size(color_count)
    <<pixel_color_index::size(count), picture::bitstring>>
  end
```

> Have an exercise there you create a pixel, perhaps remove test_picture and replace it with create_pixel.

@meatball133 I'm not sure what that `create_pixel` function should do? Can you show me an example?

exercises/concept/paint-by-number/.docs/hints.md
angelikatyborska: I assume that means you skipped reading the instructions midway and didn't see the sentence "Now, solve this task with recursion and..."?

exercises/concept/paint-by-number/.docs/hints.md
angelikatyborska: Would this tiny wording tweak help?
```suggestion
- Use the special `::bitstring` type to specify that each of the bitstring fragments is of unknown size.
```

angelikatyborska:
> Another thing that I asked myself was if <<something::size(2)>> or <<something::2>is more idiomatic. I did not really get that from the docs.

I honestly have no idea. We would need an Elixir dev who has worked in embedded systems to know this, and that is definitely not me 😅. Since most Elixir devs are web devs, I also didn't see any mentions of this choice in style guides etc.

meatball133:
They way I solved it I thought it made sense to have:
```elixir
def create_pixel (color_count, pixel_color_index) do
    count = palette_bit_size(color_count)
    <<pixel_color_index::size(count)>>
end
# and then

def prepend_pixel(picture, color_count, pixel_color_index) do
    pixel = create_pixel(color_count, pixel_color_index)
    <<pixel::bitstring, picture::bitstring>>
end
```

But now looking at how it is supposed to be solved my suggestion may not make as much sense.

angelikatyborska:
@meatball133 I added a new section to the introduction that hopefully would have lead you to the expected implementation of `prepend_pixel`: [`4497b69` (#1295)](https://github.com/exercism/elixir/pull/1295/commits/4497b69c0dd5e641c11b8e118c0959905d1aac60)

meatball133:
Yeah, I think I just did something weird I dunno, that addition is helpful. I guess this system is nothing I have seen in other languages and thereby I guess my brain saw the last section and thought it could do that, but I think this new addition really helps.

exercises/concept/paint-by-number/.docs/hints.md
vaeng: Yes, I think that would show the difference to the previous use case better.

angelikatyborska:
I'll merge this now but if anyone has any more comments/replies, keep them coming! I can always open a follow up PR with more improvements.

PR #1297, opened 2023-04-02
Sync bank-account

angelikatyborska:
Bank account has received canonical data!

`update` gets split into two functions: `deposit` and `withdraw`, with some extra validations (cannot deposit/withdraw non-positive amounts, cannot withdraw more than you have).


jiegillet (approved):
Very nice, thank you!
Not implementing those cases makes sense.

PR #1308, opened 2023-04-12
Correct typos where ::bitstring has been mistyped

repent:

PR #1307, opened 2023-04-12
Fix formatting in paint-by-numbers introduction

stmllr:
* Fixes a typo 
* Splits lines each with <=80 chars

PR #1309, opened 2023-04-12
Fixed typos and formatting of md files

moziac:
Fixed the ::bistring typos and changed the formatting, the original was difficult to read due to the constant need to scroll.

### DIFF
--- a/concepts/bitstrings/introduction.md
+++ b/concepts/bitstrings/introduction.md
@@ -1,21 +1,31 @@
 # Introduction
 
-Working with binary data is an important concept in any language, and Elixir provides an elegant syntax to write, match, and construct binary data.
+Working with binary data is an important concept in any language, and Elixir 
+provides an elegant syntax to write, match, and construct binary data.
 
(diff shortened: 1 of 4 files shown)


angelikatyborska:
Thank you for the input!

The typos have been addressed by other people before you opened this PR (https://github.com/exercism/elixir/pull/1307, https://github.com/exercism/elixir/pull/1308, https://github.com/exercism/elixir/pull/1311).

As for the formatting changes, you will notice that all the concept introductions and concept exercise instructions and hints don't have any kind of line length limit. If I were to agree with you that long tines are a problem, we would need to reformat all of those files. Making formatting changes to a single file wouldn't be very helpful and it would break consistency.

I am personally of the opinion that line length in markdown files for paragraphs (not code blocks) doesn't matter. In most editors, you can turn on soft wraps for overflowing lines. You should even be able to control that setting per file type, in case you only want them in markdown files and not in source code files. If reading markdown files in your editor is uncomfortable, I would recommend researching some plugins for your editor, or maybe consider a dedicated app just for reading markdown files? Or you can always open up the exercise instructions in a browser from exercism.org.

If you still feel strongly that all markdown files should be formatted to conform to a specific length limit, please open an issue in this repository so we can discuss further. It would require adding a formatter of some kind (prettier?) because I am not willing to format all those makrdown files by hand 😉.

PR #1317, opened 2023-04-24
Adding another test for Top Secret exercise.

### LINKED ISSUE #1315: Top Secret could use another test
Exercise Top Secret, `decode_secret_message_part/2` function. I believe there should be another test when the AST (the first argument of the fucntion) is formed from the code consisting of only primitive types such as `12` or `:ok`.
For example, if `ast = TopSecret.to_ast("12")` or `ast = TopSecret.to_ast(":ok")`, then `TopSecret.decode_secret_message_part(ast, acc)` should return `{ast, acc}`. However, there is no test for that; all the tests assume that the AST is formed from the code containing a function. One could pass all the tests even though the function raises an `ArgumentError` for such an argument. This test could as well help the solver with task the Task 5 of the exercise as some nodes simply contain only primitive types; took me quite some time to notice that one should check if the ast in the function is actually a tuple.

I am willing to do this myself, but if I am correct one should first open an issue here.

krstopro:
Added another test for the exercise Top Secret that checks if the `decode_secret_message_part/2` function works for ASTs made of code consisting only of basic types.
- Fixes #1315.

### DIFF
--- a/exercises/concept/top-secret/test/top_secret_test.exs
+++ b/exercises/concept/top-secret/test/top_secret_test.exs
@@ -86,7 +86,7 @@ defmodule TopSecretTest do
       assert actual_ast == ast
       assert actual_acc == acc
     end
-    
+
     @tag task_id: 2
     test "works for AST made only of basic types" do
       acc = ["abc"]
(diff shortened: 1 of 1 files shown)


angelikatyborska (approved):
Thank you!

PR #1320, opened 2023-04-26
Ensure module is loaded before testing

### LINKED ISSUE #1319: [Dancing Dots] "defines a __using__ macro" test is flakey in certain scenarios
When running the test `Animation`.`defines a __using__ macro` I found that it would only pass the first time when using the VS Code Elixir Test Runner extension.

Subsequent runs of the test failed until I changed the `Animation` module arbitrarily (for example, changing `init(_opts)` to `init(_)`, but the change doesn't matter). This test would fail 100% of the time if I ran the `Animation` module's tests one by one leading up to testing for `__using__` (the test runner extension runs tests and test groups one by one, rather than all at once as `mix test` does).

This lead me down the hex docs for `macro_exported?` that is used in the test, and it notes that this function does not load the module first if it isn't loaded. I'm not quite sure why the module isn't loaded in this scenario, but it caused some frustration as I'm not familiar with behaviours and the exercise code I wrote was correct as checked by the Exercism test runner:

https://exercism.org/tracks/elixir/exercises/dancing-dots/solutions/dbaer

Adding `Code.ensure_loaded/1` before testing that the macro is defined resolves this issue, and a PR resolving this should be referenced below.

dabaer:
that `__using__` is defined. Fixes #1319

### DIFF
--- a/exercises/concept/dancing-dots/test/dancing_dots/animation_test.exs
+++ b/exercises/concept/dancing-dots/test/dancing_dots/animation_test.exs
@@ -15,6 +15,7 @@ defmodule DancingDots.AnimationTest do
 
     @tag task_id: 2
     test "defines a __using__ macro" do
+      Code.ensure_loaded(DancingDots.Animation)
       assert macro_exported?(DancingDots.Animation, :__using__, 1)
     end
 


angelikatyborska (approved):
Thank you!

I'm going to close and reopen the PR to see if that will make all the CI checks run.

PR #1321, opened 2023-05-18
Add a test case for File Sniffer exercise

cr0t:
Recently, with one of my students _(hey, Pul!)_, we found out that many community solutions to the File Sniffer exercise (including mine) contains a bug in the `verify/2` logic. _The [`exemplar.ex`](https://github.com/exercism/elixir/blob/c0fa09c469b08504b951f36c11516dafcfcbb6a6/exercises/concept/file-sniffer/.meta/exemplar.ex) has the same issue too!_

With a straightforward (and quite simplistic) approach, we can have code like this:

```elixir
def verify(file_binary, extension) do
  ext_type = type_from_extension(extension)
  bin_type = type_from_binary(file_binary)

  if bin_type == ext_type do
    {:ok, bin_type}
  else
    {:error, "Warning, file format and file extension do not match."}
  end
end
```

Given that numerous solutions rely on a Map of file extensions and/or on pattern matching on bit strings, this leads to a problem when both (`type_from_binary/1` and `type_from_extension/1`) functions return `nils`. With the snippet above, we will return `{:ok, nil}`, which is somewhat not what might be expected (in a real program).

The current test suite doesn't cover such a thing (when both extension and binary format are not expected in the solution).

_But we must encourage our students to think about such things._

With these changes, I added a very tiny data.dat file and three tests to cover this.

The _(simplest)_ solution to the snippet we showed above, and the new tests I've added, will be to add a check for extension (for example):

```elixir
def verify(file_binary, extension) do
  bin_type = type_from_binary(file_binary)
  ext_type = type_from_extension(extension)

  if bin_type == ext_type and not is_nil(bin_type) do
    {:ok, bin_type}
  else
    {:error, "Warning, file format and file extension do not match."}
  end
end
```


angelikatyborska (approved):
Thank you, that's a very good catch and good fix for this exercise 🙏

Issue #1283, opened 2023-03-03
Mention function/variable naming possibilities and conventions in the learning module


angelikatyborska:
There are two pieces of information that are missing from the learning module, that users for a programming language should know:
- what are valid function/variable names
- what are typical function/variable names

We should make sure that the introductions to some of the early concept exercises mention:

- Casing (functions and variables snake_case, module names PascalCase)
- Which characters are allowed in function, variable, and module names (people found it surprising to see an argument named `legacy?` in one exercise)
- What are some naming conventions
	- boolean functions with a question mark (mention this in the boolean concept?)
	- raising functions with an exclamation mark (mention this in the raising errors concept?)

Review the introductions and find appropriate places to insert this information.

PR #1325, opened 2023-06-12
Mention function/variable naming possibilities and conventions in the learning module

angelikatyborska:
Resolves https://github.com/exercism/elixir/issues/1283

This text contains a white lie about which letters are allowed in which names. I'm simplifying and saying that ASCII letters are allowed everywhere, but in reality function and variable names (not module names) support more letters, e.g. `spät? = true`.

The convention that `!` functions raise errors has already been mentioned in the introduction of the errors concept.


neenjaw (approved):
Nice improvements here.

PR #1327, opened 2023-06-20
Fix dead links to hexdocs.pm

CarlosHSF99:
Many links pointing to documentation in hexdocs.pm, like the ones in the [Guards](https://exercism.org/tracks/elixir/concepts/guards) concept introduction page, are incorrect (outdated?) as they include an extraneous path segment, "master".

Fixed every link with such an error in the Elixir track.


angelikatyborska:
Big big thanks 🙏 I wouldn't have noticed the problem on my own.

angelikatyborska (approved):

PR #1330, opened 2023-06-22
Set up dead link checker in CI

### LINKED ISSUE #1328: Set up a dead link checker
Apparently we had a lot of dead links in the repo (https://github.com/exercism/elixir/pull/1327) and didn't even know 😞. We should set up something like the problem specifications repo has to find dead links automatically: https://github.com/exercism/problem-specifications/blob/main/.github/workflows/ci.yml#L126-L138

CarlosHSF99:
Resolves #1328

Implementation was done following the example in the issue, but including json in the type of files to be tested to check the links.json files for the complementary links in the concepts pages.

Also fixed some links so as not to leave the repo in an invalid state for the test being added:
 - updated some urls to http**s**
 - fixed slack link in [resources](https://exercism.org/docs/tracks/elixir/resources)
 - added the irc link in resources to the new .lycheeignore file as it is not supported by lychee

Tested with [act](https://github.com/nektos/act), and it seems to be working, with every link passing the test.


angelikatyborska:
I would love to test it with act too, but I'm having some trouble using it. Can you tell me which docker image did you use? The medium one `catthehacker/ubuntu:act-20.04` or the large one `catthehacker/ubuntu:full-20.04`?

angelikatyborska:
Alright, I managed to get act to work with the medium docker container. I just needed the right flag because I'm on an M1 mac (` act -j linkChecker --container-architecture linux/amd64`)

Something is not right. Lychee reports finding 46 links in total, but we have 60 concepts in this repo, each concept has two `.md` files and one `.json` file, I would say there's at least 10 links per concepts in those three files. So I would expect the total link count of the whole repo to be above 500, maybe even above 1000.

If I test and mess up a link on purpose in files for example `concepts/basics/about.md`, `concepts/exceptions/links.json`, or `exercises/practice/affine-cipher/.meta/config.json`, the link checker doesn't detect that 🤔

angelikatyborska:
That's more like it!
```
| | Status        | Count |
| |---------------|-------|
| | 🔍 Total      | 1158  |
| | ✅ Successful | 1108  |
| | ⏳ Timeouts   | 0     |
| | 🔀 Redirected | 0     |
| | 👻 Excluded   | 1     |
| | ❓ Unknown    | 0     |
| | 🚫 Errors     | 49    |
```

I'll fix the remaining errors in this PR and merge, thank you very much 💜

CarlosHSF99:
Thank you! Although I was still unsure of the results, I couldn't find the same number of links with a python script, and the lack of single quotes does not seem to cause a problem in the problem-specifications repo.

angelikatyborska:
Related PR: https://github.com/exercism/problem-specifications/pull/2293

CarlosHSF99:
Fixed the ci.yml file. In the previous commit I forgot this one, having only fixed the pr.ci.yml.

angelikatyborska:
I did not expect how many links are going to pose problems for the link checker 🤦. Now it's twitter blocking not logged in users, and some random timeouts from other pages that aren't really broken links. I'll wait a bit and try to re run the check... I hope it won't always be so flaky.

dabaer:
@angelikatyborska Would it be helpful to enable the `--cache` flag? It seems Lychee doesn't cache results by default, which is slightly contributing to hitting rate limits on some domains.

angelikatyborska:
I lost my patience and ignored all the links that are giving false positives wrongly on CI (they work in the browser and they work with curl and with lychee from my local machine)

angelikatyborska (approved):

PR #1332, opened 2023-06-24
Add Elixir 1.15 and OTP 26

angelikatyborska:


dabaer:
So :invalid_configuration is being caught in the GenServer init function and returning {:stop, error} as opposed to {:error, error} as the test expects.

Was this working on prior elixir versions? I think it's just the test that would need changing

dabaer:
https://github.com/exercism/elixir/blob/08d41282251dd469790b801f35171f4a2f4cf666/exercises/concept/take-a-number-deluxe/.meta/exemplar.ex#L41

Should be ` {:error, error} -> {:error, error} `

Sorry i'm not quite sure how to add files or suggestions to a PR

angelikatyborska:
I think you cannot add changes to my PR.

Anyway, I'm not yet convinced that would be the correct thing to do. The GenServer specification clearly says that the `init` function is supposed to return `{:stop, reason}`: https://hexdocs.pm/elixir/GenServer.html#c:init/1

I think the difference in behavior here is that calling `start_link` now doesn't return a value at all 🤔 maybe we need to add trapping exits in the test process?

dabaer:
Ah you are right. I was thinking about trapping the exit to. For a second I thought it was in this exercise but it was a different one.

I don't see anything regarding GenServer in the 1.15 release notes, so maybe it's a bug in Elixir?

angelikatyborska:
It must be an unexpected change from OTP 26. I just tested with Elixir 1.15 and OTP 25 and GenServer stopped in `init` returns a value there. It stops returning a value with OTP 26 😕

dabaer:
Very weird. I am having good luck with trapping the exits in tests:

```elixir
test "min and max numbers get validated" do
      Process.flag(:trap_exit, true)

      pid = spawn_link(fn ->
        TakeANumberDeluxe.start_link(min_number: 999, max_number: 99)
      end)
      assert_receive {:EXIT, ^pid, :normal}

      pid = spawn_link(fn ->
        TakeANumberDeluxe.start_link(min_number: :not_a_number, max_number: 99)
      end)
      assert_receive {:EXIT, ^pid, :normal}

      pid = spawn_link(fn ->
        TakeANumberDeluxe.start_link(min_number: 1, max_number: "not a number")
      end)
      assert_receive {:EXIT, ^pid, :normal}
    end
```

dabaer:
I'm not sure if updating dialyxer was necessary (I imagine they have more and better checks as the version in mix is quite old) but I changed to runtime: false as well as they recommend.

```elixir
{:dialyxir, "~> 1.3.0", runtime: false}
```

And tweaked a very old application definition that broke for me when I first tried adding dialyzer to the code path.

```elixir
def application do
  [extra_applications: [:logger, :dialyzer]]
end
```

Other than that tests, dialyzer, and configlet lint are all green.

angelikatyborska:
Yup, upgrading dialyzer sounds like a good idea 🙂 thanks for the hint with applications - it would take me forever to find that problem.

I reported the changed OTP behavior on the Elixir forum, waiting for advice: https://elixirforum.com/t/genserver-start-link-when-stopped-in-init-no-longer-returns-a-value-in-otp-26/56672

But I think we can merge this PR with trapping exits to keep the test working.

dabaer:
Your solution is way more pragmatic, when I thought of trapping the exists I went and changed all the assertions 😆

dabaer:
Ah yes, they changed the formatter to switch charlists to a sigil instead. All `''` will format to `~c""`.

As for the Bank test error now I don't get that locally.

dabaer:
If I run that test specifically with `test-exercises.sh` it passed, but if I run `mix test` in the exercise directory, I get a fail

```elixir
1) test newly opened account has zero balance (BankAccountTest)
     test/bank_account_test.exs:10
     Assertion with == failed
     code:  assert BankAccount.balance(account) == 0
     left:  nil
     right: 0
     stacktrace:
       test/bank_account_test.exs:11: (test)
```

angelikatyborska:
> Ah yes, they changed the formatter to switch charlists to a sigil instead. All '' will format to ~c"".

That will mean we need even more changes, I created a ticket: https://github.com/exercism/elixir/issues/1333

angelikatyborska:
The bank account test fails on my machine only sometimes, here's the failure:

```
  1) test can handle concurrent transactions (BankAccountTest)
     test/bank_account_test.exs:99
     ** (exit) exited in: Task.await(%Task{mfa: {:erlang, :apply, 2}, owner: #PID<0.146.0>, pid: #PID<0.149.0>, ref: #Reference<0.0.18691.79558460.803012613.29020>}, 100)
         ** (EXIT) time out
     code: |> Enum.map(fn task -> Task.await(task, 100) end)
     stacktrace:
       (elixir 1.15.0) lib/task.ex:846: Task.await_receive/3
       (elixir 1.15.0) lib/enum.ex:1693: Enum."-map/2-lists^map/1-1-"/2
       (elixir 1.15.0) lib/enum.ex:1693: Enum."-map/2-lists^map/1-1-"/2
       test/bank_account_test.exs:109: (test)
```

dabaer:
I saw entries regarding Task using new optimizations from OTP in the changelog for 1.15, though it would be odd if that were causing the Tasks to take more than ~80ms longer than they should be to complete.

It's not exactly a heavy codepath here, seems like something is fishy in the GenServer or the Task.

I'm out of my office so I can't debug today unfortunately, but I'll be back tomorrow

dabaer:
So I bumped my virtual machine down to 1 core only, and I can reproduce this test failure 100% of the time that way. Perhaps OTP has changed the way it schedules processes which makes the combined tasks take more than 100ms to complete.

Theoretically I could see `Task.await` counting the time it takes for the Task to be scheduled by OTP in it's timeout, though I can't really find any low level information in the docs to confirm that.

I wonder if the best course of action would be simply to increase the timeout. I tested using `await` as well as `await_many` and it looks like the number of tasks are simply overloading the CIs processing threads, causing the individual tasks to take too long.

They both required the same timeout setting of 1000ms to complete in time. Both failed at 100ms and 500ms.

```elixir
0..999
|> Enum.map(fn _ ->
  Task.async(fn ->
    :timer.sleep(Enum.random(0..20))
    :ok = BankAccount.deposit(account, 1)
  end)
end)
|> Enum.map(fn task -> Task.await(task, 1000) end)
```

```elixir
0..999
|> Enum.map(fn _ ->
  Task.async(fn ->
    :timer.sleep(Enum.random(0..20))
    :ok = BankAccount.deposit(account, 1)
  end)
end)
|> Task.await_many(1000)
```

angelikatyborska:
Thank you for the investigation, it saved a lot of my own time 😅 

I increased the await time and now the CI is green 🎉

neenjaw (approved):
Changes to use the new digit and bump dial user look good to me. 👍

PR #1335, opened 2023-06-25
Fix `printf` usage with strings containing variables

### LINKED ISSUE #1334: test_exercises.sh print fails when test output includes `%`
When the output of a test result includes the special `%` character (such as when the stack trace contains a struct or a map), `printf` throws an error, and clips the test output as shown below.

```sh
test output -----------------------------------------------------------------
bin/test_exercises.sh: line 125: printf: `T': invalid format character
Including tags: [:pending]


BankAccountTest [test/bank_account_test.exs]
  * test withdraw twice [L#40]
  * test withdraw twice (0.04ms) [L#40]
  * test cannot withdraw from closed account [L#74]
  * test cannot withdraw from closed account (1.5ms) [L#74]
  * test withdraw once [L#32]
  * test withdraw once (0.03ms) [L#32]
  * test cannot deposit negative [L#95]
  * test cannot deposit negative (0.01ms) [L#95]
  * test can do multiple operations sequentially [L#49]
  * test can do multiple operations sequentially (0.03ms) [L#49]
  * test newly opened account has zero balance [L#12]
  * test newly opened account has zero balance (0.01ms) [L#12]
  * test cannot withdraw more than deposited [L#81]
  * test cannot withdraw more than deposited (0.02ms) [L#81]
  * test incrementing balance from another process then checking it from test process [L#117]
  * test incrementing balance from another process then checking it from test process (0.01ms) [L#117]
  * test implementation for multiple account support [L#138]
  * test implementation for multiple account support (0.02ms) [L#138]
  * test cannot check balance of closed account [L#60]
  * test cannot check balance of closed account (0.02ms) [L#60]
  * test can handle concurrent transactions [L#101]
  * test can handle concurrent transactions (455.9ms) [L#101]

  1) test can handle concurrent transactions (BankAccountTest)
     test/bank_account_test.exs:101
     ** (exit) exited in: Task.await(Tests failed
Error: Process completed with exit code 1.
```

dabaer:
Variables must not be used in the first argument of `printf` as it may cause errors or unintended behaviour when the variable contains special characters such as `%`.

Fixes #1334.

### DIFF
--- a/bin/test_exercises.sh
+++ b/bin/test_exercises.sh
@@ -25,7 +25,7 @@ function test_or_tests () {
     then
       tests+="s"
     fi
-    printf "%s" "$tests"
+    printf '%s' "${tests}"
 }
 
 # Initialize counts / array
@@ -59,7 +59,7 @@ do
     exercise_name=$(basename $exercise)
     test_count=$((test_count+1))
 
-    printf "\\033[33mTesting\\033[0m: $exercise_name "
+    printf "\\033[33mTesting\\033[0m: "; printf '%s' "${exercise_name} "
 
     exercise_config=".meta/config.json"
     files_to_remove=($(jq -r '.files.solution[]' "${exercise_config}"))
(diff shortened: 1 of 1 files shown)


bin/test_exercises.sh
```diff
@@ -25,7 +25,7 @@ function test_or_tests () {
     then
       tests+="s"
     fi
-    printf "%s" "$tests"
+    printf '%s' "${tests}"
```
dabaer, line 28: We don't actually need to do this here as we know the variable doesn't contain any special characters, but it's done to retain consistency with the rest of our variable-containing-printf-statements.

angelikatyborska (approved):
Thank you, much appreciated! I have seen this error happen many times, but was always too lazy to investigate.

PR #1337, opened 2023-07-09
add missing test cases to "allergies" exercise

elepedus:
Adds test cases for "shellfish", "tomatoes", "chocolate", "pollen" and "cats

### DIFF
--- a/exercises/practice/allergies/test/allergies_test.exs
+++ b/exercises/practice/allergies/test/allergies_test.exs
@@ -30,8 +30,33 @@ defmodule AllergiesTest do
     end
 
     @tag :pending
+    test "allergic to just shellfish" do
+      Allergies.list(4) |> assert_is_a_set_containing(~w[shellfish])
+    end
+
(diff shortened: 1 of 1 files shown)


angelikatyborska:
Hi! Thank you but I have to reject this PR. Tests for practice exercises are defined in a repository shared between all programming languages on Exercism called [problem specifications](https://github.com/exercism/problem-specifications). The maintainers are expected to implement only the test cases specified there. [This is the list of tests specified for the `list` function in this exercise](https://github.com/exercism/problem-specifications/blob/169dfc55e338377c9728c7d88e39b68d84c2262b/exercises/allergies/canonical-data.json#L449-L565). If you feel that this exercise would benefit from more tests, you can start a discussion about that [on the Exercism forum](https://forum.exercism.org/). If a few regular maintainers agree with you on the forum, you could then open a pull request to the problem specifications repository with the new tests, and from there the tests would be propagated to all programming languages on Exercism, Elixir included.

PR #1338, opened 2023-07-11
Adds trailing question marks

mafeka:
Since the documentation mentions that booleans should - by convention - have trailing question marks, it seems helpful to continue that convention in the code.


angelikatyborska:
Hi! Thanks for the contribution. That's a nice little improvement 🙂 could you also make the same changes in the file `.meta/exemplar.ex` and fix the PR title and description to say "question marks"?

angelikatyborska (approved):
Thanks!

PR #1342, opened 2023-07-16
Fix pacman scores logic in description

cshintov:
`and` ---> `or`


cshintov:
My bad! It was unnecessary!

PR #1343, opened 2023-07-18
Improvement in the explanation of the pipe operator (|>)

luciamariaalvarezcrespo:
In this pull request, I propose an improvement in the explanation of the pipe operator (`|>`) in the exercise's `README` file. Currently, the explanation may give the impression that the `|>` operator should be used in every function call, which can be confusing for new Elixir users and those not familiar with functional programming concepts.

I have rewritten the explanation to clarify that the `|>` operator is used to chain functions starting from the second call, and that it is not necessary to use the operator in the first function call. I hope this improvement helps new Elixir users better understand how the pipe operator works and how to use it effectively in their programs.

I appreciate your feedback and suggestions to continue improving the learning material!

Thank you very much for reviewing this pull request!!! 🌷

### DIFF
--- a/exercises/concept/high-school-sweetheart/.docs/introduction.md
+++ b/exercises/concept/high-school-sweetheart/.docs/introduction.md
@@ -44,6 +44,14 @@ Elixir provides many functions for working with strings in the `String` module.
 
 The `|>` operator is called the pipe operator. It is used to chain function calls together in such a way that the value returned by the previous function call is passed as the first argument to the next function call.
 
+It is important to use the pipe operator with prudence and adhere to best practices. Avoid using the pipe operator when performing a single function call, as it can lead to unnecessary complexity. In such cases, it is recommended to call the function directly with its arguments.
+
+For example:
+
+```elixir
+String.split("hello", "")
+```
+
 When using the pipeline, you can place the `|>` operator after the result of a function call to pass its return value as the first argument to the next function. However, there is no need to use the `|>` operator on the first function call, as you can simply call the function with its arguments.
 
 Here's an example:
--- a/exercises/concept/high-school-sweetheart/.docs/introduction.md
+++ b/exercises/concept/high-school-sweetheart/.docs/introduction.md
@@ -63,4 +63,4 @@ Here's an example:
 # => "HELLO?!"
 ```
 
-In this example, we first call `String.upcase/1` with the argument `"hello"`, and then we pass the result of that call `("HELLO")` as the first argument to `Kernel.<>/2` along with the argument `"?!"`.
+In this example, we "pipe" the string `"hello"` into the `String.upcase` function, and the resulting string is piped into `Kernel.<>` as the first argument.
--- a/exercises/concept/high-school-sweetheart/.docs/introduction.md
+++ b/exercises/concept/high-school-sweetheart/.docs/introduction.md
@@ -42,7 +42,11 @@ Elixir provides many functions for working with strings in the `String` module.
 
 ## Pipe Operator
 
-The `|>` operator is called the pipe operator. It can be used to chain function calls together in such a way that the value returned by the previous function call is passed as the first argument to the next function call.
+The `|>` operator is called the pipe operator. It is used to chain function calls together in such a way that the value returned by the previous function call is passed as the first argument to the next function call.
+
+When using the pipeline, you can place the `|>` operator after the result of a function call to pass its return value as the first argument to the next function. However, there is no need to use the `|>` operator on the first function call, as you can simply call the function with its arguments.
+
+Here's an example:
 
 ```elixir
 "hello"
@@ -50,3 +54,5 @@ The `|>` operator is called the pipe operator. It can be used to chain function 
 |> Kernel.<>("?!")
 # => "HELLO?!"
 ```
+
+In this example, we first call `String.upcase/1` with the argument `"hello"`, and then we pass the result of that call `("HELLO")` as the first argument to `Kernel.<>/2` along with the argument `"?!"`.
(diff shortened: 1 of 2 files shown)


exercises/concept/high-school-sweetheart/.docs/introduction.md
```diff
@@ -42,11 +42,17 @@ Elixir provides many functions for working with strings in the `String` module.
 
 ## Pipe Operator
 
-The `|>` operator is called the pipe operator. It can be used to chain function calls together in such a way that the value returned by the previous function call is passed as the first argument to the next function call.
+The `|>` operator is called the pipe operator. It is used to chain function calls together in such a way that the value returned by the previous function call is passed as the first argument to the next function call.
```
dabaer, line 45: I like this adjustment, the original language insinuates that `|>` is but one of the ways to chain functions together, but I don't believe there is another way to really do so.

This would also need to be updated in [pipe-operator/introduction.md](https://github.com/exercism/elixir/blob/main/concepts/pipe-operator/introduction.md).

exercises/concept/high-school-sweetheart/.docs/introduction.md
```diff
@@ -42,11 +42,17 @@ Elixir provides many functions for working with strings in the `String` module.
...
+
+When using the pipeline, you can place the `|>` operator after the result of a function call to pass its return value as the first argument to the next function. However, there is no need to use the `|>` operator on the first function call, as you can simply call the function with its arguments.
```
dabaer, line 47: This seems to conflict with [pipe-operator/about.md](https://github.com/exercism/elixir/blob/main/concepts/pipe-operator/about.md), which notes pipelines such as `String.upcase("hello") |> String.split("")` go against best practice.

It could also cause confusion as we do the opposite in the example below.

exercises/concept/high-school-sweetheart/.docs/introduction.md
```diff
@@ -42,11 +42,17 @@ Elixir provides many functions for working with strings in the `String` module.
...
+Here's an example:
 
 ```elixir
 "hello"
 |> String.upcase()
 |> Kernel.<>("?!")
 # => "HELLO?!"
 ```
+
+In this example, we first call `String.upcase/1` with the argument `"hello"`, and then we pass the result of that call `("HELLO")` as the first argument to `Kernel.<>/2` along with the argument `"?!"`.
```
dabaer, line 58: More concise language could be something like

```
In this example, we "pipe" the string `"hello"` into the `String.upcase` function, and the resulting string is piped into `Kernel.<>` as the first argument.
```

exercises/concept/high-school-sweetheart/.docs/introduction.md
luciamariaalvarezcrespo: Done ✅

exercises/concept/high-school-sweetheart/.docs/introduction.md
luciamariaalvarezcrespo: Thank you for your review and feedback! I appreciate your attention to detail. You are correct that there may be some confusion in the current description of the pipe operator. I addressed this issue by updating the introduction to make it clearer and more consistent with the best practices.

Let me know if you have any other suggestions or concerns :).

exercises/concept/high-school-sweetheart/.docs/introduction.md
luciamariaalvarezcrespo: Done ✅

angelikatyborska:
Hi! Thank you for the contribution 🙏 and thank you @dabaer for doing reviews, much appreciated 👍 

I would like to get the pipe operator description improved, but I'm afraid I disagree with most of what you have added. Let me explain:

> It is important to use the pipe operator with prudence and adhere to best practices.

I don't like how strong this language is. The pipe operator isn't really a dangerous operator so people can get away with using however they want to and not get in trouble. There are two common pitfalls, [operation precedence and anonymous functions](https://hexdocs.pm/elixir/Kernel.html#%7C%3E/2-pitfalls). We could mention the problems with operator precedence if you think that would help?

> Avoid using the pipe operator when performing a single function call, as it can lead to unnecessary complexity

I agree that this is a good practice (supported by those two popular style guides: https://github.com/lexmag/elixir-style-guide#needless-pipeline & https://github.com/christopheradams/elixir_style_guide#avoid-single-pipelines), but I wouldn't say it leads to complexity. It can lead to confusion because most Elixir devs don't use pipes for a single function call, so most Elixir devs will think that somebody forgot to write the second function call. I think we can give this advice without justifying it.

> However, there is no need to use the `|>` operator on the first function call

This is just a preference. [One popular styleguide recommends to always start a pipe chain with a value, not a function call](https://github.com/rrrene/elixir-style-guide#pipe-chains). I don't want to tell people to do it this way or that way as it doesn't really matter.

Let me know what you think and if there's anything else about the pipe operator that we should clarify?

One more detail: the file `exercises/concept/high-school-sweetheart/.docs/introduction.md` is not supposed to be modified manually. All changes need to be made to the `concepts/pipe-operator/introduction.md` file, and then they can be automatically propagated to the other file by running `./bin/configlet generate` (the configlet tool must first be downladed by running `./bin/fetch-configlet`).

dabaer:
> One more detail: the file `exercises/concept/high-school-sweetheart/.docs/introduction.md` is not supposed to be modified manually. All changes need to be made to the `concepts/pipe-operator/introduction.md` file, and then they can be automatically propagated to the other file by running `./bin/configlet generate` (the configlet tool must first be downladed by running `./bin/fetch-configlet`).

Good catch. Sheesh I was even looking for that after I had done it myself before, but GitHub Mobile only shows the first directory in a PR and I didn't even notice 🙃

angelikatyborska:
I haven't received a response to my doubts about this PR in half a year (https://github.com/exercism/elixir/pull/1343#issuecomment-1642258974), so I'm going to close this PR.

PR #1346, opened 2023-08-03
ELIXIR:community-garden add new test to Task4

zulkris:
there are no necessary test for https://exercism.org/tracks/elixir/exercises/community-garden/ which checking the using  `plot_id` param really used in `release` function

f.e. current code works well (no tests fails)
```
  def release(pid, _plot_id) do
    %{next_id: next_id} = Agent.get(pid, & &1)
    Agent.update(pid, fn %{next_id: next_id} -> %{plots: [], next_id: next_id} end)
  end
```

This PR adds new test to check this


angelikatyborska (approved):
Well done! That was a terrible oversight from our side, thank you for fixing it 🙂

PR #1347, opened 2023-08-08
Add extra test to Lasagna

Callisto13:
The exercise can be completed without doing it properly, this extra test ensures against hardcoded values in the solution.


angelikatyborska (approved):
Thank you!

PR #1350, opened 2023-08-29
Improvement of type spec for balance method of bank account exercidse

serhiy-yevtushenko:
I added {:ok, account_closed} to balance type spec, as this is expected output for this exercise according to test


angelikatyborska:
Thank you! Could you make the same change in the example solution?

serhiy-yevtushenko:
Done - added to .meta/example.ex (I hope that this is example solution)

PR #1353, opened 2023-09-04
docs(triangle): fix description of inequality theorem

sevillaarvin:
https://mathworld.wolfram.com/TriangleInequality.html

>  Geometrically, the right-hand part of the triangle inequality states that the sum of the lengths of any two sides of a triangle is **greater than the length of the remaining side**.


angelikatyborska:
Please read the contributing guide before contributing: https://github.com/exercism/elixir/blob/main/CONTRIBUTING.md

> For practice exercises, instructions.md come from the [problem specifications](https://github.com/exercism/problem-specifications) repository. They cannot be changed without updating them in that repository first.

PR #1365, opened 2023-09-15
Sync Parallel Letter Frequency tests with canonical specs

antoine-duchenet:
Hi !

I updated the tests to match the problem specifications.
A minor update of the `@reject_characters` RegExp in the example implementation was needed to pass the new tests.

Relates to #1027 .

### DIFF
--- a/exercises/practice/parallel-letter-frequency/.meta/example.ex
+++ b/exercises/practice/parallel-letter-frequency/.meta/example.ex
@@ -7,7 +7,7 @@ defmodule Frequency do
   The number of worker processes to use can be set with 'workers'.
   """
 
-  @reject_characters ~r/[0-9\s,]/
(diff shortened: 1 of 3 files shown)


jiegillet (commented):
This is great, thanks for doing it!
I only have a small request, could you format the `@dostoevsky` strings with triple quotes:
```elixir
@dostoevsky [
  """"I am a sick man.... I am a spiteful man. I am an unattractive man.
  I believe my liver is diseased. However, I know nothing at all about my disease, and do not
  [....]
  """,
  """
  Gentlemen, I am joking, and I know myself that my jokes are not brilliant
  [...]
  """
  ]
```
And I think it's fine not to worry about an extra newline here or there in this case :)

jiegillet (approved):
Very nice, thank you!

PR #1367, opened 2023-09-16
Add practice exercise `ledger`

jiegillet:
It has been a while since I added an exercise :)
This one got tests added a while back.

It's a refactor exercise, it was my first time doing it, I don't know if I went too far with the bad code.
Some pieces of the code would only work on the given test cases, is that fair game?

### DIFF
--- a/exercises/practice/ledger/lib/ledger.ex
+++ b/exercises/practice/ledger/lib/ledger.ex
@@ -3,12 +3,13 @@ defmodule Ledger do
   Format the given entries given a currency and locale
   """
   @type currency :: :usd | :eur
+  @type locale :: :en_US | :nl_NL
   @type entry :: %{amount_in_cents: integer(), date: Date.t(), description: String.t()}
 
-  @spec format_entries(currency(), String.t(), list(entry())) :: String.t()
+  @spec format_entries(currency(), locale(), list(entry())) :: String.t()
   def format_entries(currency, locale, entries) do
     header =
-      if locale == "en_US" do
+      if locale == :en_US do
         "Date       | Description               | Change       \n"
       else
         "Datum      | Omschrijving              | Verandering  \n"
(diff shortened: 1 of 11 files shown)


exercises/practice/ledger/lib/ledger.ex
```diff
@@ -0,0 +1,103 @@
+defmodule Ledger do
+  @doc """
+  Format the given entries given a currency and locale
+  """
+  @type currency :: :usd | :eur
+  @type entry :: %{amount_in_cents: integer(), date: Date.t(), description: String.t()}
+
+  @spec format_entries(currency(), String.t(), list(entry())) :: String.t()
+  def format_entries(currency, locale, entries) do
```
angelikatyborska, line 9: Maybe the locale could also get its own custom type?

exercises/practice/ledger/test/ledger_test.exs
```diff
@@ -0,0 +1,153 @@
...
+        date: ~D[2015-01-01],
+        description: "Freude schoner Gotterfunken"
+      }
+    ]
+
+    assert Ledger.format_entries(:usd, "en_US", entries) ==
+             """
+             Date       | Description               | Change\s\s\s\s\s\s\s
+             01/01/2015 | Freude schoner Gotterf... |   ($1,234.56)
+             """
```
angelikatyborska, line 86: Did you mean for the boilerplate solution to be failing this test? In the markdown exercise, the boilerplate passes all tests.

angelikatyborska (approved):
Yaaay, Jie is contributing here again 😻 

>  Some pieces of the code would only work on the given test cases, is that fair game?

Absolutely.

I'm approving, feel free to merge with or without making any more changes.

exercises/practice/ledger/lib/ledger.ex
jiegillet: I thought about it, I think I decided not to because somehow I thought capital letter were not allowed in atoms... But that's silly, so I made it a type after all :)

exercises/practice/ledger/test/ledger_test.exs
jiegillet: My bad! I made a value wrong on purpose, but I made it too wrong and somehow forgot to check :(

PR #1368, opened 2023-09-17
Add test for race condition in community gardens

cshintov:
If the `register` function uses `get` and `update` instead of `get_and_update` concurrent operations results in non-unique ids.

### DIFF
--- a/exercises/concept/community-garden/test/community_garden_test.exs
+++ b/exercises/concept/community-garden/test/community_garden_test.exs
@@ -102,4 +102,24 @@ defmodule CommunityGardenTest do
     assert {:ok, pid} = CommunityGarden.start()
     assert {:not_found, "plot is unregistered"} = CommunityGarden.get_registration(pid, 1)
   end
+
+  @tag task_id: 6
+  test "register multiple plots concurrently" do
+    {:ok, pid} = CommunityGarden.start()
+
+    # Spawn 25 processes that register a plot concurrently
+    for _ <- 1..25 do
+      Task.async(fn -> CommunityGarden.register(pid, "user") end)
+    end
+    |> Enum.map(&Task.await/1)
+    |> Enum.map(& &1.plot_id)
+
+    # Get the list of registered plot IDs
+    plot_ids =
+      CommunityGarden.list_registrations(pid)
+      |> Enum.map(& &1.plot_id)
+
+    # Assert that each plot ID is unique
+    assert Enum.uniq(plot_ids) == plot_ids
+  end
 end
--- a/exercises/concept/community-garden/test/community_garden_test.exs
+++ b/exercises/concept/community-garden/test/community_garden_test.exs
@@ -45,6 +45,32 @@ defmodule CommunityGardenTest do
     assert plot_3.plot_id == 3
   end
 
+  @tag task_id: 3
+  test "registered plots have incremental unique id when registered concurrently" do
+    {:ok, pid} = CommunityGarden.start()
+
+    total_plots = 20
+    test_process_pid = self()
+
+    Enum.each(1..total_plots, fn n ->
+      spawn(fn ->
+        plot = CommunityGarden.register(pid, "Mary Bumblebee #{n}")
+        send(test_process_pid, {n, plot})
+      end)
+    end)
+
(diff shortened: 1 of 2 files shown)


angelikatyborska (commented):
Hi! Thank you for bringing this problem to my attention 🙏 

I am unable to write a solution that would fail the new test on my machine. I tried this:

```elixir
  def register(pid, register_to) do
    state = Agent.get(pid, fn state -> state end)
    %{registry: registry, next_id: next_id} = state
    new_plot = %Plot{plot_id: next_id, registered_to: register_to}
    updated = Map.put(registry, next_id, new_plot)
    state = %{state | registry: updated, next_id: next_id + 1}
    Agent.update(pid, fn x -> state end)

    new_plot
  end
```

I suspect that due to the nature of the problem (concurrency and being unlucky with timing), it might be impossible to write a test that would fail reliably if `Agent.get_and_update` was not used.

Maybe it would be a better idea to add an analyzer check that ensures `Agent.get_and_update` is called from `CommunityGarden.register`?

Some links:
- This is the analyzer repository https://github.com/exercism/elixir-analyzer
- This is an example of a similar check for a different exercise: https://github.com/exercism/elixir-analyzer/blob/198d0da32b9efeaf14c2b132e1e14b0523d38aae/lib/elixir_analyzer/test_suite/newsletter.ex#L112-L117
- And this is where the text for the comments is stored: https://github.com/exercism/website-copy/tree/main/analyzer-comments/elixir (different repository)

cshintov:
> Hi! Thank you for bringing this problem to my attention 🙏
> 
> I am unable to write a solution that would fail the new test on my machine. I tried this:
> 
> ```elixir
>   def register(pid, register_to) do
>     state = Agent.get(pid, fn state -> state end)
>     %{registry: registry, next_id: next_id} = state
>     new_plot = %Plot{plot_id: next_id, registered_to: register_to}
>     updated = Map.put(registry, next_id, new_plot)
>     state = %{state | registry: updated, next_id: next_id + 1}
>     Agent.update(pid, fn x -> state end)
> 
>     new_plot
>   end
> ```
> 
> I suspect that due to the nature of the problem (concurrency and being unlucky with timing), it might be impossible to write a test that would fail reliably if `Agent.get_and_update` was not used.

Actually I misunderstood the cause of the problem.

I tried your solution and found no problem with `get` and `update` as you said.  

So I tried my solution again and it was giving me non-unique ids. The problem was not get and update, it was how I used them.

```
  def register(pid, register_to) do
    new_id = Agent.get(pid, & &1.id) + 1

    new_plot = %Plot{plot_id: new_id, registered_to: register_to}
    Agent.update(pid, &%{id: new_id, plots: [new_plot | &1.plots]})

    new_plot
  end
```

I first extract the current latest id, construct new plot with incremented id and then finally makes the update. This causes concurrent register to give non-unique ids. 

Gives error!
non-unique-ids

No errors!
no-error-when-excluded

angelikatyborska:
Sorry for the late reply!

I can see now that your wrong solution (https://github.com/cshintov/exercism/blob/master/elixir/community-garden/lib/community_garden.ex) fails this new test reliably. So I decided to add the new test that you're proposing, but I modified it to incorporate the feedback that I had about it.

@jiegillet Could you take a look if I didn't make any silly mistakes?

exercises/concept/community-garden/test/community_garden_test.exs
```diff
@@ -45,6 +45,32 @@ defmodule CommunityGardenTest do
...
+        send(test_process_pid, {n, plot})
+      end)
+    end)
+
+    plot_ids =
+      Enum.map(1..total_plots, fn n ->
+        receive do
+          {^n, plot} -> plot.plot_id
+        after
+          100 -> nil
```
angelikatyborska, line 67: 20 * 100ms = 2s, so even is somebody messes up their solution so bad they trigger the timeout every time, they should still see some output of the test. I think the test runner only gets killed by exercism after 3 * average_run_time (which is 3 * 4s in our case)

jiegillet (approved):
I didn't dive deep into this one, but the added test makes sense to me and looks sensible

PR #1370, opened 2023-09-18
Fix AllYourBase to follow problem specs

pedrosnk:
For All your base exercise, we need it's valid to return input or output error when both are provided. This commit allows both to be returned from the tested function.

Because we are expecting an error to be either input or output, I'm using a regex on the tests to match any of the errors to be valid.

### DIFF
--- a/exercises/practice/all-your-base/.meta/example.ex
+++ b/exercises/practice/all-your-base/.meta/example.ex
@@ -5,12 +5,12 @@ defmodule AllYourBase do
   """
 
   @spec convert(list, integer, integer) :: {:ok, list} | {:error, String.t()}
-  def convert(_, _, output_base) when output_base < 2 do
-    {:error, "output base must be >= 2"}
-  end
-
   def convert(_, input_base, _) when input_base < 2 do
     {:error, "input base must be >= 2"}
+  end
+
+  def convert(_, _, output_base) when output_base < 2 do
+    {:error, "output base must be >= 2"}
   end
 
   def convert(digits, input_base, output_base) do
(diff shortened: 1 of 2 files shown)


angelikatyborska (commented):
Thank you for bringing this to my attention! Actually according to the Exercism-wide problem specifications, the error should be `"input base must be >= 2"` in case both input and output bases are negative (see https://github.com/exercism/problem-specifications/blob/main/exercises/all-your-base/canonical-data.json#L251-L263).

Could you modify your PR to follow that expectation?

pedrosnk:
TIL about the problem-specifications repo. In fact I don't believe there's much to change on the the tests if the specification must be followed.

Would make sense to update the elixir instructions following the error specification?

angelikatyborska:
> Would make sense to update the elixir instructions following the error specification?

You mean the file `exercises/practice/all-your-base/.docs/instructions.md`? That file is copied from problem specifications word for word and cannot be changed.

What we need is a single word change in the test file (compared to before your initial changes), to change the error message from `'output"` to `"input"`, and then a change in the file `exercises/practice/all-your-base/.meta/example.ex` so that the example solution passes the changed tests.

pedrosnk:
I follow you now, sorry, I thought it was already correct.

Thanks for walking me through it. Just patched the correct fix.

pedrosnk:
test was failing because `exercises/practice/all-your-base/.meta/example.ex` needed to be updated. Just update it and rerun `bin/ci.sh` on my local machine. Seems to be passing now.

angelikatyborska (approved):
Thank you!

PR #1374, opened 2023-09-25
doc: update Collatz Conjecture

icyrainz:
Add explicit handling requirement for input case of 0.
Fix test case string casing.


angelikatyborska:
> The docs changes should be globally done in the problem specs right ?

Correct!

Please read `CONTRIBUTING.md` before opening PRs:

https://github.com/exercism/elixir/blob/main/CONTRIBUTING.md?plain=1#L43

PR #1378, opened 2023-10-05
Test updated for Freelancer Rates: Corrects a invalid test

fparadas:
This fixes one of freelancer rates test.

It was testing if a discount of 20% over a value of 1 was 1.2, the correct value should be 0.8.

### DIFF
--- a/exercises/concept/freelancer-rates/test/freelancer_rates_test.exs
+++ b/exercises/concept/freelancer-rates/test/freelancer_rates_test.exs
@@ -83,7 +83,7 @@ defmodule FreelancerRatesTest do
     @tag task_id: 4
     test "it applies the discount" do
       # 1.25
-      assert FreelancerRates.days_in_budget(480, 60, 20) == 1.2
+      assert FreelancerRates.days_in_budget(480, 60, 20) == 0.8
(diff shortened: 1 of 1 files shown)


exercises/concept/freelancer-rates/test/freelancer_rates_test.exs
```diff
@@ -83,7 +83,7 @@ defmodule FreelancerRatesTest do
     @tag task_id: 4
     test "it applies the discount" do
       # 1.25
-      assert FreelancerRates.days_in_budget(480, 60, 20) == 1.2
+      assert FreelancerRates.days_in_budget(480, 60, 20) == 0.8
```
jiegillet, line 86: I can't find an issue in this test.

budget = 480, hourly_rate = 60, discount = 20%
daily_rate = 8 * 60 = 480
daily_rate_with_discount = 480 - 20% = 384
days_in_budget = 480 / 384 = 1.25 (rounded to 1.2)

Can you point to the problem?

fparadas: You are correct, my bad

PR #1381, opened 2023-10-17
Sync word count with canonical tests

jonathan-h-grebe:
## Summary
This PR syncs the tests for word-count with the canonical data by
- removing tests whose requirements contradict the problem specification as described in instructions.md
  - "hyphens" - this required the solution to treat words containing a hyphen as single words, whereas the instructions state "The only punctuation that does not separate words is the apostrophe in contractions."
  - "German" - this required the solution to handle German script characters, whereas the instructions state "The subtitles from these dramas use only ASCII characters."
  - "Polish - same as with German
- eliminating test description inconsistencies. eg change "count one of each" to "count one of each word"

## References
[discussion on forum](http://forum.exercism.org/t/mismatch-between-specification-and-tests-for-elixirs-word-count-exercise/7707/7)

### DIFF
--- a/exercises/practice/word-count/test/word_count_test.exs
+++ b/exercises/practice/word-count/test/word_count_test.exs
@@ -6,25 +6,25 @@ defmodule WordCountTest do
   end
 
   @tag :pending
-  test "count one of each" do
+  test "count one of each word" do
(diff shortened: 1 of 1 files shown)


angelikatyborska (approved):
Perfect 🙏  thank you

PR #1384, opened 2023-10-26
Run configlet format

angelikatyborska:
Run `./bin/configlet fmt --update` so that the CI starts passing again


jiegillet (approved):
Mmh I love the smell of a fresh automated formatting PR 🌹

PR #1385, opened 2023-10-27
Add Pop Count practice exercise

kahgoh:
Adds [Pop Count](https://forum.exercism.org/t/new-exercise-suggestion-pop-count/7632) exercise

### DIFF
--- a/config.json
+++ b/config.json
@@ -1590,7 +1590,10 @@
           "bit-manipulation"
         ],
         "prerequisites": [
-          "bit-manipulation"
+          "bit-manipulation",
+          "multiple-clause-functions",
+          "pattern-matching",
+          "recursion"
         ],
         "difficulty": 3
       },
--- a/config.json
+++ b/config.json
@@ -1702,6 +1702,18 @@
         "difficulty": 3
       },
       {
+        "slug": "pop-count",
+        "name": "Pop Count",
+        "uuid": "8a9d6e1e-9188-44c9-8681-ade84d340dd0",
+        "practices": [
+          "bit-manipulation"
+        ],
+        "prerequisites": [
+          "bit-manipulation"
+        ],
+        "difficulty": 3
+      },
+      {
         "slug": "word-count",
         "name": "Word Count",
         "uuid": "4a24ba2f-ae92-4095-be53-64bc881422ea",
--- a/config.json
+++ b/config.json
@@ -1583,6 +1583,18 @@
         "difficulty": 3
       },
       {
+        "slug": "pop-count",
+        "name": "Pop Count",
+        "uuid": "8a9d6e1e-9188-44c9-8681-ade84d340dd0",
+        "practices": [
+          "bit-manipulation"
+        ],
+        "prerequisites": [
+          "bit-manipulation"
+        ],
+        "difficulty": 3
+      },
+      {
         "slug": "prime-factors",
         "name": "Prime Factors",
         "uuid": "47328272-b2f6-42e4-a5a3-a5d276bfbdee",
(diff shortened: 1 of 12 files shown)


angelikatyborska:
Hi! Thanks for this.

Two details jump out to me:
- The CI is failing because the expected exercise order in the config file is to sort by difficulty first, then alphabetically. This new exercise should be right after perfect-numbers
- In Elixir, all function names must use `snake_case`

config.json
```diff
@@ -1582,6 +1582,18 @@
...
       },
+      {
+        "slug": "pop-count",
+        "name": "Pop Count",
+        "uuid": "8a9d6e1e-9188-44c9-8681-ade84d340dd0",
+        "practices": [
+          "bit-manipulation"
+        ],
+        "prerequisites": [
+          "bit-manipulation"
```
angelikatyborska, line 1593: I wonder what other concepts are necessary here. Definitely not just bit-manipulation because on the Elixir track, you can learn bit-manipulation together with anonymous functions literally as a second exercise, right after basics.

Your solution uses multiple clause functions, guards, and tail call recursion. I assume it could also be solved with normal recursion? If it could be solved with normal recursion only, that would be better as a prerequisite than tail call recursion as it's easier.

Do you know which language already published this exercise and has some community solutions that I could look at?

jiegillet: There are some published [on the Elm track](https://exercism.org/tracks/elm/exercises/pop-count/solutions/
). 
Some use, tail recursion, recursion, some don't use recursion at all (Elm compile to JS, so you know you have at most 32 bits).

angelikatyborska, line 1593: Alright, I think this is very easy without tail tail recursion, so those are the prerequisites that I think it needs:

```suggestion
          "bit-manipulation",
          "multiple-clause-functions",
          "pattern-matching",
          "recursion"
```

exercises/practice/pop-count/test/pop_count_test.exs
```diff
@@ -0,0 +1,24 @@
...
+
+    @tag :pending
+    test "4 eggs" do
+      assert PopCount.egg_count(89) == 4
+    end
+
+    @tag :pending
+    test "13 eggs" do
+      assert PopCount.egg_count(2_000_000_000) == 13
+    end
```
angelikatyborska, line 22: @jiegillet wrote:

> (Elm compile to JS, so you know you have at most 32 bits).

You cannot make that assumption in Elixir. Maybe this is actually a good case for an Elixir-specific test, to teach people that integer can be arbitrarily big?
```suggestion
    end
    
    @tag :pending
    test "100 eggs" do
      assert PopCount.egg_count(1_267_650_600_228_229_401_496_703_205_375) == 100
    end
```

jiegillet: That's a lot of eggs 🍳

kahgoh: This sounds reasonable to me

exercises/practice/pop-count/.meta/config.json
```diff
@@ -0,0 +1,19 @@
+{
+  "authors": [
+    "kahgoh"
+  ],
```
jiegillet, line 4: Since she suggested a new test, I would suggest to add @angelikatyborska to the list of contributors (currently empty therefore missing)

angelikatyborska: It's fine, I have enough attributions on the Elixir track 😁 time to merge!

angelikatyborska (approved):

angelikatyborska:
Thanks again 🙏

PR #1389, opened 2023-11-14
CI: pin GitHub Actions workflows

ErikSchierboom:
This PR updates GitHub Actions workflows to a specific version.
This ensures that the workflow will always run the same code, which makes your build _stable_.
It will also prevent a potential security issue where a tag could be replaced by a malicious commit without consumers being aware of it.

The PR updates each non-SHA based workflow reference with the SHA of the referenced version/tag, so the current behavior should not change.

See https://exercism.org/docs/building/github/gha-best-practices#h-pin-actions-to-shas for more information.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,10 +22,10 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
 
       - name: Use Elixir
-        uses: erlef/setup-beam@v1.16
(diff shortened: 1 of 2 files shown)


angelikatyborska:
Did you check whether dependabot will respect this convention out of the box? Usually its PRs update tags https://github.com/exercism/elixir/pull/1383

ErikSchierboom:
> Did you check whether dependabot will respect this convention out of the box?

Yes. We've been using this for a couple of years now and it works well (if you'd like, you could even add a version comment at the end and that will be updated too).

angelikatyborska (approved):

PR #1392, opened 2023-11-19
Sync poker

angelikatyborska:
Two new tests, two tests reimplemented. Our implementations of the two reimplemented tests were already deviating from problem specs in that they were each comparing two different pairs of hands in one test 🤔  I just made them exactly like the problem specs ask for.


jiegillet (approved):
Looks good.
Going back to the specs is always a good choice :)

PR #1393, opened 2023-11-25
Add file-sniffer test that fails for bitstrings

angelikatyborska:
I'm currently thinking through this exercise because of the analyzer issues, and I realized that we could use one more test. This new unit test will catch two problems:
- people using `String.starts_with?` (it crashes for non-binaries)
- people using `::bistring` instead of `::binary` (I saw a few of those in community solutions)


jiegillet (approved):
Good ideas!

PR #1395, opened 2023-12-07
Stack underflow: specify 2 numbers are expected

angelikatyborska:
In response to feedback from Discord:

> It's kind of implied, but it's not totally obvious that the divide function should be written with the expectation that it will receive a list of exactly two numbers: a divisor and its denominator. Most of the community solutions i looked at were written according to this assumption, and would fail if given a list of, say, [2, 2, 8], and the test only covers cases with at most two numbers in the list. When i read the exercise I kind of assumed I needed to accept lists of arbitrary length and wrote my implementation with that in mind. Maybe the exercise should make it explicit that only a list of two numbers is considered valid?


jiegillet (approved):
Nice!

PR #1398, opened 2024-01-02
Remove printable charlists from bird-count (lists)

angelikatyborska:
Bird count is one of the first learning exercises, and it's about lists. Sooner or later Elixir learners will hit the confusing behavior of lists of integers being charlists, and thus printable list of integers being always printed as charlists. However, I think we can postpone that happening to a later exercise, ideally the one about charlists 🙂 

This issue was brought to my attention by this forum post: https://forum.exercism.org/t/bird-count-test-case-appears-to-use-sigils/8922

### DIFF
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -5,3 +5,4 @@ github.com/pulls
 twitter.com
 regexr.com
 papyr.com
+reddit.com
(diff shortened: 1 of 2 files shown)


jiegillet (approved):
Sounds good to me

PR #1403, opened 2024-01-07
Sync knapsack

angelikatyborska:
No changes - problem specs changed the type of the input from an empty object to an empty array. We've been using an empty list all along.


jiegillet (approved):
👍

PR #1404, opened 2024-01-07
Sync ledger

angelikatyborska:
There was a factual error in the test that got fixed in problem specs. The test description claimed to use the same date for two entries, but different dates were used.


jiegillet (approved):
Huh. That was missed for a long time  :)

PR #1411, opened 2024-01-14
Fix decimal to binary youtube link

angelikatyborska:
Some youtube video got removed.

According to the way back machine, the removed video is "How to Convert Decimal to Binary" by tecmath and it was 3 minutes long. I found something similar.


jiegillet (approved):
I now know how to convert decimal to binary 4️⃣  -> 💯

PR #1417, opened 2024-01-18
Clarify solution requirements for boutique-suggestions

pisanovo:
Clarify solution requirements for the exercise boutique-suggestions based on the available test cases.

### DIFF
--- a/exercises/concept/boutique-suggestions/.docs/instructions.md
+++ b/exercises/concept/boutique-suggestions/.docs/instructions.md
@@ -77,3 +77,5 @@ BoutiqueSuggestions.get_combinations(tops, bottoms, maximum_price: 50)
 #       %{item_name: "Jeans", base_color: "blue", price: 30}}
 #    ]
 ```
+
+For your final solution you can assume that each top and bottom always includes `:price` and `:base_color` fields.


pisanovo:
I know it states that each clothing item is stored as a map with the following structure, however given task 1 and 2 it might make sense to explicitly state that for the final solution `price` and `base_color` fields are always present.

exercises/concept/boutique-suggestions/.docs/instructions.md
```diff
@@ -77,3 +77,5 @@ BoutiqueSuggestions.get_combinations(tops, bottoms, maximum_price: 50)
 #       %{item_name: "Jeans", base_color: "blue", price: 30}}
 #    ]
 ```
+
+For your final solution you can assume that each top and bottom always includes `:price` and `:base_color` fields.
```
angelikatyborska, line 81: This is already stated by the first sentences of tasks 2 and 3 respectively:

> ## 2. Filter out clashing outfits
> Each piece of clothing has a `:base_color` field, 

> ## 3. Filter by combination price
> Each piece of clothing has a `:price`

It shouldn't be necessary to repeat that.

If you think the examples without prices and base colors are confusing, I would rather extend the examples.

pisanovo: To me it was a bit confusing when my final solution which passes all tests could not run all task examples due to missing fields. I guess the examples could be extended but this would come at the expense of longer descriptions. Not sure if others had the same issue, if not leave it as is.

PR #1418, opened 2024-01-18
Make community-garden Livebook compatible

pisanovo:
For the community-garden exercise Task 5 should include the plot with id 1 after it was released in the task prior.


pisanovo:
Makes sense, thanks for the reply.

PR #1419, opened 2024-01-20
Set up MarkdownCodeBlockFormatter

angelikatyborska:
I wrote an Elixir formatter plugin that is able to format Elixir code in Markdown files. @jiegillet @neenjaw what do you think?

(I'll publish it to hex before merging)


docs/TESTS.md
```diff
@@ -173,7 +176,7 @@ and Erlang types. To create a plt with sensible defaults run:
 $ mix dialyzer --plt
 ```
 
-Finally it can be run with:
+Finally, it can be run with:
```
angelikatyborska, line 179: Not a formatter feature 😛

concepts/keyword-lists/about.md
```diff
@@ -59,6 +61,8 @@ if age >= 16, do: "beer", else: "no beer"
 
 This may look like `if` accepts two arguments, but the `do:` and `else:` pair is actually a single argument - a keyword list. The same code could be written as:
 
+[]: # (elixir-formatter-disable-next-block)
```
angelikatyborska, line 64: Where desired, formatting can be turned off. For example here we want to show off a non-standard way to format this code to prove a point.

mix.lock
```diff
@@ -1,4 +1,6 @@
 %{
   "dialyxir": {:hex, :dialyxir, "1.3.0", "fd1672f0922b7648ff9ce7b1b26fcf0ef56dda964a459892ad15f6b4410b5284", [:mix], [{:erlex, ">= 0.2.6", [hex: :erlex, repo: "hexpm", optional: false]}], "hexpm", "00b2a4bcd6aa8db9dcb0b38c1225b7277dca9bc370b6438715667071a304696f"},
+  "earmark_parser": {:hex, :earmark_parser, "1.4.39", "424642f8335b05bb9eb611aa1564c148a8ee35c9c8a8bba6e129d51a3e3c6769", [:mix], [], "hexpm", "06553a88d1f1846da9ef066b87b57c6f605552cfbe40d20bd8d59cc6bde41944"},
```
angelikatyborska, line 3: Nope, that's not needed, a leftover from other experiments!

neenjaw (approved):
Nice, this looks pretty good!

angelikatyborska:
Part of the problem is that there is no one Markdown standard 😞 This syntax:

```
[]: # (elixir-formatter-disable-next-block)
```

It's an abused reference-style link. More generically, it's:
```
[reference-name]: https://example.com (Some link title)
```

Now, I don't know if all Markdown implementations support reference-style links, but Exercism definitely depends on them. I'm pretty sure that `#` must be a valid link href, and `elixir-formatter-disable-next-block` is a valid link title. The only thing I'm unsure about is an empty reference name. I didn't see any problems so far. I could also be non-empty (could be anything), but `configlet generate` moves all non-empty link references to the bottom of the markdown file which messes up with my intended use 😁

jiegillet (approved):
Outstanding! I love formatters :)

I left a couple of questions.

exercises/practice/leap/.approaches/operators/content.md
```diff
@@ -27,7 +27,7 @@ However, if `left` is *false*, `right` has to be evaluated to determine the outc
 
 Another thing to consider when using Boolean operators is their precedence.
 ```elixir
-true or false and false
+true or (false and false)
```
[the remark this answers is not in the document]
angelikatyborska: Well spotted

PR #1427, opened 2024-01-30
Sync anagram

angelikatyborska:
Two new test cases from https://github.com/exercism/problem-specifications/commit/524e2b3e186a5a307cd55d974e3c71796a947ebd


jiegillet (approved):
Neat. Concerning the new warning `[no important files changed]`, it seems to me that we should not add it, because those cases (especially the last one) are pretty important.
What do you think?

angelikatyborska:
Oh, a warning! Yeah, the warning mentions that changing tests counts as changing important files, so this PR changes important files.

Issue #1432, opened 2024-02-01
Typo In Take-A-Number Deluxe Step 6


jake-abed:
Paragraph three of Step 6 of Take-A-Number Deluxe currently reads:

> Modify the return values of all implemented callbacks (init/1 and all handle_* callbacks) to set a timeout. Use the **_the_** value under the key :auto_shutdown_timeout in the current machine state. Do not add the timeout to the {:stop, reason} return value of init/1 - timeouts only apply after the server has started its receive loop.

Small typo, but there are two "the" words back to back in the step. This could be saved for later, but I would be happy to just submit a quick PR to fix it. The exercise itself is totally fine.

angelikatyborska:
Closed in https://github.com/exercism/elixir/pull/1433

PR #1433, opened 2024-02-02
Fix typo in Take-A-Number Deluxe Step 6

jake-abed:
Paragraph three of step 6 had an additional 'the' in it. Quick and easy tidy up.

Referenced in issue #1432.

Edit: This should be labeled as `x:size/tiny`!


jiegillet (approved):
Awesome, thank you!

PR #1434, opened 2024-02-02
Update `darts` hints to mention Elixir gotchas

### LINKED ISSUE #1430: Darts exercise hint for sqrt workaround doesn't seem to work
I was just trying to solve the "Darts" exercise. The hints include the correct formula to get the distance from origin to any given coordinate on a plane, but the provided workaround for the lack of a native square root function in elixir doesn't seem to work. The spec for `Integer.pow()` is expecting integers for both the base and the exponent arguments. The spec for `Float.pow()` is explicitly expecting a **float** for the base, and a **number** for the exponent. 

For the first test case of (-9, 9)

```elixir
iex(3)> -9 * -9 + 9 * 9
162
iex(4)> Integer.pow(162, 0.5)
** (FunctionClauseError) no function clause matching in Integer.pow/2

    The following arguments were given to Integer.pow/2:

        # 1
        162

        # 2
        0.5

    Attempted function clauses (showing 1 out of 1):

        def pow(base, exponent) when is_integer(base) and is_integer(exponent)

    (elixir 1.16.0) lib/integer.ex:105: Integer.pow/2
    iex:4: (file)
iex(4)> Float.pow(162, 0.5)
** (FunctionClauseError) no function clause matching in Float.pow/2

    The following arguments were given to Float.pow/2:

        # 1
        162

        # 2
        0.5

    Attempted function clauses (showing 1 out of 1):

        def pow(base, exponent) when is_float(base) and is_number(exponent)

    (elixir 1.16.0) lib/float.ex:116: Float.pow/2
    iex:4: (file)
iex(4)>
```

Checking the example.ex in the repo, I see that neither `Integer.pow` or `Float.pow` were used there, but rather the erlang math library. I'm apologize if I am missing something else, as I'm not great at geometry OR elixir.

jiegillet:
Closes #1430

### DIFF
--- a/exercises/practice/darts/.docs/hints.md
+++ b/exercises/practice/darts/.docs/hints.md
@@ -2,6 +2,6 @@
 
 ## General
 
-- The distance of a point `(x, y)` from the center of the target `(0, 0)` can be calculated with `sqrt(x*x + y *y)`
-
-- You can calculate the square root of `x` by raising it to the power of `1/2`. In other words, `sqrt(x) == pow(x, 0.5)`
+- The distance of a point `(x, y)` from the center of the target `(0, 0)` can be calculated with `sqrt(x * x + y * y)`.
+- Elixir doesn't have a built-in square root function, but the Erlang function `:math.sqrt/1` is available.
+- Alternatively, you can calculate the square root of `x` by raising it to the power of `1/2`. In other words, `def sqrt(x), do: Float.pow(x, 0.5)`. Note that `Float.pow/2` only accept floats, integers need to be converted explicitly, for example with `1.0 * x`.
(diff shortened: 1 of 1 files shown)


exercises/practice/darts/.docs/hints.md
```diff
@@ -2,6 +2,6 @@
 
 ## General
 
-- The distance of a point `(x, y)` from the center of the target `(0, 0)` can be calculated with `sqrt(x*x + y *y)`
-
-- You can calculate the square root of `x` by raising it to the power of `1/2`. In other words, `sqrt(x) == pow(x, 0.5)`
+- The distance of a point `(x, y)` from the center of the target `(0, 0)` can be calculated with `sqrt(x * x + y * y)`.
+- Elixir doesn't have a built-in square root function, but the Erlang function `:math.sqrt/1` is available.
+- Alternatively, you can calculate the square root of `x` by raising it to the power of `1/2`. In other words, `def sqrt(x), do: Float.pow(x, 0.5)`. Note that `Float.pow/2` only accept floats, integers need to be converted explicitly, for example with `1.0 * x`.
```
jiegillet, line 7:
```suggestion
- Alternatively, you can calculate the square root of `x` by raising it to the power of `1/2`. In other words, `def sqrt(x), do: Float.pow(x, 0.5)`. Note that `Float.pow/2` only accepts float arguments, integers need to be converted explicitly, for example with `1.0 * x`.
```

angelikatyborska (approved):

PR #1438, opened 2024-02-08
City office > add new solutions to type address_map test

Astating:
The [City Office exercise](https://exercism.org/tracks/elixir/exercises/city-office) teaches about typespec.

Due to the constraints around testing type definitions, the tests are testing them as strings against hard-coded possibilities.

I wish to allow at least `%{(:street | :postal_code | :city) => String.t()}` as an alternative solution to `%{street: String.t(), postal_code: String.t(), city: String.t()}`

### DIFF
--- a/exercises/concept/city-office/test/form_test.exs
+++ b/exercises/concept/city-office/test/form_test.exs
@@ -223,7 +223,14 @@ defmodule FormTest do
         "%{postal_code: String.t(), street: String.t(), city: String.t()}",
         "%{postal_code: String.t(), city: String.t(), street: String.t()}",
         "%{city: String.t(), street: String.t(), postal_code: String.t()}",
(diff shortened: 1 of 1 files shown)


jiegillet:
Hi @Astating.

If my understanding is correct, `%{street: String.t(), postal_code: String.t(), city: String.t()}` is a less strict type than the other ones, since it would allow an address with missing fields like `%{street: "street name"}`. Is that your understanding as well?

Astating:
Hi @jiegillet 

Damn, I thought it would work like a Typescript Record where all the keys in the union would be required and had tested it with Dialyzer. I had errors with partial maps, but I now see it's only because the fields were used in the function. You're right, no errors based on the contract alone.

I'm sorry for all that !
Closing the request 🙈

PR #1440, opened 2024-02-25
Set up DoctestFormatter

angelikatyborska:
I wrote another formatting plugin 😁 I don't feel that this one is very much needed since we barely have any doctests, so feel free to reject this. I wrote the plugin just for fun based on a suggestion from a friend.

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -1,4 +1,4 @@
 [
-  plugins: [MarkdownCodeBlockFormatter],
+  plugins: [MarkdownCodeBlockFormatter, DoctestFormatter],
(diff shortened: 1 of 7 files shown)


jiegillet (approved):
It all looks great, there will never be a good reason to reject automated formatting, even if it's not used very much :)

Is this conference-driven development? :p

angelikatyborska:
> Is this conference-driven development? :p

Haha, in a way 😂  I wrote the initial formatter because I wanted to have it for this repo. After I gave a lightning talk about it, a person from the audience asked if they can format doctests with it, so I also wrote a doctest formatter.

PR #1450, opened 2024-03-12
Add regex visualizer to ressource in elixir/concepts/regular-expressions

cyrilbois:
Hi, I added Cyrilex, a regex visualizer and tester, to ressource in elixir/concepts/regular-expressions

### DIFF
--- a/concepts/regular-expressions/links.json
+++ b/concepts/regular-expressions/links.json
@@ -24,6 +24,10 @@
     "description": "Online Regex Tester - Regexr"
   },
   {
+    "url": "https://extendsclass.com/regex-tester.html",
+    "description": "Online Regex Tester And Visualizer - Cyrilex"
+  },
+  {
     "url": "https://regexcrossword.com/",
     "description": "Regex Crossword"
   }


angelikatyborska:
Hi!

Sorry for not replying earlier, a notification about this issue got lost in my mailbox, which is always full of Exercism-related GitHub notifications 😁 

Could you tell me why you think this website is worth adding as a resource? Does it have more features than the websites that are already listed? There are already 7 links in this file, which is a lot, including https://regex101.com/ and https://regexr.com/ which are both pretty good regex testers/debuggers.

angelikatyborska:
Hmm. I see the graph, but I don't see the option to generate a string from the pattern 🤔 

I don't feel comfortable accepting your PR for two reasons:
- You didn't tell me in the PR that you are the author of the website you want added. It feels to me like you're trying to use Exercism for self-promotion.
- The website has a banner telling me not to use privacy browser extensions, which I find personally annoying.
Screenshot 2024-04-02 at 19-55-15 Online Regex tester and visualizer - Python PHP Ruby JavaScript Java MySQL

cyrilbois:
To generate a string, it is located just above the string to be tested (on the right, see "Generate a string from RegEx (Beta)").
I understand, I should have been more transparent.
Can you tell me which browser you are using (for the error message)?

PR #1452, opened 2024-03-19
Update introduction.md

tommasocerruti:
Add an example also for the DateTime module, making it analogous to the other modules.


angelikatyborska:
Thank you, but we omitted the code sample or the mention of the DateTime's sigil on purpose.

Timezones are out of scope for this exercise because they always require an external dependency with a database of timezones. And we don't use external dependencies in any of the Elixir Exercism exercises.

PR #1455, opened 2024-03-28
Fix typo in the "dig deeper" section of the Leap Year exercise

padawin:
The rule about leap year uses modulo 4, not 5.


angelikatyborska (approved):
Thank you!

PR #1459, opened 2024-04-02
Sync roman-numerals

angelikatyborska:
Reordered existing cases, added one new case (3888)


jiegillet (approved):
Thanks for reordering :)

PR #1460, opened 2024-04-11
Include all IO in the last test

CamilleMo:
This fix adds all the IO that are displayed when the CharacterSheet.run() function is called


angelikatyborska:
This change goes against the intention of the test. The test is named:
> "it inspects the character map"

There is already another test that tests that the questions are printed:
```elixir
    test "it asks for name, class, and level" do
      io =
        capture_io("Susan The Fearless\nfighter\n6\n", fn ->
          RPG.CharacterSheet.run()
        end)

      assert io =~ """
             Welcome! Let's fill out your character sheet together.
             What is your character's name?
             What is your character's class?
             What is your character's level?
             """
    end
```

PR #1471, opened 2024-04-28
Update instructions.md to reflect actual map

codecakes:
The documentation claims:

```
Clothing items are stored as a map:

%{
  item_name: "Descriptive Name",
  price: 99.00,
  base_color: "red"
}
```

While the actual structure is like this in the tests:

``` elixir
%{
  item_name: "Descriptive Name",
  price: 99.00,
  color: "Ochre Red",
  base_color: "red"
}
```


angelikatyborska:
Thanks!

angelikatyborska (approved):

PR #1474, opened 2024-05-14
Extend test for log-parser

hrubi:
I have seen several "Community Solutions" for this exercise which were not correct:

  def valid_line?(line) do
    line =~ ~r/^\[DEBUG|INFO|WARNING|ERROR\]/
  end

This would yield wrong result for some lines, which are now covered by the test.

### DIFF
--- a/exercises/concept/log-parser/test/log_parser_test.exs
+++ b/exercises/concept/log-parser/test/log_parser_test.exs
@@ -35,6 +35,8 @@ defmodule LogParserTest do
     @tag task_id: 1
     test "level must be wrapped in square brackets" do
       assert LogParser.valid_line?("ERROR something really bad happened") == false
+      assert LogParser.valid_line?("[INFO the latest information") == false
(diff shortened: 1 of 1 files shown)


angelikatyborska:
Thank you for pointing this out 🙏

I think adding 4 separate tests for this is a bit of an overkill, so I would prefer to extend the existing test case with a few more examples, like this:

```
    @tag task_id: 1
    test "level must be wrapped in square brackets" do
      assert LogParser.valid_line?("ERROR something really bad happened") == false
      assert LogParser.valid_line?("[INFO something really bad happened") == false
      assert LogParser.valid_line?("DEBUG] something really bad happened") == false
    end
```

That should be enough to catch most of those flawed community solutions, right?

hrubi:
I agree, thanks .

angelikatyborska (approved):

angelikatyborska:
Thanks again!

PR #1481, opened 2024-05-17
Add new practice exercise `game-of-life`

jiegillet:
It's been a while since the last exercise, it's always a pleasure.

I watched the recent 48in24 video about the minesweeper exercise, and I was a bit frustrated not to have seen a higher-order solution using matrix convolution kernels or something, and since this exercise is also about counting neighbors, it was the perfect way to pass my frustration :)

### DIFF
--- a/bin/bootstrap_practice_exercise.exs
+++ b/bin/bootstrap_practice_exercise.exs
@@ -1,8 +1,8 @@
 # Generate all files required for a practice exercise.
 # File content is filled as much as possible, but some, including tests, need manual input.
 #
-# Run the following command from the root of the repo:
-# $ elixir bin/boostrap_practice_exercise.exs complex-numbers
+# Do not run directly, instead use this script:
+# $ bin/generate_practice_exercise.sh complex-numbers
 # Pass the name of the exercise (e. g., "complex-numbers") as an argument
 
 Mix.install([
@@ -143,9 +143,6 @@ module =
 
 ## Step 1: create folder structure
 
-Mix.Generator.create_directory("exercises/practice/#{exercise}")
-Mix.Generator.create_directory("exercises/practice/#{exercise}/.docs")
-Mix.Generator.create_directory("exercises/practice/#{exercise}/.meta")
 Mix.Generator.create_directory("exercises/practice/#{exercise}/lib")
 Mix.Generator.create_directory("exercises/practice/#{exercise}/test")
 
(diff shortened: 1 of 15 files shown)


bin/bootstrap_practice_exercise.exs
```diff
@@ -243,38 +240,6 @@ url =
 :inets.start()
 :ssl.start()
 
-# .docs/instructions.md
-{:ok, {_status, _header, description}} =
-  :httpc.request(:get, {url ++ ~c"/description.md", []}, [], [])
-
-Mix.Generator.create_file("exercises/practice/#{exercise}/.docs/instructions.md", description)
```
jiegillet, line 250: configlet can create these files now, so it's better to let it do it (and also there are `introduction.md` files now that this didn't handle)

exercises/practice/game-of-life/.meta/example.ex
```diff
@@ -0,0 +1,44 @@
...
+  """
+
+  @spec tick(matrix :: list(list(0 | 1))) :: list(list(0 | 1))
+  def tick([]), do: []
+
+  def tick(matrix) do
+    neighbors = count_neighbors(matrix)
+
+    Enum.zip_with(matrix, neighbors, fn row_matrix, row_neighbor ->
+      Enum.zip_with(row_matrix, row_neighbor, &rule/2)
```
jiegillet, line 13: Aw man, Elixir 1.11 doesn't have `Enum.zip_with`, it's the heart and soul of my algorithm :(

jiegillet: I was looking up when 1.17 might come out, and noticed that 1.11 is not maintained security-wise. Maybe that's also a good cutoff for us (says the super biased contributor)

angelikatyborska: That's a nice table! How come I never knew before that Elixir versions have a "security support" lifespan? Feel free to drop 1.11 if it's in the way, but maybe when we ad 1.17, we can keep 1.12.

angelikatyborska: (I'm too lazy to come up with an alternative example solution to this exercise)

jiegillet: > That's a nice table! How come I never knew before that Elixir versions have a "security support" lifespan? Feel free to drop 1.11 if it's in the way, but maybe when we ad 1.17, we can keep 1.12.

There is also this page: https://hexdocs.pm/elixir/compatibility-and-deprecations.html which is probably a better source.

I feel a bit conflicted about removing a whole version just for one function in an example solution. Maybe I'll simply re-define it with a comment.

jiegillet: Seems like [1.17 is on its way,](https://elixirforum.com/t/elixir-v1-17-0-rc-0-released/63764) perfect timing. I'll wait for that.

angelikatyborska:
@jiegillet rebase/merge main? 1.17 is here 🙂

jiegillet:
Excellent suggestion :)

We have a seemingly dead link:
> José Valim, the creator of Elixir, explains [in his 2012 conference talk](https://vimeo.com/53221562) how he built the language for applications to be:

But actually the link works. Maybe it's just an anti-bot policy... What do we do in those cases?

angelikatyborska:
`.lycheeignore`

angelikatyborska (approved):

PR #1484, opened 2024-05-25
Yacht: Reorder 3+2 full house test case in order to lower the chance of passing it coincidentally

esimonov:
The existing versions of "Full house two small, three big" and "Full house three small, two big" test cases share a similarity which may help pass them by chance. Namely, in both cases the group of two is comprised of the element which is a head of the list. This causes the following "solution" to be accepted:

```elixir
def score(:full_house, dice) do
    case Enum.split_with(dice, fn x -> x == hd(dice) end) do
      {[a, a], [b, b, b]} -> 2*a + 3*b
      _ -> 0
    end
end
 ```

The present PR makes a tiny change in the 3+2 test case that will require to also consider an alternative grouping:

```elixir
def score(:full_house, dice) do
    case Enum.split_with(dice, fn x -> x == hd(dice) end) do
      {[a, a], [b, b, b]} -> 2*a + 3*b
      {[a, a, a], [b, b]} -> 3*a + 2*b # <-- This one.
      _ -> 0
    end
 end
 ```
 
 Seems like there's nothing more to it:
  
 ```sh
 ./bin/configlet sync --update --exercise yacht
Updating cached 'problem-specifications' data...
Checking exercises...
The `yacht` exercise has up-to-date docs, filepaths, metadata, and tests!
```

### DIFF
--- a/exercises/practice/yacht/test/yacht_test.exs
+++ b/exercises/practice/yacht/test/yacht_test.exs
@@ -63,7 +63,7 @@ defmodule YachtTest do
 
   @tag :pending
   test "Full house three small, two big" do
-    assert Yacht.score(:full_house, [5, 3, 3, 5, 3]) == 19
+    assert Yacht.score(:full_house, [3, 5, 3, 5, 3]) == 19
   end
 
   @tag :pending


angelikatyborska:
Hi! Thank you for the PR, but I have to reject it. Let me explain.

All practice exercises on Exercism and their test data are defined in a shared central repository called [problem specifications](https://github.com/exercism/problem-specifications/). That repository is very well reviewed, each change requires 3 approvals, and generally people taking care of this repository put in a lot of effort figuring out the best test data for any given exercise.

In this repository, I simply copy-and-paste the work done in the central repository with the test data, any only apply changes if there are Elixir-specific reasons to do so. I do not get involved in analyzing the logic behind the exercises because that would be just too much work for me. So I want to keep the test cases in this repo identical with the ones from the problem specifications repo.

To request a change to the test data of the yacht exercise, you would need to edit [this file](https://github.com/exercism/problem-specifications/blob/main/exercises/yacht/canonical-data.json) and get a PR approved in the problem specifications repo first. Then I would propagate it here.

PR #1485, opened 2024-05-25
Exchange dead link for 'with'

angelikatyborska:
This was the original: https://web.archive.org/web/20150511141513/http://twitter.com/@Joseph_Kain

It's from 2015 and talks about version 1.2 being upcoming 😁 maybe not the best message for beginners. So I decided against linking to its wayback machine copy, and just find a new link.

### DIFF
--- a/concepts/with/links.json
+++ b/concepts/with/links.json
@@ -4,8 +4,8 @@
     "description": "Getting Started guide - With"
   },
   {
-    "url": "http://learningelixir.joekain.com/learning-elixir-with/",
-    "description": "Learning Elixir - Learning Elixir's with"
+    "url": "https://elixirschool.com/en/lessons/basics/control_structures#with-3",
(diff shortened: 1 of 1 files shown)


jiegillet (approved):
Nice, thank you!

> This was the original: https://web.archive.org/web/20150511141513/http://twitter.com/@Joseph_Kain

I don't see anything there.

Anyway, the new link is fine.
How would you feel about adding this link https://hexdocs.pm/elixir/main/code-anti-patterns.html#complex-else-clauses-in-with ?
Mentioning an anti-pattern while people are still learning seems like a good idea

angelikatyborska:
Sorry, I accidentally linked to the guy's twitter account. That was the original: https://web.archive.org/web/20221206143050/http://learningelixir.joekain.com/learning-elixir-with/

Issue #1482, opened 2024-05-22
Take-A-Number not enough information


mallowmew:
Take-A-Number is the first exercise in the PID and Processes concepts. I don't think Take-A-Number gives enough information to allow for success as a first introduction to these concepts.

Note: I have been working through each concept by doing the introductory exercise and then moving onto the next concepts - I'm saving the 'practice' exercises for when I know the whole toolkit.

- The introduction to the concept / exercise does not show a good example of where actually to put the `receive do ... end` block, and neither does [the Elixir documentation](https://hexdocs.pm/elixir/processes.html). 
- Even when your `receive do ... end` block is inside the function supplied to `spawn`, the examples in the introduction and [the Elixir docs](https://hexdocs.pm/elixir/Kernel.html#spawn/1) only show supplying an anonymous function `fn` as parameter. 
- 'Use recursion' won't just work with an anonymous function.
- The syntax for passing a named function in Elixir is not obvious. Both syntaxes that I found can lead to a solution but I was stuck here for a long time on something that doesn't seem directly related to what the exercise is trying to teach.
- Before you do Task 4 (Stop) you can't run the tests to see if you're getting closer to the solution.

angelikatyborska:
Hi! Thank you so much for the feedback 🙏 

## Re: receive/do

> The introduction to the concept / exercise does not show a good example of where actually to put `the receive do ... end` block

The introduction describes where to put `the receive do ... end` block. It says:

> It is a common pattern to implement a recursive function, for example named `loop`, that calls `receive/1`, does something with the message, and then calls itself to wait for more messages.

## Re: passing a named function

> The syntax for passing a named function in Elixir is not obvious. 

You're absolutely right, we forgot to mention anywhere in the course how to pass a named function as an argument 😱 I'll create a PR to fix that.

## Re: cannot test before task 4 implemented

> Before you do Task 4 (Stop) you can't run the tests to see if you're getting closer to the solution.

Can you give me more details about that? How exactly did your partial solutions look like, and what was preventing you from running the tests? I find it possible to run tests on partial solutions for every step of the exercise. I'm attaching videos to demonstrate.


https://github.com/exercism/elixir/assets/7852553/f11728d1-f848-4076-b4c4-765aa67ed203


https://github.com/exercism/elixir/assets/7852553/ea32a4fb-9c4e-4268-afe4-a20cb6361df2


[...]

PR #1486, opened 2024-05-25
Take-a-number needs a named function example

angelikatyborska:
Partially addresses https://github.com/exercism/elixir/issues/1482

In take-a-number, it was difficult to figure out how to pass a named function to `spawn`. The example in the introduction used an anonymous function, and we actually don't teach anywhere how to use the capture operator to pass named functions as arguments. The only place this is mentioned, is the anonymous function's concept's "about" page, which most people will never see (rendered only if you browse the concept after learning it). None of the concept introductions mention how to use the capture operator for named functions.

### DIFF
--- a/concepts/processes/introduction.md
+++ b/concepts/processes/introduction.md
@@ -5,11 +5,11 @@ In Elixir, all code runs inside processes.
 By default, a function will execute in the same process from which it was called. When you need to explicitly run a certain function in a new process, use `spawn/1`:
 
 ```elixir
-spawn(fn -> 2 + 2 end)
+spawn(&my_function/0)
 # => #PID<0.125.0>
 ```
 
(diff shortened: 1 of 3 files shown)


jiegillet (approved):
Small increment, but worth doing, thank you.

Don't forget to make the same change in the original into/about files.

angelikatyborska:
A bit late, but I noticed now that the exemplar solution doesn't even use function capturing 🙃 I'm afraid this change might have added even more confusion...

```elixir
defmodule TakeANumber do
  def start() do
    spawn(fn -> loop(0) end)
  end

  defp loop(state) do
    receive do
      {:report_state, sender_pid} ->
        send(sender_pid, state)
        loop(state)

      {:take_a_number, sender_pid} ->
        state = state + 1
        send(sender_pid, state)
        loop(state)

      :stop ->
        nil

      _ ->
        loop(state)
    end
  end
end
```

PR #1493, opened 2024-06-11
fix the disable comments for the markdown formatter

DerMolly:
According to docs these comments should look like this. Hopefully this will prevent them from showing up on the exercism page e.g. <https://exercism.org/tracks/elixir/exercises/rpn-calculator> in the section "Try/Rescue"

CleanShot 2024-06-11 at 23 59 25@2x

See:
https://hexdocs.pm/markdown_code_block_formatter/readme.html#disabling-the-formatter-for-specific-code-blocks

### DIFF
--- a/concepts/docs/about.md
+++ b/concepts/docs/about.md
@@ -74,7 +74,7 @@ If you have Elixir installed on your computer, you can use it in [the interactiv
 
 In `iex`, you can type [`h`][iex-h], followed by a space and a module name or a function name, to read its documentation.
 
-[]: # (elixir-formatter-disable-next-block)
+[//]: # (elixir-formatter-disable-next-block)
 
 ```elixir
 iex()> h String.upcase/1
--- a/concepts/keyword-lists/about.md
+++ b/concepts/keyword-lists/about.md
@@ -21,7 +21,7 @@ Keyword.keyword?([{"month", "April"}])
 
 If you want to use characters other than letters, numbers, and `_` in the key, you need to wrap it in quotes. However, that does not make it a string - it is still an atom.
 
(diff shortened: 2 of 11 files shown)


angelikatyborska:
Ooops, I didn't notice that they actually show up on the website 🙈 

If we change them from `[]` to `[//]`, then running `./bin/configlet generate` moves the fake comment to the end of the generated file (`exercises/concept/rpn-calculator/.docs/introduction.md`). Which causes the CI to fail (the CI checks that all autogeneated files are up to date), and makes the comment useless because the comment's position in the file is important.

@ErikSchierboom can you advise, why do empty links without an url show up on the website? E.g. https://exercism.org/tracks/elixir/exercises/leap/approaches/operators

Screenshot 2024-06-12 at 06 20 12

ErikSchierboom:
> @ErikSchierboom can you advise, why do empty links without an url show up on the website?

It looks like that is the default behavior of our markdown library

DerMolly:
So should we remove the comments altogether then? Or do you think it feasable to keep them and make them work like intended. Maybe some option in the markdown library / configlet or something.

ErikSchierboom:
> Maybe some option in the markdown library

I'm positive there isn't such an option available. We could manually override the default behavior, although I'm not sure that is possible. @iHiD what do you think?

angelikatyborska:
We'll have to live with forced autoformatting in concept introductions 🤷 it's not the worst thing ever. I pushed one commit that updates that one code snippet that was resisting formatting before.

@DerMolly thank you for bringing this issue to our attention 🙏

angelikatyborska (approved):

PR #1496, opened 2024-06-17
Fix hint in kitchen calculator

meatball133:
The third hint doesn't follow what you are suppose to in that task


angelikatyborska (approved):

angelikatyborska:
Thanks!

PR #1501, opened 2024-07-05
Remote Control Car: Add pattern matching notion in introduction

Nanonej:
This notion is asked in several task of this exercise but not introduced, so here is an introduction suggestion.


angelikatyborska:
Thanks a lot 🙏 that's a very good point. I rephrased it a bit so that the mention of pattern matching is its own section because it doesn't fit in "Accessing fields and updating"

angelikatyborska (approved):

PR #1510, opened 2024-08-07
Fixed Typos in Instructions of Pig Latin

Lasamat:
Fixed 2 typos in Pig-Latin.
I checked with [Problem Specifications](https://github.com/exercism/problem-specifications/blob/main/exercises/pig-latin/instructions.md) and although they differ generally by wording the typos are not in that one.

I don't know if they are supposed to be identical but they are not and I would atleast like to fix the typos.

### DIFF
--- a/exercises/practice/pig-latin/.docs/instructions.md
+++ b/exercises/practice/pig-latin/.docs/instructions.md
@@ -19,7 +19,7 @@ For example:
 
 ## Rule 2
 
-If a word begins with a one or more consonants, first move those consonants to the end of the word and then add an `"ay"` sound to the end of the word.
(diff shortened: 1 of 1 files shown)


angelikatyborska:
> I don't know if they are supposed to be identical but they are not and I would atleast like to fix the typos.

Yes, they are supposed to be identical, and they get updated automatically with a tool called "configlet". See this github issue if you want to learn how to update all docs of an exercise: https://github.com/exercism/elixir/issues/1027

I can see that your PR changes `pig-latin` instructions to be exactly the same as they would be if updated via configlet, so it's good to merge 👍 thanks

angelikatyborska (approved):

PR #1511, opened 2024-08-10
Refine Take-A-Number

angelikatyborska:
Resolves https://github.com/exercism/elixir/issues/1482

This PR reverts https://github.com/exercism/elixir/pull/1486 - see my new comment in that PR to know why.

This PR re-distributes work between step 1 and step 2. Previously, step 1 assumed you would already start the receive loop. That doesn't make sense because there is no unit test for step 1 that could test that you have done this correctly. So I'm moving that to step 2.

### DIFF
--- a/exercises/concept/take-a-number/.docs/hints.md
+++ b/exercises/concept/take-a-number/.docs/hints.md
@@ -8,8 +8,7 @@
 ## 1. Start the machine
 
 - The machine should run in a new process. There is [a built-in function that starts a new process][kernel-spawn-1].
-- You will need another function that the new process will execute. You can name it, for example, `loop`.
-- Use the [capture operator][special-forms-capture] to pass a named function as an argument.
+- You will need another function that the new process will execute.
 
 ## 2. Report the machine state
 
@@ -39,4 +38,3 @@
 [kernel-spawn-1]: https://hexdocs.pm/elixir/Kernel.html#spawn/1
 [kernel-receive]: https://hexdocs.pm/elixir/Kernel.SpecialForms.html#receive/1
 [kernel-send]: https://hexdocs.pm/elixir/Kernel.html#send/2
-[special-forms-capture]: https://hexdocs.pm/elixir/Kernel.SpecialForms.html#&/1
--- a/exercises/concept/take-a-number/.docs/hints.md
+++ b/exercises/concept/take-a-number/.docs/hints.md
@@ -8,15 +8,17 @@
 ## 1. Start the machine
 
 - The machine should run in a new process. There is [a built-in function that starts a new process][kernel-spawn-1].
-- You will need another function that the new process will execute.
+- Use a 0-arity anonymous function when starting a new process. This function doesn't need to do anything yet.
 
 ## 2. Report the machine state
 
-- The machine's process needs to respond to messages.
+- You will need a new named function. You can name it, for example, `loop`.
+- This new function should accept one argument, the state.
+- The 0-arity anonymous function from the previous step can call the new named function, passing the initial state.
+- The new named function needs to respond to messages.
 - There is [a built-in function that waits for a message to arrive in the process's mailbox][kernel-receive].
 - There is [a built-in function that sends a message to another process][kernel-send].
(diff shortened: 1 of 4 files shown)


exercises/concept/take-a-number/.docs/hints.md
```diff
@@ -8,16 +8,17 @@
...
-- You will need another function that the new process will execute. You can name it, for example, `loop`.
-- Use the [capture operator][special-forms-capture] to pass a named function as an argument.
+- Use a 0-arity anonymous function when starting a new process. This function doesn't need to do anything yet.
 
 ## 2. Report the machine state
 
-- The machine's process needs to respond to messages.
+- You will need a new named function. You can name it, for example, `loop`.
+- This new function should accept one argument, the state.
+- The 0-arity anonymous function from the previous step can call the new named function, passing the initial state.
```
angelikatyborska, line 17: I hope those instructions are explicit enough to lead people to a solution similar to the exemplar

exercises/concept/take-a-number/.docs/instructions.md
```diff
@@ -15,12 +15,15 @@ Note that each time you run this code, the PID may be different.
 
 ## 2. Report the machine state
 
-Modify the machine so that it can receive `{:report_state, sender_pid}` messages. It should send its current state (the last given out ticket number) to `sender_pid` and then wait for more messages.
+Modify the machine so that the newly spawned process is ready to receive messages (start a _receive loop_) with an initial state of `0`. It should be able to receive `{:report_state, sender_pid}` messages. As a response to those messages, it should send its current state (the last given out ticket number) to `sender_pid` and then wait for more messages.
```
angelikatyborska, line 18: I hope that the phrase "receive loop" will remind people of the paragraph with the same title from the introductions

exercises/concept/take-a-number/.docs/instructions.md
```diff
@@ -15,12 +15,15 @@ Note that each time you run this code, the PID may be different.
...
 ```elixir
 machine_pid = TakeANumber.start()
+
+# a client sending a message to the machine
 send(machine_pid, {:report_state, self()})
 
+# a client receiving a message from the machine
 receive do
   msg -> msg
 end
```
angelikatyborska, line 29: Are those comments helpful? I was staring at this snippet and I was afraid that people might not understand that this code is an example code of how to test the implementation of the machine, and not the code that should go into the implementation of the machine 🤔

exercises/concept/take-a-number/.docs/instructions.md
jiegillet: Have you considered adding some example code in the receive loop section? I think code will make it a lot more memorable. Something like:

```elixir
def loop(ping_count) do
  receive do
    {:ping, sender_pid} ->
      send(sender_pid, :pong)
      loop(ping_count + 1)
  end
end
```

It's pretty close to the solution though.

exercises/concept/take-a-number/.docs/instructions.md
jiegillet: I see your concern, but that's usually what those code snippets do.
Maybe if you wanted to be even more explicit, you could frame those as tests? 
```elixir
  test "reports its own state" do
    pid = TakeANumber.start()
    send(pid, {:report_state, self()})
    assert_receive 0
  end
```
but that might not help so much. 
I think adding the comments is fine for now.

exercises/concept/take-a-number/.docs/hints.md
jiegillet: I think they should be

jiegillet (approved):
It looks like a nice improvement.
Sorry I didn't look too closely at #1486.

exercises/concept/take-a-number/.docs/instructions.md
angelikatyborska: Yes, many times, but it is clearly against the recommendation:

> Code examples should only be used to introduce new syntax (students should not need to search the web for examples of syntax). In other cases provide descriptions or links instead of code.

Defining a recursive function is not new syntax...

PR #1513, opened 2024-08-10
Use the type testing functions

mikegehard:
This is more explicit about what the expectation is.

### DIFF
--- a/exercises/concept/freelancer-rates/.docs/introduction.md
+++ b/exercises/concept/freelancer-rates/.docs/introduction.md
@@ -45,7 +45,18 @@ However, when doing division, the result will always be a float, even if only in
 
 To convert a float to an integer, you can discard the decimal part with [`trunc/1`][trunc-1].
 
+There are functions to check if a value is a specific type. These are defined in
+[`Kernel`][kernel-type-check-operators].
+
+```elixir
+is_integer(3)
+# => true
+```
+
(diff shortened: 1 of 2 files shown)


angelikatyborska:
Thank you for the suggestion, but I disagree with the change. I think the change makes the test output in case of a failure less understandable, and it also removes the extra check of what the actual value is.

Before your changes, a failed test shows the expected number and the returned number line by line, with a highlighted `.0`, making the difference very clear. You can see immediately that your implementation returned an integer and not a float.

Screenshot 2024-08-11 at 12 27 58

After your changes, a failed test's highlighted part says "expected true, got false", which is a bit unclear for a function that returns a number, not a boolean. You have to read the whole test to notice that it calls `is_float`. Additionally, this failure doesn't tell you at all what is the type of the value that you returned. It only tells you that you didn't return a float.

Screenshot 2024-08-11 at 12 27 47


For this reason, I have to reject this PR.

mikegehard:
Thanks for the feedback. Is there any room for a discussion on the pros/cons of each approach or is this decision final?

angelikatyborska:
I consider this to be a very minor issue and would rather not spend more of my time on it, but if you have more thoughts on the topic, you can share them.

mikegehard:
Gotcha. I don't want to take up more of your time.

PR #1514, opened 2024-08-12
Sync metadata and docs

angelikatyborska:


jiegillet (approved):
Nice!
I updated the sync since it took me so long to review :D

PR #1515, opened 2024-08-12
Sync protein-translation

angelikatyborska:
- One new test added
- One test, for "AAA", removed, because "AAA" is a known amino acid and shouldn't cause errors. The test for "XYZ" is taking over as the main test for "invalid codon" (see: https://forum.exercism.org/t/incorrect-test-in-protein-translation/10616/13)


jiegillet (approved):
Thank you!, looks good

Issue #1517, opened 2024-08-12
'Series' problem template inconsistencies to other problems.


Lasamat:
In the starter template the parameters 's' and 'size' have underlines before them. This looks like an error.

```elixir
defmodule StringSeries do
  @doc """
  Given a string `s` and a positive integer `size`, return all substrings
  of that size. If `size` is greater than the length of `s`, or less than 1,
  return an empty list.
  """
  @spec slices(s :: String.t(), size :: integer) :: list(String.t())
  def slices(_s, _size) do
  end
end
```

angelikatyborska:
Thank you for the bug report!

This is not an error. Argument names can start with an underscore to mark that they are supposed to be unused arguments. But most of the exercises don't use that feature in the solution stub file, so we can remove it from `series` as well.

PR #1518, opened 2024-08-13
Do not use ignored arguments in solution boilerplate

angelikatyborska:
Resolves https://github.com/exercism/elixir/issues/1517

We don't use ignored arguments in solutions boilerplate for most exercises, so we shouldn't use it in any of them. I was grepping for `\(_`, and `, _` to find underscored arguments in practice exercises and only found those two.


jiegillet (approved):
Nice!
I also took a look and didn't find other arguments we should change.

PR #1524, opened 2024-09-25
Changed secret.add example values to avoid further confusion

### LINKED ISSUE #1521: `secrets` has confusing tests and examples by using the same value for two different variables
This:
```elixir
    test "add 3" do
      add = Secrets.secret_add(3)
      assert add.(3) === 6
    end
```

and this:

```elixir
adder = Secrets.secret_add(2)
adder.(2)
# => 4
```

has driven a student to attempt solutions like this:
```elixir
  def secret_add(secret) do
    fn x -> x + x end
  end
# or
  def secret_add(secret) do
    fn x -> secret + secret end
  end
```

We should change the examples to use different values.

Newters:
This alters the values of the secret.add function task to make it more clear for future students. Resolves #1521.

### DIFF
--- a/exercises/concept/secrets/test/secrets_test.exs
+++ b/exercises/concept/secrets/test/secrets_test.exs
@@ -3,7 +3,7 @@ defmodule SecretsTest do
 
   describe "secret_add" do
     @tag task_id: 1
-    test "add 3" do
+    test "add 4" do
       add = Secrets.secret_add(4)
       assert add.(3) === 7
     end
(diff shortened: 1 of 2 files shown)


exercises/concept/secrets/test/secrets_test.exs
```diff
@@ -4,8 +4,8 @@ defmodule SecretsTest do
   describe "secret_add" do
     @tag task_id: 1
     test "add 3" do
```
angelikatyborska, line 6:
```suggestion
    test "add 4" do
```

angelikatyborska:
Thank you!

angelikatyborska (approved):

PR #1529, opened 2024-10-05
Update instructions.md

srcsintheta:
Exercise should state restrictions of the online tests.

### DIFF
--- a/exercises/concept/bird-count/.docs/instructions.md
+++ b/exercises/concept/bird-count/.docs/instructions.md
@@ -6,6 +6,8 @@ You decided to bring your bird watching to a new level and implement a few tools
 
 You have chosen to store the data as a list of integers. The first number in the list is the number of birds that visited your garden today, the second yesterday, and so on.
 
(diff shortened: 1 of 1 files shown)


angelikatyborska:
Hi! Thank you for the contribution.

Bird count is a learning exercise. When you start it, you are first presented with an introduction to the concept it's supposed to teach (in this case recursion), and only then with the exercise instructions. This order of content and emphasis on the concept should imply that the concept must be used to solve the exercise. This is true of every single learning exercise, for all programming languages, not just Elixir. It's a basic assumption for all learning exercises and shouldn't need to be repeated in each exercise's introductions.

Additionally, learning exercises are meant for people completely new to the specific programming language. When writing content for learning exercises, I need to assume that the reader is not familiar with any Elixir concepts except for the ones already covered by previous learning exercises. Adding the note you're suggesting would be mentioning concepts that learners haven't heard about before, which might cause confusion and fear that they missed or forgot something.

For those two reasons, I feel like it would not be beneficial to accept the changes you're proposing. Sorry!

PR #1530, opened 2024-10-05
Update instructions.md

ghost:
The description as is doesn't make sense. It states at function 2: "Store the default initial score in a module attribute. It will be needed again."

The reset function however simply resets to 0, and implementing so passes all tests. There's no module attribute needed, neither to be defined nor to be used at any point. The enforcement to use one when adding a player is either arbitrary or not properly reflected in the exercise structure as is.

This pull request provides the minimal needed rewording to avoid confusion, stating outright a default attribute shall be used for a reset value.

### DIFF
--- a/exercises/concept/high-score/.docs/instructions.md
+++ b/exercises/concept/high-score/.docs/instructions.md
@@ -1,6 +1,6 @@
 # Instructions
 
-In this exercise, you're implementing a way to keep track of the high scores for the most popular game in your local arcade hall.
+In this exercise, you're implementing a way to keep track of the high scores for the most popular game in your local arcade hall. Note: Store a reset value (value to reset to) as an attribute and use in your functions where sensible (at least in the reset_score() function).
 
 ## 1. Define a new high score map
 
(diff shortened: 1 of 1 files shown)


jiegillet:
Hi.

> The description as is doesn't make sense. It states at function 2: "Store the default initial score in a module attribute. It will be needed again."
> 
> The reset function however simply resets to 0, and implementing so passes all tests. There's no module attribute needed, neither to be defined nor to be used at any point. The enforcement to use one when adding a player is either arbitrary or not properly reflected in the exercise structure as is.

It's true that a solution without module attributes passes the tests, but we have an analyzer rule to check that the module attribute is used ([rule](https://github.com/exercism/elixir-analyzer/blob/main/lib/elixir_analyzer/test_suite/high_score.ex) and [tests](https://github.com/exercism/elixir-analyzer/blob/main/test/elixir_analyzer/test_suite/high_score_test.exs)). Does it not work properly?

What we are trying to do here is to identify a common value so that it can be reused, and defining it in an idiomatic way (module attribute).

> This pull request provides the minimal needed rewording to avoid confusion, stating outright a default attribute shall be used for a reset value.

I'm all for avoiding confusion, but to be honest I'm not sure your version does the trick. We should not mention  in the introduction what students have to do in step 4.

What we could do to address your point is add something in step 4 along the lines of "The reset value (zero) is the same as the default initial score you defined earlier, reuse the same module attribute."
What do you think?

PR #1536, opened 2024-11-25
Updates secrets for order of operations clarity

druyang:
The original example is slightly confusing because division and multiplication can be swapped in some cases and does not clearly show the order of operations:

* The expected output (14) could be achieved through either (6 * 7) / 3 or (6 / 3) * 7. 
* As a result you could write both `secret_function2.(secret_function1.(x))` or `secret_function1.(secret_function2.(x))`, but only 1 will pass test cases when the anonymous functions have varying order of operations. 
* In the revised version, the order matters and is clear (3+1=4, then 4\*2=8 is different from 3\*2=6, then 6+1=7)


angelikatyborska (approved):
That's a very good point. Thank you for the improvements 💜

PR #1538, opened 2024-12-09
Sync docs, metadata and tests

jiegillet:
This PR syncs the docs and metadata with configlet, adds some new test cases, and re-implements one.
I didn't do a separate PR for the tests because they are straighforward.


angelikatyborska (approved):
Looks great, indeed simple changes

PR #1539, opened 2024-12-09
[#1528 ] Remove mentions and uses of `unless`

### LINKED ISSUE #1528: Deprecate `unless` in Elixir 1.18
In Elixir 1.18, [`unless` is being soft-deprecated](https://github.com/elixir-lang/elixir/pull/13851). We should remove it from our syllabus and from a few of the example solutions.

jiegillet:
Closes #1528 

It's a bit unfortunate that some links (like `https://hexdocs.pm/elixir/case-cond-and-if.html#if-unless`) mention unless in the hash, but it'll be a quick fix. We could also wait for 1.18 before merging this.

### DIFF
--- a/exercises/concept/name-badge/.docs/hints.md
+++ b/exercises/concept/name-badge/.docs/hints.md
@@ -2,7 +2,7 @@
 
 ## General
 
-- Read about `if` in the official [Getting Started guide][getting-started-if-unless] or on [elixirschool.com][elixirschool-if-unless].
+- Read about `if` in the official [Getting Started guide][getting-started-if] or on [elixirschool.com][elixirschool-if].
 
 ## 1. Print a badge for an employee
 
@@ -21,5 +21,5 @@
 
 [kernel-if]: https://hexdocs.pm/elixir/Kernel.html#if/2
 [getting-started-basic-strings]: https://hexdocs.pm/elixir/basic-types.html#strings
-[getting-started-if-unless]: https://hexdocs.pm/elixir/case-cond-and-if.html#if-unless
-[elixir-school-if-unless]: https://elixirschool.com/en/lessons/basics/control-structures/#if-and-unless
+[getting-started-if]: https://hexdocs.pm/elixir/case-cond-and-if.html#if-unless
+[elixirschool-if]: https://elixirschool.com/en/lessons/basics/control-structures/#if-and-unless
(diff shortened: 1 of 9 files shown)


exercises/concept/name-badge/.docs/hints.md
```diff
@@ -2,7 +2,7 @@
 
 ## General
 
-- Read about `if` in the official [Getting Started guide][getting-started-if-unless] or on [elixirschool.com][elixirschool-if-unless].
+- Read about `if` in the official [Getting Started guide][getting-started-if] or on [elixirschool.com][elixirschool-if].
```
jiegillet, line 5: The Elixir School links was broken

exercises/concept/name-badge/.docs/hints.md
```diff
@@ -21,5 +21,5 @@
 
 [kernel-if]: https://hexdocs.pm/elixir/Kernel.html#if/2
 [getting-started-basic-strings]: https://hexdocs.pm/elixir/basic-types.html#strings
-[getting-started-if-unless]: https://hexdocs.pm/elixir/case-cond-and-if.html#if-unless
-[elixir-school-if-unless]: https://elixirschool.com/en/lessons/basics/control-structures/#if-and-unless
+[getting-started-if]: https://hexdocs.pm/elixir/case-cond-and-if.html#if-unless
+[elixirschool-if]: https://elixirschool.com/en/lessons/basics/control-structures/#if-and-unless
```
angelikatyborska, line 25: For some reason, the hash has a number at the end. It doesn't work without it:
```suggestion
[elixirschool-if]: https://elixirschool.com/en/lessons/basics/control-structures/#if-and-unless-0
```

angelikatyborska:
I'm fine with either merging it right now and then creating a new ticket to update the hash locations in the docs to the new ones without `unless` (looks like it will be https://hexdocs.pm/elixir/main/case-cond-and-if.html#if). Or just keep this PR open until 1.18 is released and update the links before merging. Whichever you prefer 🙂

jiegillet:
I think we can merge it, I can write an update on the ticket open and keep it open.
It's probably the same amount of work to do the change on `main` than on this branch. Plus, if we keep even one person from looking up unless, it's worth it :)

angelikatyborska (approved):

PR #1544, opened 2025-01-15
exercises/practice/anagram: add an append with instructions about the order of the return values

IsaacG:


neenjaw:
Why this approach rather than change the test to not be dependent on the order?

neenjaw:
I went digging to find the forum thread which prompted this PR, would have been nice to have more description with the motivation here.  I don't think the order should matter, I will submit a candidate PR with a fixed test

IsaacG:
Clarify what this track expects. Context: https://forum.exercism.org/t/14816/

IsaacG:
> I went digging to find the [forum thread](http://forum.exercism.org/t/anagram-exercise-instructions-and-tests-dont-align-on-ordering/14816) which prompted this PR, would have been nice to have more description with the motivation here. I don't think the order should matter, I will submit a candidate PR with a fixed test

You're absolutely correct. My apologies for the inconvenience. Would you like me to update this PR to reflect that the tests will not care about the order? Or this PR can be closed, leaving the instructions ambiguous and leaving it to the tests to impose (or not) any requirements.

neenjaw:
closed in favor of PR

PR #1545, opened 2025-01-16
patch: remove order dependency from `anagram` test cases

neenjaw:
It was raised in the form that the canonical problem specifications have several flaws:
- vague language
- are not congruent with the existing tests in many tracks

This change list removes the dependency on the order of the returned words that have been validated as candidates.

### DIFF
--- a/exercises/practice/anagram/test/anagram_test.exs
+++ b/exercises/practice/anagram/test/anagram_test.exs
@@ -10,7 +10,7 @@ defmodule AnagramTest do
   @tag :pending
   test "detects two anagrams" do
     matches = Anagram.match("solemn", ~w(lemons cherry melons))
-    assert matches == ~w(lemons melons)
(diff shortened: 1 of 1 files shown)


angelikatyborska:
I really don't want to read the discussion about this on the forum because it's 60 posts... so let me just quickly ask your opinion: do you really believe this change improves the exercise for Elixir learners? If yes, I'll approve it just to avoid further discussions. But in my opinion, it is common practice that a function that receives a list and needs to filter it, is supposed to return the items in the same relative order. I think the change to the exercise takes away that learning moment 🤷

neenjaw:
No, i don't think this change meaningfully improves the life of a learning developer -- as such I will close it.  Thanks for asking that question with candour.  I agree, I think that whether this detail is the intent of the problem or not, this is an interesting lesson for aspiring developers to follow.

PR #1547, opened 2025-02-12
Prefer more concise and cross-language ** operator over Integer.pow/2

codingthat:
`**` has been available since 1.13, and will be already familiar to those coming from other languages.


angelikatyborska:
Sure, why not! We don't support solving Exercism exercises in Elixir 1.12 or lower anyway.

angelikatyborska (approved):

Issue #1552, opened 2025-03-03
prime-factors timeout hint is wrong


angelikatyborska:
This comment tells you you can just uncomment the timeout tag:
```
  # @tag timeout: 2000
  #
  # The timeout tag above will set the below test to fail unless it completes
  # in under two seconds. Uncomment it if you want to test the efficiency of your
  # solution.
```

That is wrong. All tests run in trace mode which sets timeouts to infinity. You either need to remove `trace: true` or run with flag `--no-trace` for the timeout to work.

PR #1555, opened 2025-03-09
Fix prime-factors timeout hint is wrong 

angelikatyborska:
Resolves https://github.com/exercism/elixir/issues/1552


jiegillet (approved):
Nice catch.
I doubled checked that the timeout doesn't work without `--no-trace` and that this is the only place we use `@tag timeout`

PR #1560, opened 2025-03-15
Sync flatten-array

angelikatyborska:
The updated test cases only differ by one detail - the test description uses the word "array" instead of "list". I think is is an undesirable change for the Elixir track. I am ignoring those changes, and I have added an appends file to clarify that there are no arrays in Elixir.


neenjaw (approved):
Agree, the update to array is an unfortunate choice of language. It's Gmail bad a more agnostic word wasn't selected like "collection"

PR #1561, opened 2025-03-27
Fix binary shortcut notation name and link in hints

codingthat:
The anchor from the old link no longer exists, and I added the "shortcut notation" phrase to match how the subsection under the new link refers to it.


neenjaw (approved):

jiegillet:
Thanks a lot :)

PR #1565, opened 2025-04-18
Use 22.04 instead of deprecated 20.04

codingthat:
Cf. https://github.com/actions/runner-images/issues/11101 — currently failing normal PR checks.


angelikatyborska (approved):
Thanks!

PR #1567, opened 2025-04-25
Remove Unicode test from `series` exercise

codingthat:
To be consistent with its instructions.md ("Given a string of digits...") and design.md ("This is one of the simpler `Enum` exercises in the track and we want to keep it simple.")


angelikatyborska (approved):
Thank you!

PR #1568, opened 2025-05-06
Fix for high_score_test.exs

AntonRich:
There's a bug in the test at line 140.
|> HighScore.update_score("José Valim", 486_373)
It asserts against a different number 486_374

### DIFF
--- a/exercises/concept/high-score/test/high_score_test.exs
+++ b/exercises/concept/high-score/test/high_score_test.exs
@@ -137,7 +137,7 @@ defmodule HighScoreTest do
         HighScore.new()
         |> HighScore.add_player("José Valim")
         |> HighScore.update_score("José Valim", 1)
-        |> HighScore.update_score("José Valim", 486_373)
+        |> HighScore.update_score("José Valim", 486_374)
 
       assert scores == %{"José Valim" => 486_374}
     end


exercises/concept/high-score/test/high_score_test.exs
```diff
@@ -137,7 +137,7 @@ defmodule HighScoreTest do
         HighScore.new()
         |> HighScore.add_player("José Valim")
         |> HighScore.update_score("José Valim", 1)
-        |> HighScore.update_score("José Valim", 486_373)
+        |> HighScore.update_score("José Valim", 486_374)
```
neenjaw, line 140: I disagree with your suggestion here, it just isn't apparent since no one had approved the work flows yet.

The update_score/2 function should add the value to the previous value. See the example solution here:

https://github.com/exercism/elixir/blob/b8fc879e1b625f00a75c9978582162e10fdea6ea/exercises/concept/high-score/.meta/exemplar.ex#L18-L20

neenjaw (changes requested):
Suggestion invalidates existing example solution

exercises/concept/high-score/test/high_score_test.exs
neenjaw: I appreciate your motivation to improve the track as things come up, but it's good to discuss these things as potential issues before leaping to a patch.  This is a pretty trivial change, so you probably didn't lose a lot of time, but we have seen it be problematic with submissions where large changes/refactors are performed without consultation.

PR #1569, opened 2025-05-10
Add practice exercise `split-second-stopwatch`

jiegillet:
Brand new exercise!

I have been using Cursor for a little while, and I must say that this really helps with implementing tests like that. You make one change in the test files, tab-tab-tab and all the next ones change as well. You gotta be careful, but it really saves time. It also pretty much implemented the solution without me really asking for it. Pretty interesting experience.

### DIFF
--- a/config.json
+++ b/config.json
@@ -2288,7 +2288,7 @@
       },
       {
         "slug": "split-second-stopwatch",
-        "name": "Split Second Stopwatch",
+        "name": "Split-Second Stopwatch",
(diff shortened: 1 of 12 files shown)


jiegillet:
Oh no! I forgot that `Duration` was only introduced in Elixir 1.17, we need to support from 1.13.
I'll think of some other way to implement it, maybe using `Time`, but not today, so I'll set this back to draft.

jiegillet:
Done. Using `Time` actually made the code cleaner.

neenjaw (approved):
LGTM

PR #1575, opened 2025-06-07
Add new practice exercise `baffling-birthdays`

jiegillet:
New exercise.

This one is a bit of a special one, and I might have gotten carried away with the math :D

It involves randomness, and therefore the tests have the potential to be flaky. I did the math to ensure that the flakiness is kept to a minimum (like one fail per 100k runs), but the safer I make the tests, the less strict they become.

### DIFF
--- a/config.json
+++ b/config.json
@@ -2339,6 +2339,21 @@
         "difficulty": 5
       },
       {
+        "slug": "baffling-birthdays",
+        "name": "Baffling Birthdays",
+        "uuid": "e5ebca45-1e54-45b6-89bb-0cddfd23b6e1",
+        "practices": [
+          "randomness"
+        ],
+        "prerequisites": [
+          "randomness",
+          "dates-and-time",
+          "lists",
+          "enum"
+        ],
+        "difficulty": 6
+      },
+      {
         "slug": "game-of-life",
         "name": "Game of Life",
         "uuid": "32d53ab8-abea-4483-a8cb-ee3b908a9dd7",
--- a/config.json
+++ b/config.json
@@ -2349,7 +2349,10 @@
           "randomness",
           "dates-and-time",
           "lists",
-          "enum"
+          "enum",
+          "ranges",
+          "list-comprehensions",
+          "tuples"
         ],
         "difficulty": 6
       },
--- a/exercises/practice/baffling-birthdays/.meta/example.ex
+++ b/exercises/practice/baffling-birthdays/.meta/example.ex
@@ -1,0 +1,55 @@
+defmodule BafflingBirthdays do
+  @moduledoc """
+  Estimate the probability of shared birthdays in a group of people.
+  """
+
+  @spec shared_birthday?(birthdates :: [Date.t()]) :: boolean()
+  def shared_birthday?(birthdates) do
+    result =
+      Enum.reduce_while(birthdates, MapSet.new(), fn date, birthdays ->
+        birthday = {date.month, date.day}
+
+        if MapSet.member?(birthdays, birthday) do
+          {:halt, :found_a_shared_birthday}
+        else
+          {:cont, MapSet.put(birthdays, birthday)}
+        end
+      end)
+
+    result == :found_a_shared_birthday
+  end
+
+  @spec random_birthdates(group_size :: integer()) :: [Date.t()]
+  def random_birthdates(group_size) do
+    for _ <- 1..group_size do
+      year = generate_non_leap_year_january_first(0, 3000)
+      days_to_add = Enum.random(0..364)
+      Date.add(year, days_to_add)
+    end
+  end
+
+  defp generate_non_leap_year_january_first(min, max) do
+    year = Enum.random(min..max)
+    january_first = Date.new!(year, 1, 1)
+
+    if Date.leap_year?(january_first) do
+      Date.add(january_first, 366)
+    else
+      january_first
+    end
+  end
+
+  @spec estimated_probability_of_shared_birthday(group_size :: integer()) :: float()
+  def estimated_probability_of_shared_birthday(group_size) do
+    sample_size = 1000
+
+    shared_birthdays_count =
+      for _ <- 1..sample_size,
+          group_size |> random_birthdates() |> shared_birthday?(),
+          reduce: 0 do
+        count -> count + 1
+      end
+
+    100 * shared_birthdays_count / sample_size
+  end
+end
--- a/exercises/practice/baffling-birthdays/test/baffling_birthdays_test.exs
+++ b/exercises/practice/baffling-birthdays/test/baffling_birthdays_test.exs
@@ -1,0 +1,203 @@
+defmodule BafflingBirthdaysTest do
+  use ExUnit.Case
+
+  describe "shared birthday" do
+    # @tag :pending
+    test "one birthdate" do
+      birthdates = [~D[2000-01-01]]
+      assert BafflingBirthdays.shared_birthday?(birthdates) == false
+    end
+
+    @tag :pending
+    test "two birthdates with same year, month, and day" do
+      birthdates = [~D[2000-01-01], ~D[2000-01-01]]
+      assert BafflingBirthdays.shared_birthday?(birthdates) == true
+    end
+
+    @tag :pending
+    test "two birthdates with same year and month, but different day" do
+      birthdates = [~D[2012-05-09], ~D[2012-05-17]]
+      assert BafflingBirthdays.shared_birthday?(birthdates) == false
+    end
+
+    @tag :pending
+    test "two birthdates with same month and day, but different year" do
+      birthdates = [~D[1999-10-23], ~D[1988-10-23]]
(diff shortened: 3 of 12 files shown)


exercises/practice/baffling-birthdays/test/baffling_birthdays_test.exs
```diff
@@ -0,0 +1,203 @@
...
+        day when day <= 30 -> :math.sqrt(group_size * 11 / 365 * (1 - 11 / 365))
+        day when day == 31 -> :math.sqrt(group_size * 7 / 365 * (1 - 7 / 365))
+      end
+
+      counts_outside_95_percent_confidence_interval =
+        day_frequencies
+        |> Enum.filter(fn {day, count} ->
+          abs(count - expected_count.(day)) > 1.96 * expected_count_standard_deviation.(day)
+        end)
+        |> length()
```
angelikatyborska, line 141: Is there maybe a website that you could link in a comment in this code that could explain what's going on? My statistics knowledge is very rusty

exercises/practice/baffling-birthdays/.meta/example.ex
```diff
@@ -0,0 +1,55 @@
...
+        end
+      end)
+
+    result == :found_a_shared_birthday
+  end
+
+  @spec random_birthdates(group_size :: integer()) :: [Date.t()]
+  def random_birthdates(group_size) do
+    for _ <- 1..group_size do
+      year = generate_non_leap_year_january_first(0, 3000)
```
angelikatyborska, line 25: This is an interesting year range 😬 just so that we're clear: it's completely unnecessary to make the year random, right? To pass the tests, you might just as well choose the same non-leap year for all birthdates.

config.json
```diff
@@ -2338,6 +2338,21 @@
...
+        "name": "Baffling Birthdays",
+        "uuid": "e5ebca45-1e54-45b6-89bb-0cddfd23b6e1",
+        "practices": [
+          "randomness"
+        ],
+        "prerequisites": [
+          "randomness",
+          "dates-and-time",
+          "lists",
+          "enum"
```
angelikatyborska, line 2352: You have also used: list-comprehensions, ranges, tuples. I think list-comprehensions might be optional (you're not using cartesian products, right?), but ranges might be necessary for generating a list of a given length.

Additionally a MapSet, but we don't have a concept for that 🤷

exercises/practice/baffling-birthdays/.meta/example.ex
jiegillet: Yeah, as far as the tests are concerned you could pick `year = 2025` and be done with it, that's true.
In terms of generating random `Date.t()`, this felt more natural to me. I guess it doesn't really matter?

exercises/practice/baffling-birthdays/test/baffling_birthdays_test.exs
```diff
@@ -0,0 +1,203 @@
...
+    end
+
+    @tag :pending
+    test "among seventy people" do
+      group_size = 70
+      expected = 99.915958
+
+      # for a sample size of 100 and this delta, the assertion is expected to fail once in 100 runs
+      # for a sample size of 600 and this delta, the assertion is expected to fail once in a billion runs
+      delta = 0.83
```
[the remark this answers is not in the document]
jiegillet: I used [this wolfram alpha app](https://www.wolframalpha.com/input?i=binomial+distribution+confidence+interval&assumption=%22FSelect%22+-%3E+%7B%7B%22BinomialInterval%22%7D%2C+%22dflt%22%7D&assumption=%7B%22F%22%2C+%22BinomialInterval%22%2C+%22phat%22%7D+-%3E%220.99915958%22&assumption=%7B%22F%22%2C+%22BinomialInterval%22%2C+%22c%22%7D+-%3E%220.99%22&assumption=%7B%22F%22%2C+%22BinomialInterval%22%2C+%22n%22%7D+-%3E%22100%22)

- enter the sample proportion (`expected`)
- enter the confidence level (0.99, which means one failure in 100)
- enter the sample size (100)

That gives you the 99% interval, meaning a random pick will fall in there 99% of the time. This interval goes from 0.03417 to 0.1997, which is 2 times the delta, that's how I got the value.

After that, I entered a confidence level of 0.999999999 (one failure in a billion) and tweaked the sample size until I found an interval that's fairly close to the confidence level of 0.99, that came out to being around 600 every time. It's not an exact calculation, but it should be close enough.

I used `mix test --repeat-until-failure 100` with a sample of 100 to check if the solution would actually fail, and it does about half the time, which is expected, so it seems to work. I did not try a billion times with a sample of 600, but I ran it for a bit and never managed to make it fail 😆 


I have one dark secret for this particular test though. If you try with a sample of 100, you will almost always fail the test, because with 100 sample, most of the time you will estimate 100 or 99, which is outside of the theoretical range (99.08 to 1). But I chose to ignore it, it's complicated to explain and it's unlikely that people will only pick a 100 sample size, and if they do, they'll see the comment and pick something bigger.

config.json
jiegillet: Thank you, I added them all, it could help to know the concepts.

> Additionally a MapSet, but we don't have a concept for that 🤷

I made a [set concept for Elm](https://exercism.org/tracks/elm/concepts/set) a while ago. Want me to fork it? :)

jiegillet:
> The tests with your implementation are very fast, but I think some students might go too far with sample sizes and have problems with timeouts 🤔

I did some test by changing my sample size:
- 1M: 40s
- 100k: 4s
- 10k: 0.4s

The cutoff is 10 seconds right? That's a limit of 250k samples. That doesn't seem too bad.

config.json
angelikatyborska: Sure thing, if you have the time, it would be very nice to have that concept in Elixir too

exercises/practice/baffling-birthdays/test/baffling_birthdays_test.exs
angelikatyborska: Hmm... yes, that is way too much. I won't even pretend I understand 🙈 

I see a few potential problems with this testing code:
- students might get confused by it (but in reality they should just read the test description and know what to do)
- confused students might ask mentors about this test code, and mentors won't know how to explain it (but so what? it's important to know how to say "I don't know" 😬)
- if this test code ever needs to be changed and there's nobody with a physics or maths phd around... I won't know how to fix the tests 😅 (but then I would just throw it all away and do what Erik did: https://github.com/exercism/csharp/pull/2402/files#diff-a31c6c904e805077bb6ab5a333d6e0f154d63488697388f75ce77e4ea40a4eba)

Still, I'm pretty impressed with the effort you went through and if I understood it, I'm sure I would agree it's the only reasonable way to write a reliable test for random values 🤓 I'm fine just accepting that I don't get it and let it be merged.

exercises/practice/baffling-birthdays/test/baffling_birthdays_test.exs
```diff
@@ -0,0 +1,203 @@
...
+    test "multiple birthdates with more than one shared birthday" do
+      birthdates = [
+        ~D[1966-07-29],
+        ~D[1977-02-12],
+        ~D[2001-12-25],
+        ~D[1980-07-29],
+        ~D[2019-02-12]
+      ]
+
+      output = BafflingBirthdays.shared_birthday?(birthdates) == true
```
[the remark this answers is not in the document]
angelikatyborska: @jiegillet sorry, I should have been clearer - please make those expressions into assertions 😅 now it's:

```
    warning: use of operator == has no effect
    │
 63 │       BafflingBirthdays.shared_birthday?(birthdates) == true
    │                                                      ~
    │
    └─ test/baffling_birthdays_test.exs:63:54
```

angelikatyborska:
> The cutoff is 10 seconds right? 

I somehow remember it's 2x or 3x the average test runtime set in our config file (`"average_run_time": 4`) but maybe it's just simple 10s. Anyway similar values (8s, 10s, or 12s)

> especially considering that the tests are kind of hinting that a sample of 600 would be fine.

Agreed, I'm no longer worried about test speed. Thanks!

exercises/practice/baffling-birthdays/test/baffling_birthdays_test.exs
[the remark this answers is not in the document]
jiegillet: Oh, don't apologize for me being an idiot 😆

exercises/practice/baffling-birthdays/test/baffling_birthdays_test.exs
jiegillet: I would not have done this if it wasn't fun :)

I actually agree with your list of problems, I share your concerns. I looked at Erik's test, my tests include his (`assert map_size(month_frequencies) == number_of_months`), but yes, I am fine with future maintainers simplifying the tests if I get hit by a bus. Or if this causes more confusion than it's worth.

angelikatyborska (approved):

PR #1589, opened 2025-08-29
Try fixing empty.txt not downloading via CLI

angelikatyborska:
Reported on the forum https://forum.exercism.org/t/newsletter-exercise-assets-empty-txt-file-not-being-created/19023


angelikatyborska:
Maybe we could wait and see if they change the CL instead

jiegillet:
> Maybe we could wait and see if they change the CL instead

That would seem like the better fix. Is there an issue opened for this?

angelikatyborska:
Actually, a PR was merged already: https://github.com/exercism/cli/pull/1213

PR #1593, opened 2025-09-10
Add new practice exercise: Camicia

FraSanga:
This pull request introduces a new Elixir practice exercise called `Camicia`.

* [Link to the proposal on the forum](https://forum.exercism.org/t/new-exercise-proposal-camicia/19003)
* [Link to the Elixir implementation on the forum](https://forum.exercism.org/t/new-practice-exercise-camicia/19181)

TO-DO:

- [x] Create example.ex
- [x] Add practices & prerequisites in config.json
- [x] Edit difficulty (6?)

### DIFF
--- a/exercises/practice/camicia/test/camicia_test.exs
+++ b/exercises/practice/camicia/test/camicia_test.exs
@@ -3,79 +3,79 @@ defmodule CamiciaTest do
 
   # @tag :pending
   test "two cards, one trick" do
-    playerA = ["2"]
-    playerB = ["3"]
+    player_a = ["2"]
+    player_b = ["3"]
     expected = {:finished, 2, 1}
-    assert Camicia.simulate(playerA, playerB) == expected
+    assert Camicia.simulate(player_a, player_b) == expected
   end
 
   @tag :pending
   test "three cards, one trick" do
-    playerA = ["2", "4"]
-    playerB = ["3"]
+    player_a = ["2", "4"]
+    player_b = ["3"]
     expected = {:finished, 3, 1}
-    assert Camicia.simulate(playerA, playerB) == expected
+    assert Camicia.simulate(player_a, player_b) == expected
   end
 
   @tag :pending
   test "four cards, one trick" do
-    playerA = ["2", "4"]
-    playerB = ["3", "5", "6"]
+    player_a = ["2", "4"]
+    player_b = ["3", "5", "6"]
     expected = {:finished, 4, 1}
-    assert Camicia.simulate(playerA, playerB) == expected
+    assert Camicia.simulate(player_a, player_b) == expected
   end
 
   @tag :pending
   test "the ace reigns supreme" do
-    playerA = ["2", "A"]
-    playerB = ["3", "4", "5", "6", "7"]
+    player_a = ["2", "A"]
+    player_b = ["3", "4", "5", "6", "7"]
     expected = {:finished, 7, 1}
-    assert Camicia.simulate(playerA, playerB) == expected
+    assert Camicia.simulate(player_a, player_b) == expected
   end
 
   @tag :pending
   test "the king beats ace" do
-    playerA = ["2", "A"]
-    playerB = ["3", "4", "5", "6", "K"]
+    player_a = ["2", "A"]
+    player_b = ["3", "4", "5", "6", "K"]
     expected = {:finished, 7, 1}
-    assert Camicia.simulate(playerA, playerB) == expected
+    assert Camicia.simulate(player_a, player_b) == expected
   end
 
   @tag :pending
   test "the queen seduces the king" do
-    playerA = ["2", "A", "7", "8", "Q"]
-    playerB = ["3", "4", "5", "6", "K"]
+    player_a = ["2", "A", "7", "8", "Q"]
+    player_b = ["3", "4", "5", "6", "K"]
     expected = {:finished, 10, 1}
-    assert Camicia.simulate(playerA, playerB) == expected
+    assert Camicia.simulate(player_a, player_b) == expected
   end
 
   @tag :pending
   test "the jack betrays the queen" do
-    playerA = ["2", "A", "7", "8", "Q"]
-    playerB = ["3", "4", "5", "6", "K", "9", "J"]
+    player_a = ["2", "A", "7", "8", "Q"]
+    player_b = ["3", "4", "5", "6", "K", "9", "J"]
     expected = {:finished, 12, 1}
-    assert Camicia.simulate(playerA, playerB) == expected
+    assert Camicia.simulate(player_a, player_b) == expected
   end
 
   @tag :pending
   test "the 10 just wants to put on a show" do
-    playerA = ["2", "A", "7", "8", "Q", "10"]
-    playerB = ["3", "4", "5", "6", "K", "9", "J"]
+    player_a = ["2", "A", "7", "8", "Q", "10"]
+    player_b = ["3", "4", "5", "6", "K", "9", "J"]
     expected = {:finished, 13, 1}
-    assert Camicia.simulate(playerA, playerB) == expected
+    assert Camicia.simulate(player_a, player_b) == expected
   end
 
   @tag :pending
   test "simple loop with decks of 3 cards" do
-    playerA = ["J", "2", "3"]
-    playerB = ["4", "J", "5"]
+    player_a = ["J", "2", "3"]
+    player_b = ["4", "J", "5"]
     expected = {:loop, 8, 3}
-    assert Camicia.simulate(playerA, playerB) == expected
+    assert Camicia.simulate(player_a, player_b) == expected
   end
 
   @tag :pending
   test "the story is starting to get a bit complicated" do
-    playerA = [
+    player_a = [
       "2",
       "6",
       "6",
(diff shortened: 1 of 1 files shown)


neenjaw:
Thanks for starting a PR to add a new exercise.  I have a few questions to know how to evaluate/guide this PR:
- Is there an [exercism forum discussion](https://forum.exercism.org/) that this exercise comes from?
- Is this implemententing an exercise described in [problem specifications](https://github.com/exercism/problem-specifications) that was previously unimplemented for Elixir?
- Does this exercise practice concepts/features not covered in other practice exercises?

Some exercises have a `design.md` file to describe the intention for the exercise ([ex](https://github.com/exercism/elixir/blob/main/exercises/concept/lasagna/.meta/design.md)), perhaps that would be a good addition to this PR or maybe you could add some more detail to this PR description.

Can you give me some more information to know how to help/guide you through this process? Thanks! 💙

FraSanga:
I have created the first draft of the solution.
If anyone would like to take a look, it's there waiting for you.

I also ran test_exercises.sh and everything went smoothly.

> Some exercises have a design.md file to describe the intention for the exercise ([ex](https://github.com/exercism/elixir/blob/main/exercises/concept/lasagna/.meta/design.md))

From what I understand, this file serves to summarise the purpose of the exercise in Elixir. 
I am not an expert in this language, so I would not be able to write something that is correct from a purely theoretical point of view.

My aim, as I wrote in the proposal, is to bring as many users as possible into the world of research; even just sparking someone's curiosity would be good.

I think that until you confirm the solution, there's not much point in writing the rest.

angelikatyborska:
> Some exercises have a design.md file to describe the intention for the exercise ([ex](https://github.com/exercism/elixir/blob/main/exercises/concept/lasagna/.meta/design.md)), perhaps that would be a good addition to this PR or maybe you could add some more detail to this PR description.

IMO `design.md` files are only necessary for concept exercises. Practice exercises don't need to have a purpose other than "write more Elixir code" 🙂 

> I think that until you confirm the solution, there's not much point in writing the rest.

TBH, I never care that much about the specific solution to a new practice exercise. It only exists to prove that an exercise can be solved at all, but it's not shown to students and it doesn't have to be perfect.

The test suite and deciding which concepts must be known before attempting to solve the exercise (prerequisites) are much more important.

I'll take a look at this PR's changes now.

FraSanga:
> The test suite and deciding which concepts must be known before attempting to solve the exercise (prerequisites) are much more important.

I agree. I think the prerequisites are very important. While completing the Elixir track, I would occasionally peek at the graph to see which dots were coloured in after downloading the exercise locally.

> I'll take a look at this PR's changes now.

If you have any questions or requests, I am available immediately for the next two hours or so.

exercises/practice/camicia/test/camicia_test.exs
```diff
@@ -0,0 +1,1432 @@
...
+  @tag :pending
+  test "the king beats ace" do
+    playerA = ["2", "A"]
+    playerB = ["3", "4", "5", "6", "K"]
+    expected = {:finished, 7, 1}
+    assert Camicia.simulate(playerA, playerB) == expected
+  end
+
+  @tag :pending
+  test "jack beats king" do
```
angelikatyborska, line 45: For easier maintenance, the test names should be identical to those in problem specs (https://github.com/exercism/problem-specifications/blob/main/exercises/camicia/canonical-data.json#L74-L87)

```suggestion
  test "the queen seduces the king" do
```

exercises/practice/camicia/test/camicia_test.exs
```diff
@@ -0,0 +1,1432 @@
...
+      "4",
+      "6"
+    ]
+
+    expected = {:finished, 327, 42}
+    assert Camicia.simulate(playerA, playerB) == expected
+  end
+
+  @tag :pending
+  test "Kebler 1999" do
```
angelikatyborska, line 858: This is the test name in problem specs:
```suggestion
  test "Kleber 1999" do
```

exercises/practice/camicia/lib/camicia.ex
```diff
@@ -0,0 +1,23 @@
...
+    ## Examples
+
+      iex> Camicia.simulate(["2"], ["3"])
+      {:finished, 2, 1}
+
+      iex> Camicia.simulate(["J", "2", "3"], ["4", "J", "5"])
+      {:loop, 8, 3}
+  """
+
+  @spec simulate(list, list) :: {:finished | :loop, non_neg_integer(), non_neg_integer()}
```
angelikatyborska, line 20: We could be a bit more specific:
```suggestion
  @spec simulate(list(String.t()), list(String.t()) :: {:finished | :loop, non_neg_integer(), non_neg_integer()}
```

exercises/practice/camicia/.meta/example.ex
```diff
@@ -0,0 +1,140 @@
...
+  end
+
+  defp round(player, opponent, cards, tricks, current_player),
+    do: do_round(player, opponent, cards, tricks, :number, [], current_player)
+
+  defp do_round([], opponent, cards, tricks, _type, central_pile, current_player) do
+    {opponent ++ Enum.reverse(central_pile), [], cards, tricks + 1, change_player(current_player)}
+  end
+
+  defp do_round([card | rest], opponent, cards, tricks, :number, central_pile, current_player) do
```
angelikatyborska, line 60: As I wrote already, we don't really care that much about how an example solution looks like :) but I feel like I should warn you just for you that writing functions that accept more than 2-3 arguments will get you in trouble, in any language. It is not really possible to keep in your working memory that many arguments at once - which one is at which position and what they meant. Thus such code is pretty hard to read with understanding.

In Elixir, I would recommend replacing a long list of arguments with a single map or struct argument, for the benefit of having named keys.

exercises/practice/camicia/.meta/example.ex
```diff
@@ -0,0 +1,140 @@
...
+          rest,
+          opponent,
+          cards + 1,
+          tricks,
+          {:payment, count - 1},
+          [card | central_pile],
+          current_player
+        )
+
+      payment_card when payment_card in ["J", "Q", "K", "A"] ->
```
angelikatyborska, line 114: You repeat the definition of payment cards three times in your solution - maybe it's worth putting them in a module attribute?

exercises/practice/camicia/.meta/example.ex
```diff
@@ -0,0 +1,140 @@
...
+        count = Map.get(@counts, payment_card)
```
angelikatyborska, line 115: I'm not sure if "count" and "counts" is a good name here. It doesn't really describe what's happening at all - "making a payment". Maybe some more specific phrase like "price" or "payment amount" or "card to pay"
or "payment card value" would make the code clearer? I'm not very familiar with this game so I don't know the proper terminology.

config.json
```diff
@@ -2371,6 +2371,28 @@
...
       },
+      {
+        "slug": "camicia",
+        "name": "Camicia",
+        "uuid": "61976949-e1a6-4494-9ca0-d6d771e2e8e8",
+        "practices": [
+          "lists",
+          "mapsets",
+          "tail-call-recursion"
+        ],
```
angelikatyborska, line 2382: That's too many things that this exercise claims to be good at practicing. I still have to think about it, but I would either go with recursion, tail-call-recursion, mapsets, or nothing.

It's too complex to practice something as simple as lists.

For practicing recursion or tail-call-recursion, there's always the problem of choosing one of those because often either of them works.

If an exercise can be solved in too many different ways (which complex exercises often do), it's better to not claim that it practices anything, to avoid confusing people that will naturally come up with a solution that doesn't use a specific concept that we thought it would practice.

angelikatyborska (commented):
I'm leaving a partial review because I ran out of time for right now. I didn't fully understand the solution yet and I didn't review "prerequisites".

exercises/practice/camicia/.meta/example.ex
FraSanga: I wrote the solution on the spot.
I had a working Python script and tried to convert it to Elixir.
I'll fix it now.

config.json
FraSanga: All right, as you wish.

exercises/practice/camicia/lib/camicia.ex
```diff
@@ -0,0 +1,24 @@
...
+      iex> Camicia.simulate(["2"], ["3"])
+      {:finished, 2, 1}
+
+      iex> Camicia.simulate(["J", "2", "3"], ["4", "J", "5"])
+      {:loop, 8, 3}
+  """
+
+  @spec simulate(list(String.t()), list(String.t())) ::
+          {:finished | :loop, non_neg_integer(), non_neg_integer()}
+  def simulate(playerA, playerB) do
```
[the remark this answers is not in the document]
FraSanga: Right.

FraSanga:
🥳 
I swear I didn't forget to change the variables in the tests on purpose 😅🤣

angelikatyborska:
Congrats on your first Elixir exercise 🥳 it's live now, so you can go ahead and publish your own solution if you want to be the first one 😁

FraSanga:
There was no need to even say it, I did it as soon as I saw the notification.  
Thank you again.

PR #1596, opened 2025-09-15
Sync two-bucket tests

FraSanga:
As reported in the recurring task in this repo, I updated the tests and checked their implementation in Elixir.

In this case, two new tests were added and also implemented in Elixir.
The tests pass without changing anything.


angelikatyborska (approved):
Perfect! Thank you very much

PR #1599, opened 2025-09-17
Fix broken links

angelikatyborska:
The link https://www.richardpmann.com/beggar-my-neighbour-records.html is still live but keeps failing the lychee check 🤷


neenjaw:
> Damn, there's way too many 403 and 429s on CI. I wasn't getting them locally (using act). I'm starting to doubt whether there's any point in this CI check 😞

It's been failing for a long time and when links are actually broken, students have been quick to help.  I think is okay to just get rid of it.

PR #1620, opened 2026-03-08
Add practice exercise `line-up`

FraSanga:
https://forum.exercism.org/t/adding-exercises-to-elixir-track/33067/2

### DIFF
--- a/config.json
+++ b/config.json
@@ -877,7 +877,7 @@
       },
       {
         "slug": "line-up",
-        "name": "Line-up",
+        "name": "Line Up",
         "uuid": "f633a810-0323-496b-a728-e05f560fe393",
         "practices": [
           "cond"
@@ -885,7 +885,8 @@
         "prerequisites": [
           "integers",
           "strings",
-          "cond"
+          "cond",
+          "lists"
         ],
         "difficulty": 2
       },
--- a/config.json
+++ b/config.json
@@ -876,6 +876,22 @@
         "difficulty": 2
       },
       {
+        "slug": "line-up",
+        "name": "Line-up",
+        "uuid": "f633a810-0323-496b-a728-e05f560fe393",
+        "practices": [
+          "integers",
+          "strings",
+          "cond"
+        ],
+        "prerequisites": [
+          "integers",
+          "strings",
+          "cond"
+        ],
+        "difficulty": 2
+      },
+      {
         "slug": "nth-prime",
         "name": "Nth Prime",
         "uuid": "6d0c0785-3514-44fc-9b5a-0758989d1c6a",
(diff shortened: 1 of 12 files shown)


exercises/practice/line-up/test/line_up_test.exs
```diff
@@ -0,0 +1,197 @@
+defmodule LineUpTest do
+  use ExUnit.Case
+
+  # Names are chosen without Unicode characters to keep the exercise simple.
```
jiegillet: You can remove those comments, they are meant for the problem spec maintainers and exercise creators, not the students.

exercises/practice/line-up/test/line_up_test.exs
```diff
@@ -0,0 +1,197 @@
...
+  end
+
+  @tag :pending
+  test "format exceptional ordinal numeral 123" do
+    name = "Yma"
+    number = 123
+    output = LineUp.format(name, number)
+    expected = "Yma, you are the 123rd customer we serve today. Thank you!"
+
+    assert output == expected
```
jiegillet: The generator splits up the input like this because they sometimes get pretty long.
I don't know how much work that is (a nifty regex? an AI request?) but given that the inputs are so short here, maybe we could one shot the tests
```elixir
assert LineUp.format("Yma", 123) == "Yma, you are the 123rd customer we serve today. Thank you!"
```
That remains readable. What do you think?

config.json
```diff
@@ -875,6 +875,20 @@
         ],
         "difficulty": 2
       },
+      {
+        "slug": "line-up",
+        "name": "Line-up",
```
jiegillet: I think the official name doesn't have a dash: https://github.com/exercism/problem-specifications/blob/main/exercises/line-up/metadata.toml

(BTW, this is a neat little contribution someone could make to the generator in a separate PR: read the name from that file instead of using "TODO")

config.json
```diff
@@ -875,6 +875,20 @@
...
+        "uuid": "f633a810-0323-496b-a728-e05f560fe393",
+        "practices": [
+          "cond"
+        ],
+        "prerequisites": [
+          "integers",
+          "strings",
+          "cond"
```
jiegillet: You are also using "lists" in your example, let's add that.

jiegillet (commented):
This looks great, nice job!

I left some small comments, let me know what you think.

config.json
FraSanga: > (BTW, this is a neat little contribution someone could make to the generator in a separate PR: read the name from that file instead of using "TODO")

Great idea. As soon as I have time, I'll do the PR for that too.
In the meantime, I'll create an issue so it stays marked and we can remember it. (#1621)

EDIT: precheck keeps failing, but I don't see any problems.

jiegillet: The precheck was failing because the test file wasn't formatted correctly after your changes. I pushed a commit to fix it.

jiegillet (approved):
Looks great, thank you very much :)

PR #1631, opened 2026-05-02
Sync connect

angelikatyborska:
I notices that we're using outdated symbols `:black` and `:white` for the player names that don't  show up in the exercise description, so I updated them to `:X` and `:O` to match the problem specs.

Adding two new tests.


jiegillet (approved):
Thank you :)

PR #1633, opened 2026-05-25
Format all `instructions.append.md` to start with an H1 (required but ignored) and H2

IsaacG:
See related discussions [here](https://forum.exercism.org/t/grep-instructions-append-may-need-a-visible-header/53923) as well as https://github.com/exercism/docs/pull/592. Append files should all have an H1. Including an H2 helps break the flow from the non-append file to the append file so it does not read as the same stream of thought.

### DIFF
--- a/exercises/practice/complex-numbers/.docs/instructions.append.md
+++ b/exercises/practice/complex-numbers/.docs/instructions.append.md
@@ -1,6 +1,6 @@
 # Instructions append
 
-## Implementation
+## Track specific instructions
 
 In this exercise, complex numbers are represented as a tuple-pair containing the real and imaginary parts.
 
--- a/exercises/practice/complex-numbers/.docs/instructions.append.md
+++ b/exercises/practice/complex-numbers/.docs/instructions.append.md
@@ -1,6 +1,6 @@
 # Instructions append
 
-## Track specific instructions
+## Elixir-specific changes
 
 In this exercise, complex numbers are represented as a tuple-pair containing the real and imaginary parts.
 
--- a/exercises/practice/complex-numbers/.docs/instructions.append.md
+++ b/exercises/practice/complex-numbers/.docs/instructions.append.md
@@ -1,5 +1,7 @@
 # Instructions append
 
+## Implementation
+
 In this exercise, complex numbers are represented as a tuple-pair containing the real and imaginary parts.
 
 For example, the real number `1` is `{1, 0}`, the imaginary number `i` is `{0, 1}` and the complex number `4+3i` is `{4, 3}`.
--- a/exercises/practice/custom-set/.docs/instructions.append.md
+++ b/exercises/practice/custom-set/.docs/instructions.append.md
@@ -1,5 +1,5 @@
 # Instructions append
 
-## Track specific instructions
+## Elixir-specific changes
 
 For this exercise please refrain from using the `MapSet` API.
--- a/exercises/practice/custom-set/.docs/instructions.append.md
+++ b/exercises/practice/custom-set/.docs/instructions.append.md
@@ -1,3 +1,5 @@
 # Instructions append
 
+## Implementation
+
 For this exercise please refrain from using the `MapSet` API.
--- a/exercises/practice/custom-set/.docs/instructions.append.md
+++ b/exercises/practice/custom-set/.docs/instructions.append.md
@@ -1,5 +1,5 @@
 # Instructions append
 
-## Implementation
+## Track specific instructions
 
 For this exercise please refrain from using the `MapSet` API.
--- a/exercises/practice/resistor-color/.docs/instructions.append.md
+++ b/exercises/practice/resistor-color/.docs/instructions.append.md
@@ -1,3 +1,5 @@
 # Instructions append
 
+## Implementation
+
 For this exercise, returning the list of colors is not required.
--- a/exercises/practice/resistor-color/.docs/instructions.append.md
+++ b/exercises/practice/resistor-color/.docs/instructions.append.md
@@ -1,5 +1,5 @@
 # Instructions append
 
-## Track specific instructions
+## Elixir-specific changes
 
 For this exercise, returning the list of colors is not required.
--- a/exercises/practice/resistor-color/.docs/instructions.append.md
+++ b/exercises/practice/resistor-color/.docs/instructions.append.md
@@ -1,5 +1,5 @@
 # Instructions append
 
-## Implementation
+## Track specific instructions
 
 For this exercise, returning the list of colors is not required.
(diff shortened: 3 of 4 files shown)


angelikatyborska:
I am assuming this is some sort of global platform change. Could you please provide a link to an explanation where this change is discussed and/or announced so I could familiarize myself? PRs with empty descriptions are confusing to maintainers that don't read whichever communications channel this was announced in.

IsaacG:
Apologies. See related discussions [here](https://forum.exercism.org/t/grep-instructions-append-may-need-a-visible-header/53923) as well as https://github.com/exercism/docs/pull/592

exercises/practice/complex-numbers/.docs/instructions.append.md
```diff
@@ -1,5 +1,7 @@
 # Instructions append
 
+## Implementation
```
angelikatyborska: The content that's before the append is this:
```
## Implementation Requirements

Given that you should not use built-in support for complex numbers, implement the following operations:

- **addition** of two complex numbers
- **subtraction** of two complex numbers
- **multiplication** of two complex numbers
- **division** of two complex numbers
- **conjugate** of a complex number
- **absolute value** of a complex number
- **exponentiation** of _e_ (the base of the natural logarithm) to a complex number
```

I think it will be confusing if there's a h2 "implementation requirements" and then just "implementation".

exercises/practice/custom-set/.docs/instructions.append.md
```diff
@@ -1,3 +1,5 @@
 # Instructions append
 
+## Implementation
+
```
angelikatyborska: The instructions for this exercise are:
```
# Instructions

Create a custom set type.

Sometimes it is necessary to define a custom data structure of some type, like a set.
In this exercise you will define your own set.
How it works internally doesn't matter, as long as it behaves like a set of unique elements.
```

exercises/practice/resistor-color/.docs/instructions.append.md
```diff
@@ -1,3 +1,5 @@
 # Instructions append
 
+## Implementation
+
```
angelikatyborska: The content before the append is:
```
The goal of this exercise is to create a way:

- to look up the numerical value associated with a particular color band
- to list the different band colors

Mnemonics map the colors to the numbers, that, when stored as an array, happen to map to their index in the array:
Better Be Right Or Your Great Big Values Go Wrong.

More information on the color encoding of resistors can be found in the [Electronic color code Wikipedia article][e-color-code].
```

exercises/practice/square-root/.docs/instructions.append.md
```diff
@@ -1,3 +1,5 @@
 # Instructions append
 
+## Implementation
+
```
angelikatyborska, line 4: The content before the append is:
```
Some potential approaches:

- Linear or binary search for a number that gives the input number when squared.
- Successive approximation using Newton's or Heron's method.
- Calculating one digit at a time or one bit at a time.

You can check out the Wikipedia pages on [integer square root][integer-square-root] and [methods of computing square roots][computing-square-roots] to help with choosing a method of calculation.
```

angelikatyborska (commented):
Thank you! If I understood correctly, we're free to choose whichever heading text we want, or even not add any h2s at all.

I reviewed the affected exercises and I'm not super happy with "Implementation" because it repeats a previous heading for complex-numbers.

@jiegillet @neenjaw It would be nice to get your opinion on this. I see 4 approaches:
- A) Accept this PR with its current downsides because "that's what everyone else is doing"
- B) Reject this PR and not add any h2s because "it worked so far"
- C) Choose a different h2 for all appends, but the same for each exercise. E.g. "Elixir-specific instructions" (might confuse people who aren't aware the same exercises are repeated in different languages)
- D) Choose different h2s for all appends, different for each exercise, to best fit the context. This approach will produce best results to the users right now, but it might fall apart as we pull automatic changes to the instructions and don't pay attention to the appends.

What do you think?

Thank you @IsaacG again for bringing this improvement idea to our doorstep 🙏

IsaacG:
> Thank you! If I understood correctly, we're free to choose whichever heading text we want, or even not add any h2s at all.

Absolutely.

If you're careful with the non-append MD files and are sure the text flows well from one to the other, the H2 isn't needed ... but the combined MD can get wonky if the generic MD gets updated.

If the append addresses a specific topic, a topic-specific H2 (D) is probably the "nicest" option. That requires a bit more work but shouldn't be too much work. If the H2 summaries the topic in the append file, it should be fairly change-proof vs changes to the generic files.

exercises/practice/custom-set/.docs/instructions.append.md
IsaacG: You could go with an H2 like "Restrictions" or "It's time to re-invent the wheel!"

exercises/practice/resistor-color/.docs/instructions.append.md
IsaacG: "Note" or "Ammendment" could work here.

exercises/practice/complex-numbers/.docs/instructions.append.md
IsaacG: "Data structure" or "Representation"?

jiegillet:
> * A) Accept this PR with its current downsides because "that's what everyone else is doing"

I feel you don't want me to pick this option 😆 

> * B) Reject this PR and not add any h2s because "it worked so far"

This seems like a valid option to me, looking at the instructions with appendix, it doesn't looks wrong. Maybe people who have already done the exercise on a different track might miss it though.
Note that it has the same downside as D: it might not work anymore if the instructions are updated automatically without us noticing.

> * C) Choose a different h2 for all appends, but the same for each exercise. E.g. "Elixir-specific instructions" (might confuse people who aren't aware the same exercises are repeated in different languages)

I think I'm leaning towards this. In fact I wish something like "Track specific instructions"  had been picked as the default for all tracks, this way students would be aware that those instructions may not propagate to other tracks, and they might get the habit of looking for that section when doing a familiar exercise on a new track.
 
> * D) Choose different h2s for all appends, different for each exercise, to best fit the context. This approach will produce best results to the users right now, but it might fall apart as we pull automatic changes to the instructions and don't pay attention to the appends.

That's also a valid option. Another downside is that it might be mistaken for the base instructions.

Overall, if I had to vote, I would pick C, with something that suggests that other tracks will differ there. If not, D or B would work too.

IsaacG:
```
~/exercism/elixir/exercises/practice » grep -h '^## ' */.docs/*append.md  | sort | uniq -c
      7 ## Elixir-specific changes
      7 ## Slow tests
```

angelikatyborska:
> Overall, if I had to vote, I would pick C,

Perfect, that is my choice, too! I see we already had some "Elixir-specific changes" headings in some appends, so I'm happy to propagate that heading to all other files that need it. Thanks again!

angelikatyborska (approved):

PR #1634, opened 2026-06-12
fix nth-prime typespec

Cohen-Carlisle:
I noticed that it uses non_neg_integer() but pos_integer() would be more appropriate (for both the count param and the returned prime number).

### DIFF
--- a/exercises/practice/nth-prime/.meta/example.ex
+++ b/exercises/practice/nth-prime/.meta/example.ex
@@ -2,7 +2,7 @@ defmodule Prime do
   @doc """
   Generates the nth prime.
   """
-  @spec nth(non_neg_integer) :: non_neg_integer
+  @spec nth(pos_integer()) :: pos_integer()
   def nth(count) when is_integer(count) and count > 0, do: do_nth(count, 2, [])
 
   defp do_nth(0, value, _), do: value
--- a/exercises/practice/nth-prime/lib/prime.ex
+++ b/exercises/practice/nth-prime/lib/prime.ex
@@ -2,7 +2,7 @@ defmodule Prime do
   @doc """
   Generates the nth prime.
   """
-  @spec nth(non_neg_integer) :: non_neg_integer
+  @spec nth(pos_integer()) :: pos_integer()
   def nth(count) do
   end
 end


jiegillet:
Hi!
I personally agree with you, pos_integer would be more appropriate
However there is a test that checks the output for the value 0, which means that 0 is a legitimate input.
However however, it just crashes on that value, so 0 is just as legitimate as -1 or `"seventeen"`, so maybe 0 should not be reflected in the type.
However however however, the [original test description](https://github.com/exercism/problem-specifications/blob/e2c23e29aa99997be4c95e48b549681b88f4b380/exercises/nth-prime/canonical-data.json#L41-L49) prescribes returning the error message `"there is no zeroth prime"`  which explicitly mentions 0, so maybe that should be a legitimate input, and the real fix for this would be to return an error tuple for 0, and keep crashing for -1 and `"seventeen"`.

What a mess 😆 
What do you think @Cohen-Carlisle?

Cohen-Carlisle:
Hmm, yeah that's an interesting point about the "there is no zeroth prime" error from the canonical data. The [guidance](https://github.com/exercism/problem-specifications/blob/e2c23e29aa99997be4c95e48b549681b88f4b380/README.md#conventions) says you don't necessarily need to check or generate an error message based on that string. So in this case I do think that changing to `pos_integer()` is appropriate for the input as well as the output. The other option I see is to use return values like `{:ok, prime}` and `{:error, reason}`, but it sees like overkill just to return a special error message for 0.

jiegillet:
Yeah, I think we are on the same page.

Then here is what I would propose to do: we keep `pos_integer()` because it's appropriate, but also we remove the "there is no zeroth prime" test altogether and add comment about it in `exercises/practice/nth-prime/.meta/tests.toml` explaining why we didn't implement it (like we already do in `exercises/practice/nucleotide-count/.meta/tests.toml`), something like "information on valid input is encoded in the type".

Cohen-Carlisle:
I think we could even keep that test, because it's simple to enforce in a guard (from the example solution):
```elixir
def nth(count) when is_integer(count) and count > 0, do: ...
```

Cohen-Carlisle:
@jiegillet what do you think about keeping the errors on 0 test?

jiegillet (approved):
I still don't think it makes sense to keep the 0 test, because if the type was a proper type declaration and not just a typespec, the compiler would issue a warning in the test, and I don't think it makes sense to have tests that trigger compiler warnings.

On the other hand, the compiler does _not_ trigger a warning as of 1.20 at least, so this a grey area and your suggestions is not objectively better than mine. Since you're the one doing the work, I'll approve and merge your contribution, thank you.

PR #1635, opened 2026-06-12
Update to Elixir 1.20

jiegillet:
In this PR:
- update to Elixir 1.20
- drop Elixir 1.15 (following https://elixir.hexdocs.pm/compatibility-and-deprecations.html)
- fix type and runtime warnings, tweak or remove some tests

I think this is ready, the test runner CI fails, because it needs this branch.
See my comment below for more thoughts.

### DIFF
--- a/exercises/practice/collatz-conjecture/test/collatz_conjecture_test.exs
+++ b/exercises/practice/collatz-conjecture/test/collatz_conjecture_test.exs
@@ -39,9 +39,4 @@ defmodule CollatzConjectureTest do
   test "negative value is an error " do
     assert_raise FunctionClauseError, fn -> CollatzConjecture.calc(-15) end
   end
-
-  @tag :pending
-  test "string as input value is an error " do
-    assert_raise FunctionClauseError, fn -> CollatzConjecture.calc("fubar") end
-  end
 end
--- a/exercises/practice/intergalactic-transmission/.meta/example.ex
+++ b/exercises/practice/intergalactic-transmission/.meta/example.ex
@@ -29,7 +29,7 @@ defmodule Transmission do
   @spec get_parity(bitstring()) :: <<_::1>>
   defp get_parity(data) do
     bit_count = bit_size(data)
-    <<value::size(bit_count)>> = data
+    <<value::size(^bit_count)>> = data
 
     parity_int =
       0..(bit_count - 1)
@@ -71,7 +71,7 @@ defmodule Transmission do
 
   defp finalize(data, acc) when bit_size(data) == 7 do
     bits_count = 8 - rem(bit_size(acc), 8)
-    <<head::size(bits_count), _::bitstring>> = data
+    <<head::size(^bits_count), _::bitstring>> = data
     <<acc::bitstring, head::size(bits_count)>>
   end
 end
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,8 +17,8 @@ jobs:
     runs-on: ubuntu-24.04
     strategy:
       matrix:
-        elixir: [1.19.2]
-        otp: [28.1]
+        elixir: [1.20.1]
+        otp: [29.0]
 
     steps:
       - name: Checkout code
@@ -76,8 +76,6 @@ jobs:
     strategy:
       matrix:
         include:
-          - elixir: '1.15.0'
-            otp: '26.0'
           - elixir: '1.16.0'
             otp: '26.2'
           - elixir: '1.17.0'
@@ -86,6 +84,8 @@ jobs:
             otp: '27.2'
           - elixir: '1.19.2'
             otp: '28.1'
+          - elixir: '1.20.1'
+            otp: '29.0'
 
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
@@ -121,4 +121,3 @@ jobs:
 
       - name: Run Checks
         run: bin/ci.sh
-
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 ## Setup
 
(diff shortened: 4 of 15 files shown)


jiegillet:
I did this PR in a couple of passes, and I think it's worth breaking down the categories of changes I made.

TLDR: there are classes of warnings that our CI doesn't catch, and we have some tests that awkwardly overlap with the compiler's job.

### Failing tests

This one is pretty basic, the compiler changed, so some tests started failing (`top-secret`).

### Examploid type warnings

Also basic, when compiling an exercise (or rather its exemploid solution) the compiler found some type warnings (`paint-by-number`, `intergalactic-transmission`).

### Runtime warnings

These are things like `Enum.slice([1, 2, 3, 4, 5], 3..2//-1)` that are not caught at compile time but log a warning at runtime (`kindergarten-garden`, `pascals-triangle`, `pig-latin`, `rail-fence-cipher`).

These are tricky, because they don't make the tests fail, even when using `mix test --warnings-as-errors`. I only noticed them from the logs from a [test-runner CI failure](https://github.com/exercism/elixir-test-runner/actions/runs/27396025480/job/80963383899?pr=143). They have probably been there for a long time and are not related to the 1.20 update.

I'm not sure what we need to do to catch these errors in the CI, but it would be worth doing, let's open an issue if you agree.

### Tests type warnings

These are type warnings in test files (`city-office`, `kitchen-calculator`, `beer-song`, `collatz-conjecture`, `list-ops`, `two-fer`). Some were obvious fixes, some were not, more on that in two paragraphs.

Those are tricky as well because they won't get caught by `MIX_ENV=test mix compile --warnings-as-errors` even though it really feels like they should. You can catch them with `mix test --warnings-as-errors` (you can add `--exclude test` if you don't want to run the tests).
Again, many of these predate 1.20, we just didn't notice them.

These are easy to catch, just add the flag, however I didn't do that because it would also break two other concept exercises `bread-and-potions` and `remote-control-car`. Those two explicitly pass in arguments of the wrong type to show that an error will be raised, but now of course the compiler complains about it. In  `collatz-conjecture` and `two-fer` I felt comfortable removing tests like that because they weren't in the canonical tests, but for concept exercises we need to have a broader conversation about this practice now that Elixir if officially a gradually typed language.

angelikatyborska:
Thanks for taking care of Elixir 1.20. It was on my radar too, but I'm glad I can just review a PR instead of doing all of this myself 😁 . I'll do the review in chunks, for now I'm leaving my thoughts about "Tests type warnings"

> Those are tricky as well because they won't get caught by `MIX_ENV=test mix compile --warnings-as-errors` even though it really feels like they should. 

I think it's because test files are `.exs` and those never get compiled when running `mix compile` which would produce artifacts (beam files), but they do get complied when running `mix test` without producing artifacts.

> These are easy to catch, just add the flag,

Yeah, agreed. Let's add it:  https://github.com/exercism/elixir/pull/1636

> it would also break two other concept exercises `bread-and-potions` and `remote-control-car`

This is an interesting problem. I looked at those two tests and asked myself: would a typical learner mess up their code in a way that those tests would catch?

For `bread-and-options`, the answer is not really:
```elixir
    test "cannot eat a completely new item" do
      assert_raise Protocol.UndefinedError, fn ->
        Edible.eat(%NewItem{}, %Character{})
      end
    end
```

To make this test fail, I think they would have needed to implement a fallback to any: https://elixir.hexdocs.pm/main/protocols.html#fallback-to-any

For `remote-control-car`, the answer is definitely yes:
```elixir
  test "display distance raises error when not given struct" do
    fake_car = %{
      battery_percentage: 100,
      distance_driven_in_meters: 0,
      nickname: "Fake"
    }

    assert_raise(FunctionClauseError, fn ->
      RemoteControlCar.display_distance(fake_car)
    end)
  end
```

All it takes is not using the struct name in the pattern in the function's head: `display_distance(car)` instead of `display_distance(%RemoteControlCar{})`. That "mistake" is already present in the boilerplate code because they're meant to figure out this detail themselves. It's even in the instructions:

```
Make sure the function only accepts a `RemoteControlCar` struct as the argument.
```

Would it make sense to move this check from a unit test to the analyzer? Or maybe we should find a mechanism to exclude this specific compilation warning from failing the CI?

I'll address the rest of the PR later.

exercises/practice/collatz-conjecture/test/collatz_conjecture_test.exs
```diff
@@ -39,9 +39,4 @@ defmodule CollatzConjectureTest do
   test "negative value is an error " do
     assert_raise FunctionClauseError, fn -> CollatzConjecture.calc(-15) end
   end
```
angelikatyborska, line 41: This one doesn't cause compilation warnings? Even though the typespec is `pos_integer`?

exercises/practice/intergalactic-transmission/.meta/example.ex
```diff
@@ -29,7 +29,7 @@ defmodule Transmission do
   @spec get_parity(bitstring()) :: <<_::1>>
   defp get_parity(data) do
     bit_count = bit_size(data)
-    <<value::size(bit_count)>> = data
+    <<value::size(^bit_count)>> = data
```
angelikatyborska, line 32: I'm very happy that 1.20 is finding those bugs 💯

angelikatyborska (approved):
All makes sense 💯 I'm lazy and I want to see green CI on the tooling PRs, so I'll let you update all the tooling PRs before reviewing.

LuuOW (commented):
Technical audit: code patterns and functional logic verified for system integrity.

exercises/practice/collatz-conjecture/test/collatz_conjecture_test.exs
jiegillet: The typespecs are not actually part of the type system, they are a separate system used with dialyzer.
One of the next steps in the type system development will be to provide type annotations that will probably replace the typespecs. That one will be a big PR :D

PR #1638, opened 2026-06-16
Fix misleading type_from_binary instructions in file-sniffer

frozar:
This PR fixes a misleading sentence in the File Sniffer concept exercise.

The `type_from_binary/1` function receives a binary file and determines the media type from the file signature. The previous wording mentioned the extension not matching, but extensions are handled by `type_from_extension/1` and `verify/2`.

This change clarifies that `type_from_binary/1` returns `nil` when the file signature does not match the expected ones.


angelikatyborska (approved):
Thank you, this makes sense!

PR #1643, opened 2026-07-05
Add test case for lose?(false , false)

### LINKED ISSUE #1642: Test Missing
looks like the test for Concept Pacman Rules task 3 is missing the refute Rules.lose?(false, false) case

Peacanduck:
added missing test case  refute Rules.lose?(false, false) to task 3 tests for Concept Pacman Rules

Resolves #1642

### DIFF
--- a/exercises/concept/pacman-rules/test/rules_test.exs
+++ b/exercises/concept/pacman-rules/test/rules_test.exs
@@ -55,6 +55,11 @@ defmodule RulesTest do
     test "don't lose if not touching a ghost" do
       refute Rules.lose?(true, false)
     end
+
+    @tag task_id: 3
+    test "don't lose if not touching a ghost without a power pellet active " do
+      refute Rules.lose?(false, false)
+    end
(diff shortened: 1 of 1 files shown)


angelikatyborska:
Hi! The reason we did not add a test for "not touching a ghost without a power pellet active" is that having a power pellet active is irrelevant for the losing condition if pacman is not touching any ghosts. A single test for "don't lose if not touching a ghost" should have been enough in our opinion.

Did you find yourself actually needing this test? Did you manage to write a solution that was incorrect, but passing all the tests?

Peacanduck:
yeah this passes all the tests for task 3 but will fail if you try and use the function in task 4 test 11 but everything else will pass
```
  def lose?(power_pellet_active?, touching_ghost?) do
      not power_pellet_active? and (not touching_ghost? or touching_ghost?) 
  end
```

angelikatyborska (approved):
Thank you

PR #1645, opened 2026-07-11
fix(german-sysadmin): refer to diacritics as non-ACSII letters

### LINKED ISSUE #1644: Mistake in german-sysadmin instructions regarding Latin characters
> [!WARNING]
> I'm not German nor do I know German. I just think I have good enough grasp of scripts in natural language to comment on this.

> You quickly realize that most of the company uses an old email client that doesn't recognize `müller@firma.de` as a valid email address because of the _non-Latin character_.
> 
> [Exercism: German Sysadmin Exercise](https://exercism.org/tracks/elixir/exercises/german-sysadmin) (emphasis mine)

Just a minor issue, `ü` is a Latin character, so describing that the issue is caused by non-Latin characters is inaccurate.

> Ü ([minuscule](https://en.wikipedia.org/wiki/Letter_case): ü) is a [Latin script](https://en.wikipedia.org/wiki/Latin_script) character composed of the character [U](https://en.wikipedia.org/wiki/U) and the [two dots](https://en.wikipedia.org/wiki/Two_dots_(diacritic)) diacritical mark.
> 
> \- [Wikipedia: Ü](https://en.wikipedia.org/wiki/%C3%9C)

Latin characters with diacritics are still Latin! To be a non-Latin character, it'd have to use an entirely different alphabet (script) like Cyrillic (e.g. Български) or Arabic (e.g. العربية).

I think a more correct way to describe it would be:

- … because of the _non-ASCII character_.
- … because of the _`ü` character_.

I think `non-ASCII` is the best fit because it's a more general description. A term like `diacritics` wouldn't be broad enough because `ß` does not contain diacritics but is also still a Latin character.

> ẞ… Latin [Capital Letter Sharp S](https://en.wikipedia.org/wiki/Capital_%E1%BA%9E)
> 
> \- [Wikipedia: List of Unicode characters](https://en.wikipedia.org/wiki/List_of_Unicode_characters) (The character literally has _Latin_ in the name!)

Should be a straight forward change (🤞) so I'm willing to make it! Just opening an issue first:

- to confirm if this is the right place to make the change, as I understand some exercises require updating a template maintained elsewhere, but I couldn't find one.
- if maintainers would like to discuss their preferred terminology first.

SethFalco:
See commit message(s).

---

I updated the test hints/instructions as they seemed to be explicitly wrong. Stating that diacritics or `ß` is non-Latin can be misleading.

However, I didn't update the tests as they don't outright say `ß` isn't Latin. It just asks for _"Latin letters"_ or the _"whole lowercase Latin alphabet"_ which I believe most people understand this refers to _"the 26"_ letters of the Latin alphabet.

It feels a bit contradictory, but I think this is the best route for UX/learners without getting overly pedantic about it.

> This modern form is the basis of the [Latin script](https://en.wikipedia.org/wiki/Latin_script), which is the most widely used writing system in the world, _often with [diacritics](https://en.wikipedia.org/wiki/Diacritic) or additional letters beyond the basic 26._
> 
> \- [Latin alphabet](https://en.wikipedia.org/wiki/Latin_alphabet) (emphasis mine)

## Related

* Closes https://github.com/exercism/elixir/issues/1644

### DIFF
--- a/exercises/concept/german-sysadmin/.docs/hints.md
+++ b/exercises/concept/german-sysadmin/.docs/hints.md
@@ -21,7 +21,7 @@
 
 ## 3. Substitute German characters
 
-- There is a [special syntax][syntax-reference-code-points] for getting a character's code point. It can be used in guards and it works for non-Latin letters too.
+- There is a [special syntax][syntax-reference-code-points] for getting a character's code point. It can be used in guards and it works for non-ASCII letters too.
 - There is [a built-in function][kernel-concat-list] that allows you to concatenate two lists.
 
 [syntax-reference-code-points]: https://hexdocs.pm/elixir/syntax-reference.html#integers-in-other-bases-and-unicode-code-points
--- a/exercises/concept/german-sysadmin/.docs/instructions.md
+++ b/exercises/concept/german-sysadmin/.docs/instructions.md
@@ -2,7 +2,7 @@
 
 You are working as a system administrator for a big company in Munich, Germany. One of your responsibilities is managing email accounts.
 
-You have been hearing complaints from people saying they are unable to write emails to Mr. Müller. You quickly realize that most of the company uses an old email client that doesn't recognize `müller@firma.de` as a valid email address because of the non-Latin character.
+You have been hearing complaints from people saying they are unable to write emails to Mr. Müller. You quickly realize that most of the company uses an old email client that doesn't recognize `müller@firma.de` as a valid email address because of the non-ASCII character.
 
 Telling people to give up their favorite old email client is a lost battle, so you decide to create sanitized aliases for all email accounts.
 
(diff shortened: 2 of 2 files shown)


angelikatyborska (approved):
Thank you very much!

Vukory:
Given that the letters like ü and ß *are* Latin, the test name `it allows the whole lowercase Latin alphabet` becomes ambiguous for anyone whose everyday Latin alphabet includes letters beyond a–z, and the word **whole** reinforces that, since it implies the complete alphabet. As Bosnian speaker, which Latin alphabet has more than 26 letters, that's how I read it myself.

"Latin" on its own doesn't say which letters are meant. And since the point of the PR is that these letters *are* Latin, using "Latin alphabet" as shorthand for exactly the 26 feels a little contradictory.

Would you be open to renaming the test to explicitly state that it's the 26 letters: `it allows all 26 lowercase Latin characters`?

It drops the ambiguity without getting pedantic and it reads the same for native and non-native speakers.

PR #1650, opened 2026-08-30
exercices/concept/secrets: Add link to the integer division function in the instructions

grgmiranda:
Just a little tweak, making more explicit what "integer function" refers to.

It's already linked in the HINTS, but not in the instructions.


angelikatyborska:
Hi! Thank you for the PR. I am afraid I need to reject it. Exercise introductions and instructions do not contain links _on purpose_. Exercise introductions and instructions should be concise and explain enough knowledge to be able to solve the exercise, but they shouldn't overexplain. Hints are meant to provide links to specific functions necessary to solve the exercise.

grgmiranda:
Oh, ok, sorry that I haven't notice this before, thanks!

PR #1651, opened 2026-09-04
Update filtering helper functions definition to be more direct

sabit990928:
### What and Why

I changed the phrasing in the wine-cellar exercise for helper functions to clarify the requirements. Because while solving the problem it seemed natural to extend filtering with additional match on `nil` case. Which seems like not intentional as my mentor mentioned. 

Maybe my wording is not best, but I think this part can be improved, so any feedback welcomed.


IsaacG:
+cc @exercism/elixir

